PR #23, opened 2021-01-26
[NN] TF2 Expanded Support

rshih32:
Fix multiple non eager use cases.
Fix subclassing support and tensor selection.
Separate out non eager execution, as it will try to apply to all pytest tests, and tensorflow undocumented test_util.run_all_in_graph_and_eager_modes support has missing test_session methods.

force tf.config.run_functions_eagerly(True) as graph mode graphs do not allow tensor manipulation outside of graph like for qoi calculations.


trulens/nn/models/tensorflow_v1.py_save
```diff
@@ -0,0 +1,353 @@
+import sys
```
divya-gopinath: what is this file doing? did you mean to delete? (extension .py_save is also weird)

divya-gopinath (approved):
LGTM but in future let's separate out PRs into format changes and code changes. It is difficult to parse through the diff when most of the changes are just format differences.

trulens/nn/models/tensorflow_v1.py_save
divya-gopinath: nvm, was looking through individual commits because the diff was huge. resolving.

PR #28, opened 2021-02-24
Allow all dois to specify its cut

rshih32:
This is a Doi level specification.

### DIFF
--- a/tests/unit/multi_qoi_test_base.py
+++ b/tests/unit/multi_qoi_test_base.py
@@ -25,7 +25,7 @@ class PerTimestepQoI(QoI):
 
 class RNNLinearDoi(DoI):
 
-    def __init__(self, baseline=None, resolution=10):
+    def __init__(self, baseline=None, resolution=10, cut: Cut = None):
         """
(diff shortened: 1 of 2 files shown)


tests/unit/multi_qoi_test_base.py
```diff
@@ -37,6 +37,7 @@ def __init__(self, baseline=None, resolution=10):
         resolution : int
             Number of points returned by each call to this DoI
```
divya-gopinath, line 38: can you add documentation for the new arg

divya-gopinath: (and check that the actual docs are updated)

divya-gopinath (approved):

tests/unit/multi_qoi_test_base.py
rshih32: The documentation exists for the parent class for now. There were some other requests along documentation when it had to do with parent and inherited classes so I'll figure out the best solution during that task.
I will also update the mkdocs on our next release

PR #36, opened 2021-04-06
Tensorflow Hub Layer Support

rshih32:
Handle Tensorflow 2 models written with tf hub modules.

Context:
As of 2.4.1, hub layers have become standardized, but the explanations require a unique flow. The models must be trained in eager mode, but the hub layers cannot take gradients and be explained in eager mode. 
https://github.com/tensorflow/hub/issues/622
Therefore, the model must be saved and reloaded without eager.

Added a warning if hub layers are detected, to save and reload without eager. run_functions_eagerly must also be reset to False for hub model loads to work. Hub layer detection is based on type KerasLayer. 

example hub layer types:
tensorflow_hub.keras_layer.KerasLayer
tensorflow.python.keras.saving.saved_model.load.KerasLayer

There may be non hub instances of KerasLayer so opting to not raise an exception, but warn and allow the user to make the change if needed.

Add the prepare_with_input method to fill any unfilled input placeholders in a hub layer when getting gradients. 

Add a rudimentary string anchor as hub layers have string indexed outputs.

Unit tests are willfully forgone because Tensorflow is changing their hub implementation and gradients in 2.5.1, in addition, the training and saving needed are expensive and not suitable for unit testing.

### DIFF
--- a/trulens/nn/models/keras.py
+++ b/trulens/nn/models/keras.py
@@ -253,6 +253,23 @@ class KerasModelWrapper(ModelWrapper):
             flat([layer.input for layer in layers])
             if cut.anchor == 'in' else flat([layer.output for layer in layers]))
 
+    def _prepare_intervention_with_input(
+            self, model_args, intervention, doi_tensors):
+        input_tensors = self._get_layers(InputCut())
+
+        if not all(elem in doi_tensors for elem in input_tensors):
+            doi_tensors.extend(input_tensors)
+            if len(intervention[0]) != len(model_args[0]):
+                doi_factor = len(intervention[0]) / len(model_args[0])
+                model_args_expanded = [
+                    get_backend().stack([arg] * int(doi_factor))
+                    for arg in model_args
+                ]
+                intervention.extend(model_args_expanded)
+            else:
+                intervention.extend(model_args)
+        return doi_tensors, intervention
+
     def fprop(
             self,
             model_args,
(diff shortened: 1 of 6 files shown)


trulens/nn/models/keras.py
```diff
@@ -242,10 +242,34 @@ def _get_layers(self, cut):
...
             if cut.anchor == 'in' else flat([layer.output for layer in layers]))
 
+    def _prepare_intervention_with_input(
+            self, model_args, intervention, doi_tensors):
+        input_tensors = self._get_layers(InputCut())
+        if not all(elem in doi_tensors for elem in input_tensors):
+
+            doi_tensors.extend(input_tensors)
+            if len(intervention[0]) != len(model_args[0]):
+                doi_factor = len(intervention[0]) / len(model_args[0])
```
divya-gopinath, line 263: nit: avoid calling the len of the first element twice, just add a variable like `intervention_len` or something like that. same with model args. just so it's clearer what this check is doing.

trulens/nn/models/tensorflow_v2.py
```diff
@@ -80,6 +82,14 @@ def call(*inputs, **kwargs):
     def B(self):
         return self._B
 
+    def _warn_keras_layers(self, layers):
+        keras_layers = [l for l in layers if 'KerasLayer' in str(type(l))]
+        if (keras_layers):
+            tru_logger.warn('Detected a KerasLayer in the model. This can sometimes create issues during attribution runs or subsequent model calls.\
+            If failures occur: try saving the model, de - activate eager mode with tf.compat.v1.disable_eager_execution(), \
+                set tf.config.run_functions_eagerly(False), then reload the model. '\
```
divya-gopinath, line 90: nit: grammar
"If failures occur: try saving the model, deactivating eager mode with `tf.compat.v1.disable_eager_execution()`, setting `tf.config.run_functions_eagerly(False)`, and reloading the model."

divya-gopinath (approved):

Issue #44, opened 2021-08-21
ValueError in the tensorflow 2 / keras notebook provided in the website.


DevavratSinghBisht:
ValueError in the tensorflow 2 / keras [notebook](https://colab.research.google.com/drive/1f-ETsdlppODJGQCdMXG-jmGmfyWyW2VD?usp=sharing) provided in the [website](https://www.trulens.org/welcome/).

The error occurs when executing the 6th code cell, it contains following code:
```
infl = InputAttribution(model)
attrs_input = infl.attributions(x_pp)
```

Error Log:
```
ValueError                                Traceback (most recent call last)
<ipython-input-18-9b10bbb33c8d> in <module>()
      1 infl = InputAttribution(model)
----> 2 attrs_input = infl.attributions(x_pp)

/usr/local/lib/python3.7/dist-packages/trulens/nn/attribution.py in attributions(self, *model_args, **model_kwargs)
    269             to_cut=self.slice.to_cut,
    270             intervention=D,
--> 271             doi_cut=doi_cut)
    272         # Take the mean across the samples in the DoI.
    273         if isinstance(qoi_grads, DATA_CONTAINER_TYPE):

/usr/local/lib/python3.7/dist-packages/trulens/nn/models/keras.py in qoi_bprop(self, qoi, model_args, model_kwargs, doi_cut, to_cut, attribution_cut, intervention)
    430             intervention, DATA_CONTAINER_TYPE) else [intervention]
    431 
--> 432         Q = qoi(to_tensors[0]) if len(to_tensors) == 1 else qoi(to_tensors)
    433 
    434         doi_tensors, intervention = self._prepare_intervention_with_input(

/usr/local/lib/python3.7/dist-packages/trulens/nn/quantities.py in __call__(self, y)
    109 
    110     def __call__(self, y: TensorLike) -> TensorLike:
--> 111         self._assert_cut_contains_only_one_tensor(y)
    112 
    113         if self.activation is not None:

/usr/local/lib/python3.7/dist-packages/trulens/nn/quantities.py in _assert_cut_contains_only_one_tensor(self, x)
     76                 '`{}` expected to receive an instance of `Tensor`, but '
     77                 'received an instance of {}'.format(
---> 78                     self.__class__.__name__, type(x)))
     79 
     80 

ValueError: `MaxClassQoI` expected to receive an instance of `Tensor`, but received an instance of <class 'keras.engine.keras_tensor.KerasTensor'>
```

klasleino:
[...]

PR #46, opened 2021-08-23
44 tensor type

klasleino:
Addresses #44. The issue was that the output of a keras model is `KerasTensor` which does not extend the base tensorflow `Tensor` type, so the check in QoI that requires the input to the QoI to be of type `get_backend().Tensor` was failing.

The proposed solution here is to use `get_backend().is_tensor(x)` instead of `isinstance(x, get_backend().Tensor)`. The former was updated to accept `KerasTensor`s in addition to the previously accepted types. Note that now `get_backend().Tensor` should not be used (except internally within the backend module).

### DIFF
--- a/trulens/nn/backend/__init__.py
+++ b/trulens/nn/backend/__init__.py
@@ -3,6 +3,7 @@ from enum import Enum
 import traceback
 import importlib
 from trulens.utils import tru_logger
+
 # Do not use directly, use get_backend
 _TRULENS_BACKEND_IMPL = None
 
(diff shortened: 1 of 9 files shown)


trulens/nn/backend/__init__.py
```diff
@@ -77,7 +84,6 @@ def get_backend(suppress_warnings=False):
 _ALL_BACKEND_API_FUNCTIONS = [
     'dim_order',
     'channel_axis',
-    'Tensor',
```
klasleino, line 80: I can put this back in if we think that's a good idea. My thought is it should be an internal type (since tensorflow has several tensor types, not just one), so removing it is the "right" thing to do. But on the other hand it does change the backend interface to remove it.

divya-gopinath (approved):

PR #58, opened 2021-12-15
tiling kwargs for pytorch model inputs

piotrm0:
Our NLP use case requires to tile model keyword arguments to align dimensions at doi_cut with dimensions of intervention. This implementation heuristically assumes the first model argument is batchable and every other argument that has the same first dimension as it are also batchable. The first model argument may be a keyword argument if no positional arguments are given.

Only pytorch backend is fully supported in this regard.


trulens/nn/models/pytorch.py
```diff
@@ -284,7 +286,23 @@ def fprop(
...
+                tile_shape[0] = doi_resolution
+                repeat_shape = tuple(tile_shape)
+
+                if isinstance(val, np.ndarray):
+                    val = np.tile(val, repeat_shape)
+
+                elif torch.is_tensor(val):
+                    val = val.repeat(repeat_shape)
+
+                batched_model_kwargs[k] = val
```
rshih32, line 302: nit: let's put this in a shared function that both can call.

trulens/nn/models/pytorch.py
```diff
@@ -269,7 +269,9 @@ def fprop(
         else:
             doi_repeated_batch_size = intervention[0].shape[0]
             batched_model_args = []
+            batched_model_kwargs = {}
 
+            # tile batches for interpolation
             for val in model_args:
                 doi_resolution = int(doi_repeated_batch_size / val.shape[0])
                 tile_shape = [1 for _ in range(len(val.shape))]
```
rshih32, line 277: Just a note, depending on how the memory task goes, we may want to do smarter checking and not tile everthing. We could check if doi_resolution is the expected length.
captured here: https://dev.azure.com/ailens-io/truera/_backlogs/backlog/truera%20Team/Epics/?workitem=5907

rshih32 (approved):

trulens/nn/models/pytorch.py
piotrm0: Did this, the first arg is expected to be batchable, all subsequent ones are batched only if they match the first's first dim.

trulens/nn/models/pytorch.py
piotrm0: Done.

trulens/nn/models/_model_base.py
```diff
@@ -172,8 +242,24 @@ def fprop(
                 as the input. If `attribution_cut` is supplied, also return the
                 cut activations.
         """
+
+        # TODO: Can attribution_cut preceed doi_cut or follow to_cut in the model?
```
rshih32, line 246: we can remove this todo and add to documentation. There are 2 questions here:
1) attribution cut cannot preceed  doi_cut (or gradients would break since expansion is not differentiable)
2) to_cut is not specifically tied to any of the attribution progression, it just says to which layer the fprop is going to. so it can be before or after any cut, as it can be to any layer.

trulens/nn/models/_model_base.py
```diff
@@ -172,8 +242,24 @@ def fprop(
...
+        # TODO: Use the defaults method for all backends.
+
         raise NotImplementedError
 
+    def _qoi_bprop_defaults(self, doi_cut, to_cut, attribution_cut):
+        """Defaults logic for qoi_bprop common across backends."""
```
rshih32, line 252: nit: same sentence fixing:

Replaces input parameters used in qoi_bprop to default values if not explicitly provided. These are common across backends.

trulens/nn/models/keras.py
```diff
@@ -318,20 +319,56 @@ def fprop(
...
-        # Convert `intervention` to a list of inputs if it isn't already.
-        if not isinstance(intervention, DATA_CONTAINER_TYPE):
-            intervention = [intervention]
+        print("doi_tensors=", doi_tensors)
+
+        # Finally, interventions override any previously set tensors.
+        val_map.update({k: v for k, v in zip(doi_tensors, intervention)})
+        # val_map.update({_tensor(k): v for k, v in intervention.kwargs.items()})
+
+        print(val_map)
```
rshih32, line 366: remove print statements

trulens/nn/models/pytorch.py
```diff
@@ -64,8 +68,14 @@ def __init__(
         self._layernames = list(layers.keys())
         self._tensors = list(layers.values())
 
+        if 0 == len(self._tensors):
```
rshih32, line 71: nit: swap the operands to be consistent with the rest of code:
len(self._tensors) == 0

trulens/nn/models/pytorch.py
```diff
@@ -442,26 +477,32 @@ def qoi_bprop(
...
         grads_list = []
+
         for z in zs:
             z_flat = ModelWrapper._flatten(z)
             qoi_out = qoi(y)
 
+            # TODO(piotrm): this sum is a source of much bugs for me when using
+            # attributions. If one wants a specific QoI, the sum hides the bugs
+            # in the definition of that QoI. It might be better to give an error
+            # when QoI is not a scalar.
```
rshih32, line 505: multi qoi is used for things like average qoi. its morre efficient to do it here and have 1 gradient rather than many gradients

trulens/utils/typing.py
```diff
@@ -5,19 +5,212 @@
...
+    # returning a new container.
+
+    @staticmethod
+    def lenses_elements(c: List[V]) -> Iterable['Lens[List[V], V]']:
+        """Lenses focusing on elements of a list."""
+
+        for i in range(len(c)):
+            yield Lens(
+                lambda l, i=i: l[i],
+                lambda l, v, i=i: IndexableUtils.with_(l, i, v))
```
rshih32, line 107: why do we need this class?

rshih32 (changes requested):

tests/pytorch/unit/model_wrapper_test.py
```diff
@@ -51,6 +51,43 @@ def forward(this, x):
...
+            def forward(this, X, Coeffs, divisor, **kwargs):
+                # Hoping kwargs forces the remaining arguments to be passed only by kwargs.
+
+                Degree = kwargs['Degree']
+                offset = kwargs['offset']
+
+                layer1 = this.layer1(X)
+
+                print("X=", X)
+                print("Coeffs=", Coeffs)
```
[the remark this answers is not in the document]
piotrm0: Yes, they are to be batched (non-capitals are parameters to not be batched).

trulens/nn/models/pytorch.py
piotrm0: I'd suggest to swap all others to "0 == ".

trulens/utils/typing.py
piotrm0: For unifying access to different type of args in ModelInputs .

trulens/nn/models/_model_base.py
```diff
@@ -132,22 +184,40 @@ def fprop(
...
-        `intervention`, and ending at `to_cut`.
+        Several cuts are parameters to this method. These designate how
+        information is propagated, intervened on, and returned by this method.
+        Models are evaluated starting from model_args, model_kwargs:
+
+            model_inputs -> doi_cut -> attribution_cut -> to_cut
+
+        Tensors for attribution_cut and to_cut are returned. However, the
+        returned values are batched according to the batching structure hinted
+        to by intervention. That is, if we have dimensions:
```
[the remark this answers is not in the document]
piotrm0: You mean doi_cut?

trulens/nn/models/_model_base.py
```diff
@@ -119,6 +123,54 @@ def __call__(self, x):
...
+                    "intervention needs to be given for DoI cuts that are not InputCut"
+                )
+            else:
+                # Using as_args here as sometimes interventions are passed in as single tensors but args needs a list.
+                intervention = ModelInputs(as_args(intervention), {})
+
+        return doi_cut, to_cut, intervention, model_inputs
+
     @abstractmethod
     def fprop(
```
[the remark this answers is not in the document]
piotrm0: Unsure what propagate typing meas.

trulens/nn/models/pytorch.py
rshih32: Even our inequalities are written in a way variable > constant. Not sure if there's a good reason to swap them all

trulens/nn/models/_model_base.py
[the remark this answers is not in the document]
rshih32: this fprop function needs typing.

trulens/nn/models/_model_base.py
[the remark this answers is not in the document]
rshih32: actually we can remove the cut mention, I think I didn't mean either
lets do 

However, the returned values are batched according to the batching structure hinted to by intervention, which could be larger due to expanding by the doi.

trulens/nn/models/_model_base.py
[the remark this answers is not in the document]
piotrm0: Done.

trulens/nn/models/_model_base.py
```diff
@@ -119,6 +123,54 @@ def __call__(self, x):
...
+            doi_cut = InputCut()
+        if to_cut is None:
+            to_cut = OutputCut()
+
+        if isinstance(doi_cut, InputCut):
+            # For input cuts, produce a ModelInputs container for the intervention and model inputs.
+            if intervention is not None:
+                tru_logger.warn(
+                    "intervention for InputCut DoI specified; this is ambiguous with model_args, model_kwargs"
+                )
```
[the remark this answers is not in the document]
piotrm0: Done.

trulens/nn/models/_model_base.py
```diff
@@ -119,6 +123,54 @@ def __call__(self, x):
         """
         return self.fprop(x)[0]
 
+    def _fprop_defaults(
+            self, model_args, model_kwargs, doi_cut, to_cut, intervention):
+        """Defaults logic for fprop common across backends. Also some warnings and errors."""
```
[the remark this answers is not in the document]
piotrm0: Done.

trulens/nn/models/_model_base.py
```diff
@@ -119,6 +123,54 @@ def __call__(self, x):
         """
         return self.fprop(x)[0]
 
+    def _fprop_defaults(
+            self, model_args, model_kwargs, doi_cut, to_cut, intervention):
```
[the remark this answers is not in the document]
piotrm0: Done.

trulens/nn/models/_model_base.py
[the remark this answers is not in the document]
piotrm0: Rephrased/elaborated.

trulens/nn/models/_model_base.py
piotrm0: Moving the documentation about these restrictions to bprop as I don't think they matter for fprop; fprop doesn't take any gradients

piotrm0: Also I'm unsure about your point (2), looks like the gradients done in bprop are to_cut w.r.t. attribution_cut so that implies an order regarding these two which would produce errors were it not to hold.

On the other hand, if attribution was before doi_cut, the evaluation might run but have unexpected results.

trulens/nn/models/_model_base.py
piotrm0: Done.

trulens/nn/distributions.py
```diff
@@ -283,7 +295,7 @@ def __init__(self, var: float, resolution: int, cut: Cut = None):
         self._var = var
         self._resolution = resolution
 
-    def __call__(self, z: ArrayLike) -> List[ArrayLike]:
+    def __call__(self, z: ArgsLike) -> List[ArgsLike]:
```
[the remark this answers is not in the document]
piotrm0: Done.

trulens/nn/distributions.py
```diff
@@ -211,7 +219,11 @@ def __call__(self, z: ArrayLike, *, model_inputs: Optional[ModelInputs] = None)
             for i in range(self._resolution)
         ]
 
-    def get_activation_multiplier(self, activation: ArrayLike, *, model_inputs: Optional[ModelInputs] = None) -> ArrayLike:
```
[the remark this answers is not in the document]
piotrm0: Done.

trulens/nn/models/keras.py
piotrm0: Done.

trulens/nn/models/pytorch.py
piotrm0: Generally "constant == variable" is preferable given the "constant = variable" typo becomes an error instead of a silent bug with "variable = constant" but it looks like python catches that for you so will switch.

piotrm0: Done.

trulens/nn/models/pytorch.py
piotrm0: I agree but I'm saying that even if one wants one gradient, like class qoi, but implements it incorrectly (assuming class qoi was not already provided for them), then this sum might hide their mistake(s).

rshih32: let add a warning that this sum is usually for multi output.

trulens/utils/typing.py
rshih32: resolved, just add a link to Lenses
maybe clarify C-> container, V->value
and that this is to help treat args/kwargs similarly as a convenience

piotrm0: Done.

tests/pytorch/unit/model_wrapper_test.py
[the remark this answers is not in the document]
piotrm0: Added explanation in other places.

trulens/nn/models/pytorch.py
piotrm0: Added a warning only if more than 1 dimension is being aggregated.

rshih32 (approved):

PR #60, opened 2022-01-13
add train mode to model before bprop

rshih32:
if using cuda, the model must be in train mode for gradients to work


arridhana (commented):
Do we also set this back to eval mode whenever we're done with computation that requires backprop?

rshih32:
> Do we also set this back to eval mode whenever we're done with computation that requires backprop?

done

arridhana (approved):

piotrm0:
You mean before fprop? fprop with train has non-determinism issues even on cpu.

PR #63, opened 2022-02-22
NLP examples notebook

piotrm0:
NLP input attribution example notebook based on a model from hugging face. Utilities to make that possible/convenient to implement and to visualize.

Also included a few things to handle backend detection and pytorch device specification issues.


trulens/visualizations.py
```diff
@@ -999,3 +1012,301 @@ def __call__(
             attrs_input, x, output_file, blur, threshold, masked_opacity,
             combine_channels
         )
+
+
+class Output(ABC):
```
rshih32, line 1017: nit:TextOutput
This looks like it deals specifically with text

trulens/visualizations.py
```diff
@@ -999,3 +1012,301 @@ def __call__(
...
+
+    @abstractmethod
+    def render(self, s: str) -> str:
+        ...
+
+
+# TODO(piotrm): implement a latex output format
+
+
+class Text(Output):
```
rshih32, line 1052: nit: if we are saying Output is TextOutput, maybe we call this PlainText

rshih32 (approved):

trulens/visualizations.py
piotrm0: Don't mean it to handle text only but that is what is available presently.

trulens/visualizations.py
piotrm0: Renamed to PlainText .

Issue #62, opened 2022-02-10
Random gradients in torch summarizer pipeline


rshih32:
This code has different gradients each time

```python
from trulens.nn.attribution import InputAttribution, InternalInfluence
from trulens.nn.attribution import IntegratedGradients
from trulens.nn.quantities import MaxClassQoI
from trulens.nn.distributions import PointDoi, LinearDoi
from trulens.nn.slices import Cut, InputCut, OutputCut, Slice
from trulens.nn.models import get_model_wrapper

from transformers import pipeline

summarizer = pipeline("summarization")

summarizer(
    """
    America has changed dramatically during recent years. Not only has the number of 
    graduates in traditional engineering disciplines such as mechanical, civil, 
    electrical, chemical, and aeronautical engineering declined, but in most of 
    the premier American universities engineering curricula now concentrate on 
    and encourage largely the study of engineering science. As a result, there 
    are declining offerings in engineering subjects dealing with infrastructure, 
    the environment, and related issues, and greater concentration on high 
    technology subjects, largely supporting increasingly complex scientific 
    developments. While the latter is important, it should not be at the expense 
    of more traditional engineering.

    Rapidly developing economies such as China and India, as well as other 
    industrial countries in Europe and Asia, continue to encourage and advance 
    the teaching of engineering. Both China and India, respectively, graduate 
    six and eight times as many traditional engineers as does the United States. 
    Other industrial countries at minimum maintain their output, while America 
    suffers an increasingly serious decline in the number of engineering graduates 
    and a lack of well-educated engineers.
"""
)

m = summarizer.model.to("cpu")
tokenizer = summarizer.tokenizer

batch_sentences = ["Hello I'm a single sentence", "And another sentence", "And the very very last one"]
batch = tokenizer(batch_sentences, padding=True, truncation=True, return_tensors="pt").to(m.device)
input = batch['input_ids']

wrapped_model = get_model_wrapper(m, input_shape=(None, 7), device='cpu')

embed_cut = Cut('model_encoder_layers_0_self_attn_q_proj', anchor='in')
from trulens.nn.quantities import QoI
import torch
class SummarizerQoI(QoI):
    def __call__(self, seq_2_seq_output):
        logits = seq_2_seq_output['logits']
        
        max_token, max_indices = torch.max(logits,dim=-1)
        qoi = torch.mean(max_token, 1, True)
        return qoi

infl = InternalInfluence(
    wrapped_model, 
    Slice(embed_cut, OutputCut()), 
[...]

PR #66, opened 2022-03-16
bugfix pytorch non determinism

piotrm0:
Fixes https://github.com/truera/trulens/issues/62 .


rshih32:
We will need to make sure this still works with cuda on windows

rshih32 (approved):

PR #68, opened 2022-04-01
private vs. public design

piotrm0:
There are two parts to this PR in preparation for rebatching work. First a small context manager for catching memory issues:

```
        with memory_suggestions(*param_msgs):  # Handles out-of-memory messages.
            # some memory-intensive computations
```

The second part is design of public and private methods with regard to the types of data they accept. The whole thing is documented in `typing.py` and pasted below:

# Type aliases and data container classes. 

Large part of these utilities are meant to support a design decision: public
methods are permissive in the types they accept while private methods are not. 

For example, 

    attributions(*model_args: ArgsLike, **model_kwargs: ModelKwargs)

For models that take in a single input, model_args can be a single Tensor or
numpy array. For models that take in multiple inputs, model_args must be some
iterable over Tensors or numpy arrays. The implication is that the user can call
this method as in "attributions(inputA)" instead of attributions([inputA]). The
other permissibility is that both backend Tensors and numpy arrays are supported
and the method returns results in the same type. 

List of flexible typings:

    - Single vs. Multiple Inputs -- Can pass in a single value instad of list.
      This is indicated by the `ArgsLike` type alias (paraphrased):

      ``` ArgsLike[V] = Union[
        V, Inputs[V] # len != 1 
      ]
      ```

        - AttributionMethod.attributions: model_args argument
        - DoI.__call__: z argument
        - QoI.__call__: y argument
        - LinearDoI.__init__: baseline argument
        - LinearDoI.baseline (when callable): first argument
        - ModelWrapper.fprop: model_args, intervention arguments
        - ModelWrapper.qoi_bprop: model_args argument

    - Single vs. Multiple QoI outputs -- A quantity of interest may return one
      or more tensors.

        - QoI.__call__ return

    - Backend Tensor vs. Numpy Array -- Can pass in a backend Tensor or numpy
      array. This is indicated by the `DataLike` type alias:

      ``` DataLike = Union[np.ndarray, Tensor] ```

        - AttributionMethod.attributions: model_args, model_kwargs arguments
          (return type based first arg in either).
        - DoI.__call__: z and model_inputs arguments (return type based on z).
        - QoI.__call__: y argument. Note however that QoI must return Tensor(s).
        - LinearDoI.baseline (when callable): first argument
        - ModelWrapper.fprop: model_args, model_kwargs, intervention arguments
        - ModelWrapper.qoi_bprop: model_args, model_kwargs, intervention
          arguments.

      Of the various flexibilities for the public mentioned here, this is the
      only one that is also maintained in the private methods.

    - ModelInputs input -- Can define method to accept ModelInputs structure or
      just the args. To indicate the latter, must be keyword argument named
      "model_inputs".

        - DoI.__call__
        - LinearDoI.baseline (when callable): see BaselineLike.

    - ModelInputs or args -- Can pass in single value, iterable, dictionary, or
      ModelInputs structure. These correspond to single positional arg, multiple
      positional args, keyword args, and (both positional and keyword args).
    
        - ModelWrapper.fprop: intervention argument, see InterventionLike (paraphrased):

        ``` InterventionLike = Union[
              ArgsLike[DataLike], KwargsLike, ModelInputs]
        ```

        - ModelWrapper.qoi_bprop: intervention argument: see InterventionLike

    - Callable Baselines -- Baselines can be methods that return baseline
      values. See (paraphrased) BaselineLike (paraphrased):

      ``` BaselineLike = Union[
          ArgsLike[DataLike], ArgsLike[DataLike] -> ArgsLike[DataLike],
          (ArgsLike[DataLike], ModelInputs) -> ArgsLike[DataLike]
      ]
      ```

        - LinearDoI.__init__: baseline argument

Part of the implementation of this design are private variants of public methods
that do not have most of the above flexibility. Most, however, have the tensor
vs. numpy array flexibility. These are to be used within Truera for more
consistent expectations regarding inputs/outputs.

    - DoI._wrap_public_call vs. DoI.__call__
    - DoI._wrap_public_get_activation_multiplier
    - QoI._wrap_public_call vs. QoI.__call__
    - ModelWrapper._fprop vs. ModelWrapper.fprop
    - ModelWrapper._qoi_bprop vs. ModelWrapper.qoi_bprop


notebooks/pytorch_work.ipynb.not
```diff
@@ -0,0 +1,633 @@
+{
```
rshih32, line 1: is this needed?

trulens/nn/attribution.py
```diff
@@ -54,7 +67,9 @@ class AttributionMethod(AbstractBaseClass):
     """
 
     @abstractmethod
-    def __init__(self, model: ModelWrapper, *args, **kwargs):
+    def __init__(
+        self, model: ModelWrapper, doi_per_batch=None, *args, **kwargs
```
rshih32, line 71: nit: batch_per_doi/rebatch_per_doi/rebatch_doi/rebatch_for_doi, etc seems like a more accurate names

trulens/nn/attribution.py
```diff
@@ -245,101 +270,114 @@ class given by the first integer against the class given by the
...
+        if len(list(model_inputs.values())) == 0:
+            batch_size = 1
+        else:
+            batch_size = model_inputs.first().shape[0]
 
-        # DoI supports tensor or list of tensor. unwrap args to perform DoI on
-        # top level list
+        param_msgs = [
+            f"float size = {B.floatX_size} ({B.floatX}); consider changing to a smaller type.",
+            f"batch size = {batch_size}; consider reducing the size of the batch you send to the attributions method."
```
rshih32, line 297: nit: use the parameter name so they know which parameter is being referred to

trulens/nn/attribution.py
```diff
@@ -245,101 +270,114 @@ class given by the first integer against the class given by the
...
-            qoi=self.qoi,
-            model_args=model_inputs.args,
-            model_kwargs=model_inputs.kwargs,
-            attribution_cut=self.slice.from_cut,
-            to_cut=self.slice.to_cut,
-            intervention=D,
-            doi_cut=doi_cut
-        )
+        # TODO: Consider doing the model_inputs tiling here instead of inside qoi_bprop.
+        effective_batch_size = doi_per_batch * batch_size
```
rshih32, line 325: I know this might because the review is early: is this used yet?
also if it is to be used later, thats fine. might not be bad to have separate PR

tests/unit/model_wrapper_test_base.py
```diff
@@ -205,117 +205,8 @@ def test_fprop_kwargs_intervention(self):
...
-        )[0]
-
-        # Expect handling of the non-batched values (divisor, offset) as in
-        # numpy broadcasting.
-        expected = ((X + 1.0)**(Degree + 4.0)
-                   ) * (Coeffs + 2.0) / (divisor + 3.0) + (offset + 5.0)
-
-        self.assertTrue(np.allclose(actual, expected))
-
-    def test_fprop_kwargs_intervention_tiling(self):
```
rshih32, line 217: will look to have replacement tests for these I presume?

trulens/nn/attribution.py
rshih32: actually, I see that effective_batch_size is not yet used, so I'm not exactly sure the end product of what this parameter will do. either way, maybe add a docstring on that

piotrm0: Accidental inclusion. Removing from this PR.

trulens/nn/attribution.py
piotrm0: For the ones here, I don't think there is a consistent parameter name they could use to adjust things. They have to look up how, for example, batch sizing is done in whatever system they are using.

notebooks/pytorch_work.ipynb.not
piotrm0: No, removed.

rshih32 (approved):

PR #80, opened 2022-06-15
TFHub Support

coreyhu:
Adds TFHub and nested model compatibility to the Keras backend. 
Eager execution should also work now (previously the Keras backend can't take gradient in Eager mode).


rshih32 (changes requested):

trulens/nn/models/keras.py
```diff
@@ -65,13 +97,18 @@ def __init__(
             modified. By default `False`.
         """
         self.keras = import_keras_backend()
+        self.tf = import_tensorflow()
+        self.tfhub = import_tfhub_deps()
```
piotrm0, line 101: Is this saying we only support tfhub models? (this fails if tfhub not installed)

piotrm0: Ohh, nevermind, read it wrong.

trulens/nn/models/keras_utils.py
```diff
@@ -0,0 +1,354 @@
...
+            break
+
+        non_input_layers = []
+        for layer in keras_model.layers:
+            if not isinstance(layer, keras_module.layers.InputLayer):
+                layer._name = f"{keras_model.name}/{layer.name}"
+                non_input_layers.append(layer)
+
+        if len(non_input_layers) != 1:
+            break
```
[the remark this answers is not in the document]
coreyhu: If there's more than one non-input layer, it's possible they're not run sequentially. The logic about their run order and which inputs/output tensors are fed to the next layer are contained in the parent model, so I can't dive deeper.

This loop is there to strip away input layers from deeply nested models, since the inputs don't do much more than forward their values to the next depth's input layers.

Going from `[input *3 > [input * 3 > [encoder_layers]]]`
to just `[encoder_layers]`

piotrm0:
Is there an example tfhub model usage notebook (even if its not a test) as part of this PR and if not, can one be included?

trulens/nn/models/keras_utils.py
```diff
@@ -250,7 +278,12 @@ def tracer(depth):
                     innode_indices[out_layer.name] = idx
         tracer(depth + 1)
 
-    tracer(0)
+    try:
+        innode_indices = {}
+        nodes_by_depth = model._nodes_by_depth
+        tracer(0)
+    except AttributeError:
+        return {}
```
rshih32, line 286: Can we log what this might be instead of silent fail?

coreyhu:
> Is there an example tfhub model usage notebook (even if its not a test) as part of this PR and if not, can one be included?

I'm working a usage notebook and unit tests that will be added before taking the PR out of draft

trulens/nn/models/keras_utils.py
```diff
@@ -52,56 +55,71 @@ def recurse_outputs(obj, prefix=None):
...
-            arg_paths = [get_arg_path(layer, arg) for arg in args]
-            kwarg_paths = {
-                key: get_arg_path(layer, arg) for key, arg in kwargs.items()
-            }
-
-            if next_layer.name not in layer_input_paths:
-                layer_input_paths[next_layer.name] = {
-                    "args": arg_paths,
-                    "kwargs": kwarg_paths
+    node_depths = sorted(list(model._nodes_by_depth.keys()), reverse=True)
```
coreyhu, line 87: Using model nodes here since each layer can have multiple output nodes if used in different models.

trulens/nn/models/keras_utils.py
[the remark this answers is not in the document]
coreyhu: The model can be correctly defined with connected input and output tensors. However, Keras doesn't support taking gradients of layers in outer models wrt layers in nested models. Therefore, we need to flatten the architecture so that layers of interest are exposed in the outermost model.

I made a change to properly flatten nested models so that this loop isn't necessary. It uses the inner model's `.call()` method to apply the "inner model" logic onto the "outer model" context and expose nested layers. Hopefully should be cleaner/less hacky than stripping non-input layers and calling them individually. Let me know what you think!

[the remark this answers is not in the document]
rshih32: perfect, Thanks Corey!

rshih32 (approved):

PR #81, opened 2022-06-23
TF Keras NLP Demo Notebook

coreyhu:
Adds a NLP demo notebook for TF/keras. Also links to colab on the Readme


rshih32 (approved):
Can we make get_model_wrapper(model, backend="tf.keras") 
be able to discern the backend?

piotrm0 (approved):

PR #82, opened 2022-07-05
Optional param to call model.eval()

coreyhu:
Adds optional parameter `force_eval=True` to PytorchModelWrapper. Useful since `backward()` with LSTMs must be done in training mode (or with cuda disabled).

### DIFF
--- a/trulens/nn/models/pytorch.py
+++ b/trulens/nn/models/pytorch.py
@@ -33,7 +33,15 @@ class PytorchModelWrapper(ModelWrapper):
     of Pytorch nn.Module objects.
     """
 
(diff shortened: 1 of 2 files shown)


rshih32 (approved):

trulens/nn/models/pytorch.py
```diff
@@ -337,7 +349,8 @@ def hookfn(self, inpt, outpt):
         with memory_suggestions(device=self.device):
             # Run the network.
             try:
-                self._model.eval()  # needed for determinism sometimes
+                if self.force_eval:
```
piotrm0, line 352: might no be necessary unless there are some .train() still someplace

PR #85, opened 2022-07-19
Update versioning to 0.0.12.1

coreyhu:


docs/overrides/home.html
```diff
@@ -119,7 +119,7 @@
                             </svg>
                             <div class="header__btn--text d-f fd-c">
                                 <span class="strong">DOWNLOAD</span>
-                                <span>trulens 0.0.12 (Jul 5, 2022)</span>
+                                <span>trulens 0.0.12.1 (Jul 19, 2022)</span>
```
rshih32, line 122: do we want to just "fix" the version to 0.12.1?
I think we probably were not using the version numbers correctly

PR #87, opened 2022-08-10
adding better handling of pytorch devices and test notebook

piotrm0:
get_model_wrapper modified to keep pytorch default device the same as the device of the wrapped model. Issues some warnings if they differ before making them the same.

### DIFF
--- a/trulens/nn/backend/pytorch_backend/pytorch.py
+++ b/trulens/nn/backend/pytorch_backend/pytorch.py
@@ -3,6 +3,7 @@
 # pylint: disable=no-member
 # pylint: disable=not-callable
 
+from typing import Union
 import numpy as np
 import torch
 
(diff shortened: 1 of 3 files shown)


trulens/nn/backend/pytorch_backend/pytorch.py
```diff
@@ -45,28 +49,23 @@ def set_default_device(device: str):
...
+    if isinstance(device, str):
+        device = torch.device(device)
 
-    if "cuda" in device:
-        default_device = torch.device(device)
-    elif device == "cpu":
-        default_device = torch.device(device)
-    else:
-        raise ValueError(f"Unhandled device type {device}")
+    default_device = device
```
coreyhu, line 55: maybe I'm wrong but I think we need a `nonlocal` or `global` to scope `default_device`

coreyhu (approved):

PR #91, opened 2022-09-04
Extending visualization for 3D input data

daniel-huang-1230:


trulens/nn/quantities.py
```diff
@@ -217,6 +217,20 @@ def __call__(self, y: TensorLike) -> TensorLike:
                     format(self._channel_ax)
                 )
 
+        elif len(B.int_shape(y)) == 5:
```
rshih32, line 220: maybe just merge this with the last one and do len(B.int_shape(y)) >=4

trulens/visualizations.py
```diff
@@ -454,18 +455,269 @@ def _get_hotcold(self):
         return ListedColormap(hotcold, name='hotcold')
 
 
+class Visualizer3D(object):
```
rshih32, line 458: instead of copy pasting, can we do inheritance and modular methods?

rshih32 (changes requested):

trulens/visualizations.py
daniel-huang-1230: Yes this is the plan - sorry for the late reply I started this just for experimentation so earlier I made this change without. 

I'll have a starter notebook  published and reviewed, and for sure refactor this part.

PR #93, opened 2022-09-13
Tensorflow placeholder support

rshih32:
anything that affects the tensors, can also now nest apply down dictionaries as well.
This is a typical flow for tensorflow placeholders

### DIFF
--- a/trulens/nn/backend/__init__.py
+++ b/trulens/nn/backend/__init__.py
@@ -9,8 +9,10 @@ import numpy as np
 
 from trulens.utils import tru_logger
 from trulens.utils.typing import ModelInputs
-from trulens.utils.typing import nested_map, om_of_many
-from trulens.utils.typing import TensorAKs, Tensors
+from trulens.utils.typing import nested_map
+from trulens.utils.typing import om_of_many
+from trulens.utils.typing import TensorAKs
+from trulens.utils.typing import Tensors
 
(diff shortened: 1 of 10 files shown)


trulens/utils/typing.py
```diff
@@ -278,6 +278,11 @@ def nested_map(y: OMNested[C, U],
         for i in range(len(y)):
             out.append(nested_map(y[i], fn, nest - 1))
         return y.__class__(out)
+    if isinstance(y, MAP_CONTAINER_TYPE) and nest > 0:
```
coreyhu, line 281: nitpick - `MAP_CONTAINER_TYPE` could be substituted with `collections.abc.Mapping` as a generic base class to check against

coreyhu (approved):

trulens/nn/backend/__init__.py
```diff
@@ -151,7 +150,7 @@ def tile_val(val):
         elif B.is_tensor(val):
             return B.tile(val, repeat_shape)
         else:
-            raise ValueError(f"unhandled tensor type {val.__class__.__name__}")
+            tru_logger.debug(f"Ignoring tiling of unhandled val {val.__class__.__name__}")
```
piotrm0, line 153: return val here if this is just a warning

piotrm0 (approved):

PR #98, opened 2022-09-16
Set torch device as the majority device on model.

rshih32:
More automation rather than asking user to do it themselves. There is already an option to send the device as is, which should not require you to do more.


trulens/nn/models/pytorch.py
```diff
@@ -77,40 +78,18 @@ def __init__(
...
-
-        else:
-            model.to(device)
-
-            def_device = pytorch.get_default_device()
-            if device != def_device:
-                tru_logger.warning(
-                    f"Model's device ({device}) differs from pytorch's default device ({def_device}). Changing default to model device."
+            try:
+                device_counter = Counter(
```
coreyhu, line 82: Maybe more of an edge case but how do we want to deal with models living on multiple GPUs (whether to parallelize, or not enough memory to fit on one GPU)? Is it feasible to skip the `model.to(self.device)` and keep the model as-is?

coreyhu (approved):

trulens/nn/models/pytorch.py
rshih32: I thought about this and I think we should cross that bridge when we get to it.
There's the more common case that they never actually specify their model device, and we end up using some default. of which the model.to is useful in fixing a user mistake which would otherwise fail.

I think since that is a more complex scenario, that should require more deliberation by the user, so we may add a "skip device defaulting" option in the future.

PR #102, opened 2022-10-03
Use Cut accessor method in forward prop

coreyhu:
Useful for getting valid activations and mirrors bprop logic

### DIFF
--- a/trulens/nn/models/_model_base.py
+++ b/trulens/nn/models/_model_base.py
@@ -249,7 +249,7 @@ class ModelWrapper(AbstractBaseClass):
             intervention=intervention,
             **kwargs
         )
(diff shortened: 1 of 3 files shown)


trulens/nn/models/_model_base.py
```diff
@@ -249,7 +249,7 @@ def fprop(
             intervention=intervention,
             **kwargs
         )
-
+        rets = (to_cut.access_layer(rets[0]), rets[1])
```
piotrm0, line 252: Is there a reason the same isn't done for doi_cut?

piotrm0 (approved):

rshih32 (approved):

PR #103, opened 2022-10-04
Fix unittest hierarchy and import errors

coreyhu:
- Makes unittest structures into valid python modules by adding missing `__init__.py`s. 
 - Additionally, raises an early import error for wrong TF version in TF1 tests. 
 - Also removed settings since formatter is already defined in format.sh


.vscode/settings.json
```diff
@@ -1,3 +0,0 @@
-{
-    "python.formatting.provider": "yapf"
```
piotrm0, line 2: But we still need this when formatting within vscode.. ?

piotrm0 (approved):

Issue #72, opened 2022-04-14
QoI has no access to instance being explained


piotrm0:
Quantities of interests receive output values on DoI samples but these are not necessarily the instance being explained, except in some cases like point. Because of this, one cannot define a QoI for "explained instance predicted class logits" because predicted class is not know unless the DoI sample is the same as the explained instance.

A related issue is that some may interpret QoI like the default max to mean what I described above but that is not what the max QoI implements because of the issue noted.

PR #104, opened 2022-10-06
Non visualization fixes and utilities

piotrm0:
- Reformatted a lot of comment text.
- Added some new comments for todos or unknown points.
- Added __str__ to various trulens types.
- Cleared notebook outputs.
- Added some error handling and suggestion if value cannot be extracted from fprop.
- Added another NLP baseline that swaps particular tokens.
- Added utilities for running notebooks in colab.

### DIFF
--- a/trulens/nn/quantities.py
+++ b/trulens/nn/quantities.py
@@ -16,7 +16,7 @@ roof of the car, a “car feature” not shared by convertibles.
 from abc import ABC as AbstractBaseClass
 from abc import abstractmethod
 from inspect import signature
-from typing import Any, Callable, List, Optional, Union
+from typing import Callable, List, Optional, Union
 
 from trulens.nn.backend import get_backend
 from trulens.utils.typing import DATA_CONTAINER_TYPE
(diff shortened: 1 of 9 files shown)


trulens/nn/quantities.py
```diff
@@ -46,6 +44,13 @@ class QoI(AbstractBaseClass):
     output behavior that the attributions describe.
     """
 
+    def __str__(self):
+        return render_object(self, [])
+
+    # TODO: Need to give a seperate value of y at target instance here since
+    # these are values are interventions. Cannot presently define a QoI that says:
+    # logits of the predicted class for each instance.
```
rshih32, line 52: are these comments tied to anything?

rshih32 (approved):

trulens/utils/colab.py
```diff
@@ -0,0 +1,60 @@
+# Some utilities to help with colab/jupyter notebook demos.
+
+import importlib
+from pathlib import Path
+import subprocess
+import sys
+
+
+def install_if_not_installed(packages):
```
coreyhu, line 9: Are these used anywhere?

coreyhu (approved):

trulens/utils/colab.py
piotrm0: In notebooks to be run easily in colab but I see none of them are presently in this PR. They are part of the visualizations PR. Exception to include them in this PR?

trulens/nn/quantities.py
piotrm0: Yes. https://github.com/truera/trulens/issues/72 and some task on the old task tracker which I cannot find now. Will make a new one and refer to it in the comment.

PR #145, opened 2023-06-02
Leaderboard ux

joshreini1:
Add conditional description with emojis using st.metric delta


trulens_eval/trulens_eval/Leaderboard.py
```diff
@@ -63,23 +65,51 @@ def app():
...
+                            value=f'{round(mean, 2)}',
+                            delta='✅ High'
+                        )
+                    else:
+                        feedback_cols[i].metric(
+                            label=col_name,
+                            value=f'{round(mean, 2)}',
+                            delta='⚠️ Low ',
+                            delta_color="inverse"
+                        )
```
rshih32, line 112: what's going on here? Why is this duplicated?

trulens_eval/trulens_eval/Leaderboard.py
```diff
@@ -28,6 +28,8 @@ def app():
     )
     df, feedback_col_names = lms.get_records_and_feedback([])
 
+    test_threshold = 0.5
```
rshih32, line 31: Our current implementation also used a threshold. Is there somewhere this constant can be moved so that it is shared among all downstream?

rshih32 (changes requested):

trulens_eval/trulens_eval/Leaderboard.py
joshreini1: Seems reasonable to move it to tru.py, but where else is the threshold used?

joshreini1: or alternatively, as a constant in the Feedback() class?

rshih32: Not sure, but we had red/green text being chosen by something before, did we not?

rshih32: nvm, search `0.5` in  Evaluations.py 
Lets move both of those to `tru_feedback.py` and name the variable `default_pass_fail_color_threshold`

piotrm0 (approved):

rshih32 (approved):

PR #180, opened 2023-06-09
Human feedback docs

rshih32:


docs/trulens_eval/logging.ipynb
```diff
@@ -130,6 +130,29 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "b9a1393f",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "thumb_result = True\n",
+    "tru.add_feedback(name=\"thumbs up result\", \n",
+    "                  record_id=record.record_id,\n",
+    "                  chain_id=truchain.chain_id, \n",
```
shayaks, line 152: Non-blocking: Ideally, we don't need to add a chain_id because record_ids are uuids. Is there a way to do that?

shayaks (approved):

docs/trulens_eval/logging.ipynb
rshih32: Added it as a follow up. Probably can be a wider upgrade outside of this function too

PR #183, opened 2023-06-09
Thumbs demo

rshih32:


trulens_eval/trulens_eval/examples/Example_Thumbs_Application.py
```diff
@@ -0,0 +1,90 @@
...
+# Set up Streamlit app
+st.title("Get Help from ChatGPT")
+user_input = st.text_input("What do you need help with?")
+
+if user_input:
+    # Generate GPT-3 response
+    prompt_input = user_input
+    # add context manager to capture tokens and cost of the chain
+    with get_openai_callback() as cb:
+        gpt3_response, record = generate_response(prompt_input, tc)
```
shayaks, line 57: This seems a bit superfluous as we are tracking the the cost and tokens ourselves in truchain, so we don't need to nest this within a get_openai_callback?

trulens_eval/trulens_eval/examples/Example_Thumbs_Application.py
```diff
@@ -0,0 +1,90 @@
...
+        tru.add_record(record)
```
shayaks, line 58: Why do we need to call add_record here? Isn't this going to be recorded in truchain?

trulens_eval/trulens_eval/examples/Example_Thumbs_Application.py
```diff
@@ -0,0 +1,90 @@
...
+    # Allow user to rate the response with emojis
+    col1, col2 = st.columns(2)
+    with col1:
+        thumbs_up = st.button("👍")
+    with col2:
+        thumbs_down = st.button("👎")
+
+    thumb_result = None
+    if thumbs_up:
+        # Save rating to database or file
```
shayaks, line 75: This comment seems a unnecessary here (unless we save the rating).

shayaks (approved):
Suggested some cleanup for the example.

PR #186, opened 2023-06-12
Change default behavior for tru in TruModel

shayaks:


trulens_eval/trulens_eval/tru_model.py
```diff
@@ -75,6 +75,17 @@ def __init__(
         feedbacks: Optional[Sequence[Feedback]] = None,
         **kwargs
     ):
+        if tru is None:
```
rshih32, line 78: can we update the quickstart too?

```
truchain = TruChain(chain,
    chain_id='Chain3_ChatApplication',
    feedbacks=[f_lang_match])
```

rshih32 (approved):

PR #187, opened 2023-06-12
fix latency bugs

rshih32:
* Backwards compatibility on ingested records with no latency. Set old record latency to a default value
* Remove latency from feedback insert. This field was never added


piotrm0:
Note that I adjusted latency tracker in recent PR so this might not apply.

PR #192, opened 2023-06-12
making llama_index and langchain optional

piotrm0:


trulens_eval/trulens_eval/util.py
```diff
@@ -40,6 +41,51 @@
 UNICODE_CHECK = "✅"
 UNCIODE_YIELD = "⚡"
 
+# Optional requirements.
+
+REQUIREMENT_LLAMA = (
+    "llama_index module is required for instrumenting llama_index apps. "
```
shayaks, line 47: Do we want to suggest a minimum supported version here?

piotrm0: Added suggestion but am not checking for it.

shayaks (approved):

PR #209, opened 2023-06-15
trullama model->app

piotrm0:
TruLllama bug fix due to outdated names.


piotrm0:
Contents now in https://github.com/truera/trulens/pull/210

PR #215, opened 2023-06-15
llama more instruments

piotrm0:
Added llama_index instrumentation for:
- prompts
- llms
- Various other llama_index base classes with uncertain generality.

Organized component types across langchain and llama_index and used organization in UI.


trulens_eval/DEPRECATION.md
```diff
@@ -1,6 +1,19 @@
...
+- json serialization of WithClassInfo mixin
+    - key "class_info" renamed to "__tru_class_info"
+
+- json serialization of ObjSerial class
+    - init_kwargs expanded into init_bindings, serialized by a new class
+      Bindings, containing both args and kwargs
+
+## Changes in 0.2.0
+
+### Non-breaking changes to be breaking in 1.0.0
```
rshih32, line 16: maybe remove the 1.0.0 version since we don't know the version.

Also; what does non-breaking to be breaking mean?

rshih32 (approved):

trulens_eval/DEPRECATION.md
piotrm0: What do you mean we don't know? I'm assuming it is a standard that major version numbers can introduce breaking changes but minor versions cannot hence I'm suggesting that the aliases introduced here (to make the changes non-breaking) will be removed (hence become breaking) at 1.0.0 the earliest.

piotrm0: I reworded the points.

PR #241, opened 2023-06-28
2 more methods for configuring keys

piotrm0:
Organizing various ways of getting api keys and checking for them. In the simplest case, checking for a key is done using `check_keys` which checks various sources of keys and fails if any of the required keys is not set. It also prints a warning if more than one different value for a key was found:

Screenshot 2023-07-17 at 19 07 26


trulens_eval/trulens_eval/keys.py
```diff
@@ -76,16 +155,39 @@ def get_huggingface_headers():
...
+  - in a .env file in {Path.cwd()} or its parents,
+  - explicitly passed to function `check_keys` of `trulens_eval.keys`,
+  - passed to the endpoint or feedback collection constructor that needs it (`trulens_eval.provider_apis.OpenAIEndpoint`, etc.), or
+  - set in api utility class that expects it (i.e. `openai`, etc.).
+
+For the last two options, the name of the argument may differ from {k} (i.e. `openai.api_key` for `OPENAI_API_KEY`).
+""")
+
+
+def check_keys(**kwargs):
```
rshih32, line 176: This call is just confusing to me... It adds one layer of API call on top of the already used set key methods. Then, if they are already set; its going to ignore what you type in here? But then theres no way to simply check the key without adding bogus string in.

_check_key makes sense because if no val is supplied it, will check existence. Can we just remove the "set" functionality of this method?

rshih32 (changes requested):

trulens_eval/trulens_eval/keys.py
piotrm0: Did this. The setting functionality is only part of new `check_or_set_keys`.

trulens_eval/trulens_eval/keys.py
```diff
@@ -77,55 +172,182 @@ def get_huggingface_headers():
     return HUGGINGFACE_HEADERS
 
 
-def setup_keys(**kwargs):
-    global config_file
+def _value_is_set(v: str) -> bool:
+    return not(v is None or "fill" in v or v == "")
```
rshih32, line 176: I know its unlikely; but maybe we should check against a const 
USER_FILL_SUGGEST="to fill in..."

Two reasons
* There is a very low chance that the substring fill shows up in a key string
* Better code debuggability as this check is 2 methods down from the _check_or_set function of which this is the main branching for that method's behavior

rshih32 (approved):

trulens_eval/trulens_eval/keys.py
piotrm0: Did this.

rshih32 (approved):

PR #254, opened 2023-06-30
Fix tagging

joshreini1:
Make some changes to fix tagging.

User asked about adding metadata


piotrm0:
Hard to read this PR as it says almost every file was changed.

trulens_eval/trulens_eval/schema.py
```diff
@@ -48,6 +48,7 @@
 
 RecordID = str
 AppID = str
+tags = str
```
rshih32, line 51: Probably should capitalize this, otherwise the variables might get confused

trulens_eval/trulens_eval/schema.py
```diff
@@ -400,6 +404,7 @@ def __init__(
             app_id = obj_id_of_obj(obj=self.dict(), prefix="app")
 
         self.app_id = app_id
+        self.tags = tags
```
rshih32, line 407: like here, tags could be confused as local scope or global scope. So have the outer tags renamed to Tags

rshih32 (changes requested):

trulens_eval/trulens_eval/app.py
```diff
@@ -276,6 +276,7 @@ def _post_record(
         ret_record_args['cost'] = cost
```
rshih32, line 276: @joshreini1 are tags ever in documentation or tested? Can we add that?

rshih32 (changes requested):

trulens_eval/trulens_eval/app.py
joshreini1: Added tag to quickstart example

rshih32 (approved):

PR #260, opened 2023-06-30
Google colab

rshih32:


trulens_eval/setup.py
```diff
@@ -23,7 +23,7 @@
         'requests>=2.30.0',
         'slack-bolt>=1.18.0',
         'slack-sdk>=3.21.3',
-        'streamlit>=1.22.0',
+        'streamlit>=1.13.0',
```
piotrm0, line 26: If important, make a note that version downgrade is important.

piotrm0 (approved):

PR #265, opened 2023-07-05
Contribution guidelines

joshreini1:


README.md
```diff
@@ -10,7 +10,11 @@ TruLens provides a set of tools for developing and monitoring neural nets, inclu
 
 ![Architecture Diagram](https://www.trulens.org/Assets/image/TruLens_Architecture.png)
 
-### Get going with TruLens-Eval
+### 💡 Contributing
+
+Interested in contributing? See our [contribution guide](trulens_eval/CONTRIBUTING.md) for more details.
+
+### Example Usage
```
rshih32, line 17: the readme is autogenerated so this will be overridden. Add this to the docs/trulens_eval/intro.md and let the autogenerate PR add it after its merged

rshih32 (changes requested):

README.md
rshih32: Also; have links reference exact urls rather than relative paths, as autogenerated files are placed in multiple different path locations

rshih32 (approved):

PR #273, opened 2023-07-11
CI/CD Formatting FIx

coreyhu:
Fix formatting in pipelines


trulens_eval/trulens_eval/util.py
```diff
@@ -1281,7 +1281,7 @@ def noserio_issubclass(self, class_name: str, module_name: str):
 
 # inspect.signature does not work on builtin type constructors but they are used
 # like this method. Use it to create a signature of a builtin constructor.
-def builtin_init_dummy(self, /, *args, **kwargs):
```
coreyhu, line 1284: @piotrm0 is this syntax intentional?

piotrm0: Yes. See https://docs.python.org/3/faq/programming.html#what-does-the-slash-in-the-parameter-list-of-a-function-mean . This is how builtin functions are signed.

PR #278, opened 2023-07-12
Feedback functions for evaluating against ground truth

shayaks:


trulens_eval/trulens_eval/feedback.py
```diff
@@ -740,6 +740,48 @@ def relevance(self, prompt: str, response: str) -> float:
...
+
+        def _find_response(self, prompt: str) -> Optional[str]:
+            responses = [qr["response"] for qr in self.ground_truth if qr["query"] == prompt]
+            if responses:
+                return responses[0]
+            else:
+                return None
+
+
+        def measure(self, prompt: str, response: str) -> float:
```
joshreini1, line 756: Can we just have on agreement function that simply measures similarity and use that for both model_agreement and golden set agreement?

And then in the class definition is where the comparison set is defined (either supplied golden set or generated model responses).

This would make the experience more similar for the two options. End user feedback definition would then be like:

groundtruth = Feedback(openai.Comparison(golden_set).agreement).on_input_output()
modelagreement = Feedback(openai.Comparison(openai.model_generated).agreement).on_input_output()

Imo, this is more clear on when LLMs are used and what we're measuring

shayaks: @joshreini1 what were you thinking of as the type of openai.model_generated? Is it just a callable that returns a response to a request?

joshreini1: Exactly. Could also be extended to different providers, e.g. anthropic.model_generated, or specific models perhaps, e.g. openai.gpt4_generated

PR #293, opened 2023-07-13
redact keys from dashboard

piotrm0:
Also started to split some parts off of util.py as it is too big and it is getting harder to read and avoid circular imports.


trulens_eval/trulens_eval/keys.py
```diff
@@ -12,16 +12,59 @@
...
 
 logger = logging.getLogger(__name__)
 
+# Keep track of values that should not be shown in UI (or added to DB). This set
+# is only used for cases where the name/key for a field is not useful to
+# determine whether it should be redacted.
+values_to_redact = set()
+
+# Regex of keys (into dict/json) that should be redacted.
+RE_KEY_TO_REDACT = re.compile('|'.join([
```
joshreini1, line 33: Can we be more exhaustive in key checking? User could be using any number of models in their app (Anthropic, Cohere, Bard, ...)

piotrm0: Added coverage to the ones I could find, let me know if others are missing.

trulens_eval/trulens_eval/ux/components.py
```diff
@@ -157,6 +158,14 @@ def draw_llm_info(query: JSONPath, component: ComponentView) -> None:
                 """
     df = pd.DataFrame.from_dict(llm_kv, orient='index').transpose()
 
+    # Redact any column whose name indicates it might be a secret.
+    for col in df.columns:
+        if should_redact_key(col):
```
rshih32, line 163: Do we need to do this on DB read of previously ingested keys? Or do we want to migrate the DB to remove them permanently?

rshih32 (changes requested):

trulens_eval/trulens_eval/ux/components.py
piotrm0: This is a point of discussion, should we redact before adding to database or only before showing to the user. There are some implications of the first one for my big replay task. Will schedule a time to discuss but can we go with the second option for now?

rshih32: yes lets do at least one, good with either for short term. I'm not 100% on which one is "second option" so if you can comment back explicitly on which one you choose.

piotrm0: Redact when presenting to the user but keep the keys in database (so that they can be used to replay chains).

rshih32:
Can you check that all writes are un-redacted? Though if you don't figure it out here, you might run into with the replay. we wanted no-redact on write, and redact on read. It looked to me like you were redacting the key on db write (via jsonify), but maybe I didn't follow right.

piotrm0:
> Can you check that all writes are un-redacted? Though if you don't figure it out here, you might run into with the replay. we wanted no-redact on write, and redact on read. It looked to me like you were redacting the key on db write (via jsonify), but maybe I didn't follow right.

Checked. `redact_keys` is False by default and is set to true only in 

```
def jsonify_for_ui(*args, **kwargs):
    return jsonify(*args, **kwargs, redact_keys=True, skip_specials=True)
```

rshih32 (approved):

PR #295, opened 2023-07-15
feedback implementations checks and tests

piotrm0:
Feedback implementations can be functions as well as methods of custom classes (and subclasses of provider as before).

If deferred evaluation mode is used, they must additionally be importable. This should fail at App construction if they are not importable.

If custom classes are used, they must have no constructor arguments.

Implementations still cannot be staticmethods, classmethods, or non-importable anything for deferred mode.

All failure modes are tested for and report user feedback except staticmethods, which I'm having trouble checking for right now.

### DIFF
--- a/trulens_eval/trulens_eval/util.py
+++ b/trulens_eval/trulens_eval/util.py
@@ -1486,9 +1486,10 @@ class FunctionOrMethod(SerialModel):
 
         if inspect.ismethod(c):
             self = c.__self__
-
             return Method.of_method(c, obj=self, loadable=loadable)
+        
         else:
+            
             return Function.of_function(c, loadable=loadable)
 
     def load(self) -> Callable:
@@ -1543,7 +1544,7 @@ class Function(FunctionOrMethod):
 
     module: Module
     cls: Optional[Class
-                 ]  # for static methods in a class which we view as functions
+                 ]  # for static methods in a class which we view as functions, not yet supported
     name: str
 
     @staticmethod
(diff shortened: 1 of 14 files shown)


trulens_eval/trulens_eval/util.py
```diff
@@ -1246,16 +1246,30 @@ def base_class(self) -> 'Class':
...
         return self
 
+    def _check_importable(self):
+        try:
+            cls = self.load()
+        except Exception as e:
+            raise ImportError(
+                f"Class {self} is not importable. "
+                "If you are defining custom feedback function implementations, make sure they can be imported by python scripts. "
+                "If you defined a function in a notebook, it will not be importable."
```
rshih32, line 1256: I think we need to allow local running of notebook defined functions. So far in this; I see that the failure happens at Feedback object creation. Is it possible to only fail on deserialization back from db?

piotrm0: We do allow, the important part is at https://github.com/truera/trulens/pull/295/files#diff-d2f476a66d230f8a610e2908982218970c2cc19ec1627bdef62d7744595f8eddR492

piotrm0: Added test to make sure: https://github.com/truera/trulens/pull/295/files#diff-d2f476a66d230f8a610e2908982218970c2cc19ec1627bdef62d7744595f8eddR492

rshih32 (approved):

PR #301, opened 2023-07-18
refine result pass/fail feedback categories in ui

piotrm0:
Categories now include "warning" with medium scores and pass requires a higher threshold than before. Makes examples look more prescriptive. Moved all of this out of feedback.py and into ux/styles.py .

Screenshot 2023-07-17 at 17 19 37
Screenshot 2023-07-17 at 17 19 30

### DIFF
--- a/trulens_eval/trulens_eval/ux/styles.py
+++ b/trulens_eval/trulens_eval/ux/styles.py
@@ -37,9 +37,11 @@ class CATEGORY:
         icon="?"
     )
 
+    ALL = [PASS, WARNING, FAIL] # not including UNKNOWN intentionally
+
     @staticmethod
     def of_score(score: float) -> Category:
-        for cat in [CATEGORY.PASS, CATEGORY.WARNING, CATEGORY.FAIL]:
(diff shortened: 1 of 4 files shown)


trulens_eval/trulens_eval/ux/styles.py
```diff
@@ -1,12 +1,63 @@
...
+    @staticmethod
+    def of_score(score: float) -> Category:
+        for cat in CATEGORY.ALL:
+            if score >= cat.threshold:
+                return cat
+
+        return CATEGORY.UNKNOWN
+
+
+# These would be useful to include in our pages but don't yet see a way to do
```
joshreini1, line 51: Do you mean just to show what the thresholds are?

joshreini1 (approved):

trulens_eval/trulens_eval/ux/styles.py
piotrm0: Mostly no, I meant to be able to define JS variables that contain the threshold for various purposes like coloring or otherwise.

PR #306, opened 2023-07-20
feedback metadata

piotrm0:
Feedback implementations can return a dict in addition to a float. These are interpreted as columns to add to the feedback result tables in the ui:
Screenshot 2023-07-19 at 19 30 07


trulens_eval/trulens_eval/feedback.py
```diff
@@ -1278,6 +1261,104 @@ def sentiment(self, text: str) -> float:
...
+        if ground_truth_response:
+            agreement_txt = _get_answer_agreement(
+                prompt, response, ground_truth_response
+            )
+            ret = _re_1_10_rating(agreement_txt) / 10, dict(ground_truth_response=ground_truth_response)
+        else:
+            ret = np.nan
+        return ret
+
+def model_agreement(self, prompt: str, response: str) -> float:
```
shayaks, line 1326: Why did we move this out of the openai provider class? In fact, we'll likely just want to deprecate this.

shayaks: Can we provide a deprecation notice on this one?

trulens_eval/trulens_eval/schema.py
```diff
@@ -281,6 +281,10 @@ class FeedbackCall(SerialModel):
     args: Dict[str, str]
     ret: float
 
+    # New in 0.4.0: Any additional data a feedback function returns to display
```
shayaks, line 284: This is probably new in 0.6.0?

joshreini1 (approved):
lgtm, tested llama and langchain quickstarts

PR #312, opened 2023-07-24
async acall for langchain

piotrm0:


trulens_eval/tests/unit/test_tru_chain.py
```diff
@@ -68,6 +66,59 @@ def setUp(self):
 
         self.llm = HuggingFacePipeline(pipeline=self.pipe)
 
+    @unittest.skip("TODO")
```
rshih32, line 69: is this still todo?

piotrm0: Partially, cost tracking sometimes fails but enabled test otherwise.

rshih32 (approved):

PR #314, opened 2023-07-24
improvement(record viewer): Modified the record viewer UI

walnutdust:
https://github.com/truera/trulens/assets/36098015/e3d82e5d-6b0e-4c07-a22e-bff8b6405c31

- Moved from timeline by itself to using a table
- Rows represent the call stack, added the capability to expand and hide the different rows
- When a row is selected the relevant timeline bar is highlighted and relevant information is shown below 
- Selectors are now also shown beside the call name
- Fixed previous bug with parallel calls not showing
- Calls are expanded by default


trulens_eval/trulens_eval/react_components/record_viewer/__init__.py
```diff
@@ -6,7 +6,7 @@
 # the component, and True when we're ready to package and distribute it.
 # (This is, of course, optional - there are innumerable ways to manage your
 # release process.)
-_RELEASE = True
```
shayaks, line 9: Could we keep _RELEASE = True in the commits?

shayaks (approved):

trulens_eval/trulens_eval/react_components/record_viewer/__init__.py
walnutdust: thanks for catching it - my bad

shayaks (approved):

PR #319, opened 2023-07-26
Add model metadata

shayaks:
Model metadata shows up on the leaderboard on hover:
image

This also shows up on the evaluations page as metadata:
image

This can be added in on any app as a dict:
```
tru_query_engine = TruLlama(query_engine,
    app_id='LlamaIndex_App1',
    feedbacks=[f_lang_match, f_qa_relevance, f_qs_relevance, f_groundtruth],
    metadata = {'model': 'default', 'temperature': 0}
    # feedback_mode=FeedbackMode.DEFERRED # should cause failure but does not
)
```


trulens_eval/trulens_eval/schema.py
```diff
@@ -439,6 +442,7 @@ def __init__(
         kwargs['app_id'] = "temporary"  # will be adjusted below
         kwargs['feedback_mode'] = feedback_mode
         kwargs['tags'] = ""
```
piotrm0, line 444: if we want tags and metadata to impact the generated app_id, set them here

piotrm0 (approved):

trulens_eval/trulens_eval/schema.py
rshih32: Not clear if it should or should not denote a new app_id. Will re-evaluate as usage comes in

PR #344, opened 2023-08-01
no root methods and recordkeeping with context managers

piotrm0:
Marking `App.*_with_record` calls as to deprecate and replaced by `App.with_record` and `App.awith_record`. Also other pass through methods that do not return a record but still do tracking are likewise marked to be replaced with `App.with_` and `App.awith_`.

```python
app = ...
tru_app = TruChain(app)
res_with, record_with = tru_app.with_record(app, "Who is Shayak?") # or app.__call__
```
 
Adding context manager based record gathering:

```python
with tru_app as recording:
    res_context = app("Who is Shayak?")  # should be same as res_with

record_context = recording.get() # should be same as record_with
```

Multiple records at once:
```python
with tru_app as recording:
    res_context1 = app("Who is Shayak?")
    res_context2 = app("Who is Shameek?")

records_context = recording.records # list of two records
```

If you do not want to pointers to records (but still have them show up on dashboard):

```python
res_with = tru_app.with_(app, "Who is Shayak?") # or app.__call__

# or with context manager:

with tru_app:
    res_context1 = app("Who is Shayak?")
    res_context2 = app("Who is Shameek?")
```

Calls to `app` in the above refer to calls to `app.__call__` and can be replaced by calls to any instrumented method. Calls to `app` outside of the context manager or one of the `with` methods will not be tracked.

Both invocation options support record-level metadata but in the `with_` options, care must be taken with regards to keyword vs. positional arguments. Metadata comes after positional arguments if specified by position.

```python
res_with = tru_app.with_(app, "Who is Shayak?", "this is metadata")
res_with = tru_app.with_(app, "Who is Shayak?", record_metadata="this is metadata")
res_with = tru_app.with_(app, record_metadata="this is metadata", question="Who is Shayak?")

# or with context manager:

with tru_app as recording:
    recording.record_metadata="this is metadata for all records in this context that follow this line"
    res_context1 = app("Who is Shayak?")
    res_context2 = app("Who is Shameek?")
```


joshreini1:
What’s the difference in app, truapp? Why do we have both?

piotrm0:
> What’s the difference in app, truapp? Why do we have both?

`app` is the user's app (like langchain Chain) whereas `tru_app` is TruChain or similar.

joshreini1 (approved):

rshih32 (approved):

PR #369, opened 2023-08-09
add azure openai to groundedness

joshreini1:
Fix groundedness to work with AzureOpenAI.

Also add a llama-index / AzureOpenAI example that includes groundedness to demonstrate usage.


trulens_eval/trulens_eval/feedback.py
```diff
@@ -1557,6 +1557,24 @@ def groundedness_measure(self, source: str, statement: str) -> float:
...
+                    score = self.groundedness_provider._doc_groundedness(
+                        premise=source, hypothesis=hypothesis
+                    )
+                    reason = reason + str.format(
+                        feedback_prompts.GROUNDEDNESS_REASON_TEMPLATE,
+                        statement_sentence=hypothesis,
+                        supporting_evidence="[Doc NLI Used full source]",
+                        score=score * 10,
+                    )
+                    groundedness_scores[f"statement_{i}"] = score
```
rshih32, line 1576: how is this not breaking?

_doc_groundedness is only defined for huggingface

joshreini1: Thanks for catching - this should be resolved.

joshreini1:
Currently unable to test successfully because running into content filtering with groundedness function.

rshih32 (approved):

PR #374, opened 2023-08-11
splitting large files and organizing into folders

piotrm0:
- Split `feedback.py` into the contents of the new `feedback` folder and smaller things into utils.
- Split `provider_apis.py` into `feedback/provider` folder contents.

## Module organization/dependency

Modules on lower lines should not import modules on same or above lines as
otherwise you might get circular import errors.

- `__init__.py`

- all UI/dashboard components

- `tru_chain.py` `tru_custom_app.py`

- `tru_llama.py` (note: llama_index uses langchain internally for some things)

- `tru.py`

- `utils`

    - `llama.py` `langchain.py` `trulens.py`

- `feedback`

    - `__init__.py`

    - `provider`

        - `__init__.py`

        - `endpoint`

            - `__init__.py`

            - `openai.py` `hugs.py`

            - `base.py` 

        - `hugs.py` `openai.py` `cohere.py`

        - `base.py`

    - `groundedness.py` `groundtruth.py`

    - `feedback.py` `prompts.py`

- `tru_basic_app.py` TODO: bad placement

- `app.py`

- `db.py`

- `instruments.py`

- `schema.py`

- `util.py` 

- `keys.py`

- `utils`

    - `python.py` `text.py` `generated.py`


joshreini1:
This is much needed - good idea.

Might be worth also changing the rest of our "Tru" prepended modules to just their names? This would be more consistent with how LangChain/Llama-Index handle their integrations

shayaks (approved):

piotrm0:
> This is much needed - good idea.
> 
> Might be worth also changing the rest of our "Tru" prepended modules to just their names? This would be more consistent with how LangChain/Llama-Index handle their integrations

The "Tru" modules suggestion needs some discussion.

Issue #373, opened 2023-08-11
[Bug] TruChain with AgentExecutor breaks LangChain


lariel-fernandes:
This bug affects release 0.9.0 and it was probably introduced with #362 (I tested with the codebase before and after the merge, before it works fine).

### Context
While trying to reproduce [this example from LangChain](https://python.langchain.com/docs/modules/agents/agent_types/openai_functions_agent) I noticed that the `TruChain` with an `AgentExecutor` causes `TypeError: 'NoneType' object is not subscriptable` in `langchain.chains.base`

### Reproduce
```python
from langchain import LLMMathChain
from langchain.agents import initialize_agent, Tool, AgentType
from langchain.chat_models import ChatOpenAI
from trulens_eval import Tru, TruChain

llm=ChatOpenAI(temperature=0, model="gpt-3.5-turbo-0613")

llm_math_chain = LLMMathChain.from_llm(llm, verbose=True)

tools = [
    Tool(
        name="Calculator",
        func=llm_math_chain.run,
        description="useful for when you need to answer questions about math"
    ),
]

agent = initialize_agent(tools, llm, agent=AgentType.OPENAI_FUNCTIONS, verbose=True)

app = TruChain(agent)

# works fine
agent(inputs={"input": "how much is Euler's number divided by PI"})

# raises `TypeError: 'NoneType' object is not subscriptable`
app(inputs={"input": "how much is Euler's number divided by PI"})
```
Output logs and stacktrace here: [logs.txt](https://github.com/truera/trulens/files/12323098/logs.txt)

piotrm0:
Thanks for report. Take a look at https://github.com/truera/trulens/pull/376 for fix.

lariel-fernandes:
Tested this with trulens-eval 0.10.0, it doesn't break anymore but it prints a couple warnings:
```
A new object of type <class 'langchain.chains.llm_math.base.LLMMathChain'> at 0x7ff89668ef00 is calling an instrumented method <function Chain.__call__ at 0x7ff8a74ab5e0>. The path of this call may be incorrect.
Guessing path of new object is *.app based on other object (0x7ff8965c0300) using this function.
A new object of type <class 'langchain.chains.llm.LLMChain'> at 0x7ff8974d86c0 is calling an instrumented method <function Chain.__call__ at 0x7ff8a74ab5e0>. The path of this call may be incorrect.
Guessing path of new object is *.app based on other object (0x7ff8965c0300) using this function.
```
Should that be Ok?

[...]

PR #376, opened 2023-08-11
patch instrumentation losing track of instrumented objects

piotrm0:
Might be a temporary solution to https://github.com/truera/trulens/issues/373 . Will investigate more later.


trulens_eval/examples/frameworks/langchain/langchain_math_agent.ipynb
shayaks, line 1: Please remove debug info from user facing notebook.

PR #380, opened 2023-08-15
Fix tru_basic_app outdated construction

piotrm0:


joshreini1:
Tested manually - the no_framework_quickstart notebook runs but doesn't create records. May be an issue with sqlalchemy thing though

piotrm0:
> Tested manually - the no_framework_quickstart notebook runs but doesn't create records. May be an issue with sqlalchemy thing though

My fault. Check now?

joshreini1:
> > Tested manually - the no_framework_quickstart notebook runs but doesn't create records. May be an issue with sqlalchemy thing though
> 
> My fault. Check now?

Works now 👍

joshreini1 (approved):

PR #386, opened 2023-08-16
Formatting for Langchain and Llama-Index examples

joshreini1:
Standardize formatting, make sure everything runs on the listed package versions and add descriptions where helpful.


piotrm0:
Why remove the interactive version of langchain async example?

rshih32 (approved):

PR #387, opened 2023-08-16
Updates to documentation for remote database

shayaks:
Added a new entry under logging for 'Where to Log'
image

image

Also fixed some references in the API reference for Tru.


joshreini1:
This is straightforward - don't think we need example notebooks, but it might be good here to add a few other example db connectors. Could also be in a separate page in docs called "DB Connectors" or similar

piotrm0:
Add a note of what the url equivalent is of the old database file specification?

PR #393, opened 2023-08-17
Using AWS Bedrock as Eval Model Provider

joshreini1:
AWS Bedrock can now be used as a model provider for evals for any eval that uses an LLM.

Usage Example:

`bedrock_provider = Bedrock(model_id = "amazon.titan-tg1-large", region_name="us-east-1")`

`f_relevance = Feedback(bedrock_provider.relevance).on_input_output()`

Also factor out feedback functions from openai, litellm to base LLMProvider class that is inherited


joshreini1:
NOTE: should not merge until okay from AWS

joshreini1:
Wait until feedback refactor done. Will be easier to set up feedback with bedrock at this stage

trulens_eval/trulens_eval/feedback/provider/bedrock.py
```diff
@@ -0,0 +1,1028 @@
...
+from trulens_eval.feedback import prompts
+from trulens_eval.feedback.provider.base import Provider
+from trulens_eval.utils.generated import re_1_10_rating
+
+import json
+
+logger = logging.getLogger(__name__)
+
+
+class Bedrock(Provider):
```
rshih32, line 13: Instead of dup code; can we do inheritance? 

Parent provider class something like LLMProvider(Prodvider) (has all the dup code)

and do 
```
OpenAI(LLMProvider):
    @override
    def _create_chat_completion
    @override
    etd
```
and 

```
Bedrock(LLMProvider):
    @override
    def _create_chat_completion
    @override
    etd
```

rshih32 (changes requested):

trulens_eval/trulens_eval/feedback/provider/bedrock.py
joshreini1: The usage examples are provider specific so we'd need to sacrifice that to inherit the methods with the provider implementation abstracted

rshih32: The usage examples can be unique; you just add the specific one in the override and then call the super() method. Code duplication is very hard to maintain

rshih32: Also; the usage examples are barely provider specific; right? At this point its probably better to just have mkdocs show the base class LLMProvider and have the docs there show two pieces. 

for OpenAI
```
import trulens....OpenAI
provider = OpenAI()
```

for Bedrock
```
import trulens....Bedrock
provider = Bedrock()
```

```
rest of the usage is the same
```

rshih32: Also; if model_engine is an issue; just make model_engine do nothing in bedrock and move it to the parent class and note that it is currently only needed for OpenAI() provider

joshreini1: Can you take another look @rshih32 ? I factored out all of the LLM feedbacks into base and just overriding where necessary in openai, litellm, bedrock

docs/trulens_eval/api/feedback.md
```diff
@@ -16,6 +16,8 @@ Any Custom App Wrapper: [TruCustomApp](https://www.trulens.org/trulens_eval/api/
 
 ::: trulens_eval.trulens_eval.feedback.provider.litellm
 
+::: trulens_eval.trulens_eval.feedback.provider.bedrock
+
```
rshih32, line 20: nit: also add the base class here, also maybe rename to llm_provider instead of base for clarity

trulens_eval/requirements.txt
```diff
@@ -3,6 +3,7 @@ python-dotenv
 langchain
 sqlalchemy
 alembic
+overrides
```
rshih32, line 6: nit: sorry about pointing you to the @override annotation. I think in python you can just not annotate it at all and it'll work so this probably is not needed

trulens_eval/trulens_eval/feedback/provider/base.py
```diff
@@ -17,3 +25,920 @@ def __init__(self, name: str = None, **kwargs):
...
+        self, question: str, statement: str
+    ) -> float:
+        """
+        Uses chat completion model. A function that completes a
+        template to check the relevance of the statement to the question.
+        Also uses chain of thought methodology and emits the reasons.
+
+        **Usage:**
+        ```
+        feedback = Feedback(provider.qs_relevance_with_cot_reasons).on_input_output()
```
rshih32, line 196: maybe for each of these add

Intantiate a provider with either [OpenAI](https://www.trulens.org/trulens_eval/api/feedback/#trulens_eval.trulens_eval.feedback.provider.openai.OpenAI) or [BedRock](https://www.trulens.org/trulens_eval/api/feedback/#trulens_eval.trulens_eval.feedback.provider.bedrock.Bedrock)

trulens_eval/trulens_eval/feedback/provider/openai.py
```diff
@@ -13,12 +14,13 @@
 logger = logging.getLogger(__name__)
 
 
-class OpenAI(Provider):
+class OpenAI(LLMProvider):
     """Out of the box feedback functions calling OpenAI APIs.
```
rshih32, line 18: Also add to each of the inheritors

See available feedback functions in the base [LLMProvider](https://www.trulens.org/trulens_eval/api/feedback/#trulens_eval.trulens_eval.feedback.provider.%3CCHANGEME%3E.LLMProvider)

rshih32 (changes requested):

trulens_eval/trulens_eval/feedback/provider/openai.py
joshreini1: I added this into new provider api references

trulens_eval/trulens_eval/feedback/provider/base.py
joshreini1: Added new provider api reference pages that should make this more clear.

joshreini1:
@piotrm0 - note that I'm also changing the provider organization in this PR

rshih32 (approved):

trulens_eval/trulens_eval/feedback/provider/openai.py
```diff
@@ -262,1071 +262,6 @@ def moderation_not_violencegraphic(self, text: str) -> float:
             openai_response["results"][0]["category_scores"]["violence/graphic"]
         )
 
-    def _find_relevant_string(self, full_source, hypothesis):
-        return self.endpoint.run_me(
-            lambda: self._create_chat_completion(
-                model=self.model_engine,
```
rshih32, line 268: this needs to be added to base

PR #398, opened 2023-08-24
record-level metadata

piotrm0:
Can specify metadata when creating a record:
```python
_, rec = tc.call_with_record(message, meta=meta)
```

Or can add it after but need to update the db:

```python
_, rec = tc.call_with_record(message)
rec.meta = new_meta
Tru().update_record(rec)
```

Metadata is JSON-like, so base types like strings, ints, floats are allowed as well as structures composed of Dicts and Sequences.


joshreini1:
'meta' seems like a weird name for this arg. Can we pick something more obvious?

Perhaps 'record_metadata'?

shayaks (approved):
Looks good. Agree with @joshreini1 and would be good to rename to to something more descriptive and less meta like record_metadata.

PR #408, opened 2023-08-30
make additional instrumentation part of base instruments

piotrm0:
Added capability to instrument more than the default calls. For example, templates in langchain are currently not instrumented, but they can be via this snippet (must be done before a TruChain is initialized).

```python
from trulens_eval.instruments import instrument
from langchain.prompts.base import BasePromptTemplate
instrument.method(BasePromptTemplate, "format")
```

Calls to templates (at least the format method) will now be included in record calls (and the dashboard). You can check what has been instrumented via `App.print_instrumented`:

```python
tc: TruChain = ...
tc.print_instrumented()
```

Output now includes something like this, for example:

```
Object at 0x141c69700:
	<function PromptTemplate.format at 0x1273d49d0> with path *.__app__.app.memory.prompt
	<function BasePromptTemplate.format at 0x120b20ca0> with path *.__app__.app.memory.prompt
```


trulens_eval/trulens_eval/instruments.py
```diff
@@ -1020,3 +1020,43 @@ def instrument_object(self, obj, query: Query, done: Set[int] = None):
             logger.debug(
                 f"{query}: Do not know how to instrument object of type {cls}."
             )
+
+
+class instrument:
```
rshih32, line 1025: nit: naming capitalized?  Maybe name InstrumentedDecorator

trulens_eval/trulens_eval/tru_custom_app.py
```diff
@@ -348,45 +348,18 @@ def __getattr__(self, __name: str) -> Any:
             )
 
 
-class instrument:
+from trulens_eval.instruments import instrument as base_instrument
+
+class instrument(base_instrument):
```
rshih32, line 353: same nit class names capitalized

rshih32 (approved):

trulens_eval/trulens_eval/instruments.py
piotrm0: this one is used like a method as in `@instrument` . I did however separate out the utilities not for decoration from this "class" and named it with a capital `AddInstruments`.

trulens_eval/trulens_eval/tru_custom_app.py
piotrm0: same

PR #412, opened 2023-08-31
Add bertscore, rouge and bleu scores

shayaks:
This adds Bert scores, Rouge and BLEU scores to the ground truth comparision feedback function collection.


trulens_eval/trulens_eval/feedback/groundtruth.py
```diff
@@ -92,3 +116,89 @@ def agreement_measure(
...
+            prompt (str): A text prompt to an agent. response (str): The
+            agent's response to the prompt.
+
+        Returns:
+            - float: A value between 0 and 1. 0 being "not in agreement" and 1
+                being "in agreement".
+            - dict: with key 'ground_truth_response'
+        """
+        if self.bert_scorer is None:
+            self.bert_scorer = BERTScorer(lang="en", rescale_with_baseline=True)
```
rshih32, line 137: why does bert_scorer need to be a self property?

rshih32 (changes requested):

trulens_eval/trulens_eval/feedback/groundtruth.py
shayaks: Because each time we initialize a BERTScorer, this loads a large-ish BERT model from disk. This makes it quite expensive to do on each  feedback function call.

rshih32: got it. Can we put that in a comment in the code so its documented? This PR should be fine; as its loaded in memory, but it could pop up in the future as any property on feedback functions get serialized to DB (of which this is probably not serializable). Just want to make sure any dev has context and doesn't remove it if something does pop up

rshih32 (approved):

trulens_eval/trulens_eval/feedback/groundtruth.py
shayaks: I can do that. Also open to ideas on how to make this serializable.

PR #441, opened 2023-09-12
adding more instruments to llama_index agents

piotrm0:
Instrumenting more components of the llama_index yelp agent example. Timeline now shows more details (see screenshots). Added component categories to the dashboard for Agent, Tool, and added LLM component for llama_index LLMs.

Screenshot 2023-09-11 at 18 03 14

# Add to Release notes:

Change to layout of calls in `Record.layout_calls_as_app` affects how selectors are written. Assume instrumented app has a method `somemethod` which is called more than once on an invocation of the app. Previously, `Select.Record.app.somemethod` would contain call information for only the last call of that method. Now it will contain all calls in a sequence. To get the ith, you can use index notation: `Select.Record.app.somemethod[i]`. However, to make this change somewhat backwards compatible, if there is only one call, the index can be omitted. That is, `Select.Record.app.somemethod[0].rets...` is equivalent to `Select.Record.app.somemethod.rets...`. A warning will be given if there is more than one call and an index is not used to disambiguate them.


joshreini1 (changes requested):
Can you update the key setting method to the more explicit setting? This way it’s more easily runnable in Google colab.

trulens_eval/trulens_eval/app.py
```diff
@@ -588,19 +607,19 @@ def _get_method_path(self, obj: object, func: Callable) -> JSONPath:
...
+            logger.warn(
                 f"A new object of type {type(obj)} at 0x{id(obj):x} is calling an instrumented method {func}. "
                 "The path of this call may be incorrect."
             )
             try:
                 _id, f, path = next(iter(self._get_methods_for_func(func)))
             except Exception:
-                logger.debug(
+                logger.warn(
                     "No other objects use this function so cannot guess path."
```
joshreini1, line 618: Is guessing the best language to use here?

piotrm0: I think so, this point of the code is meant to guess a path (pick one without being sure it is correct).

piotrm0:
> Can you update the key setting method to the more explicit setting? This way it’s more easily runnable in Google colab.

I think I fixed. Let me know.

joshreini1 (approved):

PR #442, opened 2023-09-12
Rename tru_query_engine and add docs to tru_wrapper usages

rshih32:
* Now that wrappers have both call and record options, the wrapper should be more generically named.
* The recorder and calling wrapper cannot be partitioned into two modes as the name, feedback functions, and metadata still needs to be specified and is better in one place.
* Added documentation and code example updates to better show possible usages


rshih32:
Closing this; we have decided not to support call/query with at all, and the context manager as recorder is the only support we want

PR #451, opened 2023-09-19
Minor UI updates

walnutdust:
## Changes

**Before**

https://github.com/truera/trulens/assets/36098015/44488376-23c7-4a5b-ac15-53b501a56ae2

**After**

https://github.com/truera/trulens/assets/36098015/45ee4c16-2328-4078-9b91-73ab638563a3

**Details**
1. Added divider to better separate record selection and record details, also adjusted typography to emphasize that we're looking into the record
Screen Shot 2023-09-19 at 8 52 59 AM
2. Added record level metrics and how they compare to app averages
Screen Shot 2023-09-19 at 8 53 41 AM
3. Organized the input/response, metadata, and feedback to fit more items above the fold (within the same scren)
Screen Shot 2023-10-04 at 9 53 22 AM

4. Minor adjustments to typography to ensure selected component is given more weight over the subcomponents header. Also fixed bug where we were showing the subcomponents header even when there was no subcomponent
Screen Shot 2023-09-19 at 8 54 54 AM
5. Adjusted spacing between title and first component
Screen Shot 2023-09-19 at 8 56 06 AM
6. Adjusted the timeline component to make each row narrower
Screen Shot 2023-09-19 at 8 57 17 AM


walnutdust:
> I'm slightly concerned about the horizontal space for the feedback displays - especially as we focus more on chain of thought reasons since they'll be cut out unless scrolling right.

that's a great point! thanks for bringing that up - should I shift it to occupy the whole space below then? do you have thoughts on what items can be grouped together - maybe input/response and metadata? or should I remove that change?

joshreini1:
A few ideas:

- Moving the prompt and response back above but displaying them next to each other horizontally would give more space for feedbacks above the fold
- Often the prompt/response expanders are duplicative to the information in the feedbacks, depending on which feedback functions are used. Would it be reasonable to conditionally show the prompt/response only if they're not shown in the feedback display? Alternatively, there might be a better way to combine the two since there's overlap in the information shown.

rshih32:
Just validating @joshreini1's comment. We want to be able to explore the full extend of the feedbacks, that is very important.

@walnutdust maybe let us know if there are new updates to consider

walnutdust:
@rshih32 sounds good - was waiting on more eyes before working on it all in one shot. if this is the main feedback I'll touch up on this and ping you both for a re-review when I'm done

joshreini1 (approved):
This is great!

PR #453, opened 2023-09-20
run unit tests on PR

rshih32:
* Add to azure pipeline
* Set non-blocking daemon threads (currently only blocking thread is dashboard)

### DIFF
--- a/.azure_pipelines/ci-eval-pr.yaml
+++ b/.azure_pipelines/ci-eval-pr.yaml
@@ -19,6 +19,7 @@ jobs:
       matrix:
         default:
           condaEnvFileSuffix: "trulens-llm"
+          testSubdirectory: "tests/unit"
 
     steps:
       - checkout: self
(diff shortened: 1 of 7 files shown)


.azure_pipelines/ci-eval-pr.yaml
```diff
@@ -51,4 +52,22 @@ jobs:
             echo "\n\nOutputting git diff for checked in files.\n"
             git diff
           fi
-        displayName: NoDiffCheck
\ No newline at end of file
+        displayName: NoDiffCheck
+      - bash: |
+          source activate $(condaEnvFileSuffix)
+
+          pip install pytest==7.0.1 pytest-azurepipelines 'langchain>=0.0.230' 'llama_index>=0.8.3'
```
joshreini1, line 59: What's the plan for keeping these packages up to date?

trulens_eval/trulens_eval/tru.py
```diff
@@ -586,6 +589,11 @@ def listen_to_dashboard(proc: subprocess.Popen, pipe, out, started):
             target=listen_to_dashboard,
             args=(proc, proc.stderr, out_stderr, started)
         )
+
+        # Purposely block main process from ending and wait for dashboard. 
+        Tru.dashboard_listener_stdout.daemon=False
```
joshreini1, line 594: Why are we adding the wait for dashboard to the test? Are we worried about dashboard breakages?

.azure_pipelines/ci-eval-pr.yaml
rshih32: these are installed on fresh environment each time. so should be up to date.

trulens_eval/trulens_eval/tru.py
rshih32: no; this is the existing functionality. Just being explicit here. 
Also; this is not the test; this is the main branch. all threading changes are actual usage.

The main thing here is currently if you call a script with the context managers; the script will never end (This is true also for pytest which is why its in this PR)

However; if you run a script with tru.run_dashboard, without this; the script will end and the dashboard will be be killed with the running script

PR #465, opened 2023-09-23
async instrumentation bugfix and example update

piotrm0:
A bug prevent proper functioning of async call instrumentation. Fixed it and updated lang_chain async example to properly demonstrate async functionality that still captures records.


joshreini1 (approved):
Lgtm!

PR #475, opened 2023-10-03
Fix unicode output errors for windows

### LINKED ISSUE #396: [Bug] Missing Package and Unicode Encode Error
Hello, you forgot to import os :D 
image
And it seems that these symbols below only work with Streamlit. The Azure Function libs do not support encoding these unicode characters. When I replace them with nothing, my app works. Could you consider removing them? 
Capture

shayaks:
Fixes #396


trulens_eval/trulens_eval/utils/text.py
```diff
@@ -3,13 +3,19 @@
 """
 
 import logging
+import sys
 
 logger = logging.getLogger(__name__)
 
+
+if hasattr(sys.stdout, "reconfigure"):
+    sys.stdout.reconfigure(errors="replace")
```
rshih32, line 12: can you add an in-code comment on why this fixes the issue?

rshih32 (changes requested):

piotrm0 (approved):

Issue #470, opened 2023-09-27
sqlalchemy.exc.OperationalError: (psycopg2.OperationalError) server closed the connection unexpectedly


niloychak2022:
I am running in ubuntu 20.04 machine. I have integrated trulens with llama-index for quering. Here is the code - 

```python
from multiprocessing.managers import BaseManager
from trulens_eval import TruLlama, Feedback, Tru, feedback
from sqlalchemy import URL
from llama_index import SimpleWebPageReader
from llama_index import VectorStoreIndex
import numpy as np

def initialize_trulens():
    global feedbacks

    # Initiating the trulens dashboard
    url_object = URL.create(
        "postgresql+psycopg2",
        username=os.environ.get("SUPERBASE_USERNAME"),
        password=os.environ.get("SUPERBASE_PASSWORD"), 
        host=os.environ.get("SUPERBASE_HOST"),
        port=os.environ.get("SUPERBASE_PORT"),
        database=os.environ.get("SUPERBASE_DATABASE")
    )
    tru = Tru(database_url= url_object)
    tru.run_dashboard()

    # Initialize Huggingface-based feedback function collection class:
    hugs = feedback.Huggingface()
    openai = feedback.OpenAI()
    # Define a language match feedback function using HuggingFace.
    f_lang_match = Feedback(hugs.language_match).on_input_output()
    # By default this will check language match on the main app input and main app
    # output.

    # Question/answer relevance between overall question and answer.
    f_qa_relevance = Feedback(openai.relevance).on_input_output()

    # Question/statement relevance between question and each context chunk.
    f_qs_relevance = Feedback(openai.qs_relevance).on_input().on(
        TruLlama.select_source_nodes().node.text
    ).aggregate(np.min)

    feedbacks = [f_lang_match, f_qa_relevance, f_qs_relevance]

def query():
    documents = SimpleWebPageReader(html_to_text=True).load_data(
         ["http://paulgraham.com/worked.html"]
    )
    index = VectorStoreIndex.from_documents(documents)
    
    query_engine = index.as_query_engine()
    tru_query_engine_recorder = TruLlama(
        query_engine,
        app_id='LlamaIndex_App1',
        feedbacks=feedbacks
    )
    response = tru_query_engine_recorder.query("What did the author do growing up?")
    print(response)
    return response

    
if __name__ == "__main__":
    print("server started...")
    initialize_trulens()
    manager = BaseManager(('', 5602), b'password')
    manager.register('query_index', query)
[...]

PR #477, opened 2023-10-03
add sql params

rshih32:
Fixes https://github.com/truera/trulens/issues/470


rshih32:
> The S/O article looks like only pool_pre_ping=True is required; the other params are illustrative

The rest are more than illustrative: https://docs.sqlalchemy.org/en/13/core/pooling.html#disconnect-handling-pessimistic

Each section goes over where it can help; I did use the same ones as they seemed reasonable

joshreini1 (approved):

PR #501, opened 2023-10-16
Fix groundedness aggregation flakiness + incorrect 0 resolution.

joshreini1:
3 fixes:

- Ensure that source statements multi output is a list so it can be aggregated.
- When we included 0 into the eval range, 0s were being resolved incorrectly to -1 by Evaluations.py because they were seen as null. This also fixes that bug.
- groundedness measure was passing an extra arg (chain_of_thought) to _groundedness_doc_in_out that was causing errors.

In addition, move the renamed method warning from the Groundedness init to just the method that was renamed.


daniel-huang-1230 (approved):
lgtm

PR #504, opened 2023-10-17
handle no pii

joshreini1:
PII Detection was not returning a score when the HF API returned null. This fixes that bug returns a 1 in that case.


trulens_eval/trulens_eval/feedback/provider/hugs.py
```diff
@@ -365,10 +365,21 @@ def pii_detection_with_cot_reasons(self, text: str):
...
 
-        hf_response = self.endpoint.post(
-            url = HUGS_PII_DETECTION_API_URL,
-            payload = payload)
-        
+        try:
+            hf_response = self.endpoint.post(
+                url = HUGS_PII_DETECTION_API_URL,
+                payload = payload)
+        except Exception as e:
```
ejisoo, line 372: Does it raise an exception? What kind? If it returns nothing because there is no PII, I wonder if that would be an exception..

joshreini1: Yes, in main you get the following exception:

```
Feedback function pii_detection_with_cot_reasons with aggregation <function mean at 0x118d99760> had no inputs.
Evaluation of pii_detection_with_cot_reasons failed on inputs: 
{'text': 'I like dogs'}
Post did not return a sequence: [].
```

ejisoo: Looking at the code [here](https://github.com/truera/trulens/blob/main/trulens_eval/trulens_eval/feedback/provider/endpoint/base.py#L174-L176), I think it would be an `AssertionError` if the response was empty because of no PII?

I'm thinking if there is actually PII but if we raise [`RuntimeError`](https://github.com/truera/trulens/blob/main/trulens_eval/trulens_eval/feedback/provider/endpoint/base.py#L172) or something else, we might incorrectly return the zero-score response (false negative)?

joshreini1: I think that is what's happening because the message matches?

> ooking at the code [here](https://github.com/truera/trulens/blob/main/trulens_eval/trulens_eval/feedback/provider/endpoint/base.py#L174-L176), I think it would be an `AssertionError` if the response was empty because of no PII?

I take your point that false negatives are costly for PII detection, but don't think we want to raise an error in cases of no pii because that will hopefully be the majority of records.

ejisoo: Oh I was referring to our `Endpoint`'s `post` that does some error handling. I don't suggest raising a new exception. I was pointing out if we catch all exceptions, it _could_ catch any of the exceptions in `requests.exceptions` plus `RuntimeError` that we could raise before the assertion and assign 0 score. Btw I kept this as a comment rather than a review, so it doesn't block. But everything else looks good to me!

daniel-huang-1230 (approved):
~Given we don't have logging framework in place yet~ - could we add a TODO comment on the exception catching so that we know the current behavior will swallow any exception and not fail fast? 

Could we also add a line to log with error msg
```logger.error(...)```

PR #522, opened 2023-10-25
example appui in quickstart

piotrm0:


jimwhite:
I've been trying to run the experimental appui examples and have finally realized that the version in main can't work without changes in serialization.  Looks like this PR was a fix for that but is no longer compatible.  Is there a working branch somewhere?  Before I try and fix this myself is there anything else I should know?

PR #523, opened 2023-10-25
dynamic leaderboard

joshreini1:
Allow leaderboard columns to only display feedbacks present for a given app id.

Before:
Screen Shot 2023-10-25 at 3 25 35 PM

After:
Screen Shot 2023-10-25 at 3 21 50 PM

### DIFF
--- a/trulens_eval/trulens_eval/Leaderboard.py
+++ b/trulens_eval/trulens_eval/Leaderboard.py
@@ -62,8 +62,9 @@ def streamlit_app():
         metadata = app_json.get('metadata')
         #st.text('Metadata' + str(metadata))
         st.header(app, help=draw_metadata(metadata))
+        app_feedback_col_names = [col_name for col_name in feedback_col_names if not app_df[col_name].isna().all()]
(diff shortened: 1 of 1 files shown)


walnutdust (approved):
LGTM; only comment is that if we anticipate users having this many feedback functions we might want to find some other way to visualize it - I'll try to see if I can find the time to take up that work

joshreini1:
> LGTM; only comment is that if we anticipate users having this many feedback functions we might want to find some other way to visualize it - I'll try to see if I can find the time to take up that work

Thanks @walnutdust ! Putting this into the new demo flow so we can show a more comprehensive view of what TruLens has out of the box

shayaks:
I'm a bit late to this, but do we make sure that columns are sorted in the right order? So that if two apps have the same/similar columns, they line up?

joshreini1:
The order of the columns will be the same across app_ids. The change is
that if a particular app id does not use a feedback, that will not show as
an empty column.

Ex:

Before:

APP 1: Feedback A |                      | Feedback B
APP 2:                      | Feedback C |

After:

APP 1: Feedback A | Feedback B
APP 2: Feedback C

PR #524, opened 2023-10-25
Improve cot reasoning

joshreini1:
Add prompting to influence llm to tie reasons back to the evaluation being performed.

Results in more consistent/helpful cot reasons.

Examples:
Screen Shot 2023-10-27 at 2 54 51 PM
Screen Shot 2023-10-27 at 2 55 00 PM

Additional changes
- Fixed docstring for model_agreement.


trulens_eval/trulens_eval/feedback/provider/base.py
```diff
@@ -142,11 +142,17 @@ def _extract_score_and_reasons_from_response(
         response = self.endpoint.run_me(
             lambda: self._create_chat_completion(messages=llm_messages)
         )
-        if "Supporting Evidence" in response:
+        if "Supporting Evidence" or "Reason" in response:
```
daniel-huang-1230, line 145: are we still preserving the "Supporting Evidence" keyword in COT prompt - seems like we now inject it in summarization & groundedness but no longer in COT prompt? 

if so, could you update the docstring above in `_extract_score_and_reasons_from_response`?

ejisoo (approved):
Nice to have the criteria and evidence broken down!

PR #529, opened 2023-10-31
fix litellm provider

joshreini1:
Changes:

- Explicitly pass through model engine to litellm.completion
- Explicitly extract the response from the openai object output.
- Standardize allowance of prompt or message input with openai provider


shayaks (commented):
Which providers has this been tested with? 
(Have we tested openai/azure openai/anthropic)?

joshreini1:
> Which providers has this been tested with? (Have we tested openai/azure openai/anthropic)?

Tested with Cohere, Anthropic and Together AI (hosted llama-2). 

LiteLLM seems to have made some changes on their end that broke our existing integration. This change should make it more stable.

joshreini1:
Confirmed this works as well with openai, azureopenai.

For Azure, you just need to add the required to your provider instantiation.

Example:

litellm = LiteLLM(model_engine="azure/<deployment name>",
                  api_base = os.environ["AZURE_API_BASE"],
                  api_version = os.environ["AZURE_API_VERSION"],
                  api_key = os.environ["AZURE_API_KEY"])

ejisoo (approved):

PR #531, opened 2023-11-01
Anthropic Integration

joshreini1:
- Show use of Anthropic Claude as feedback function
- Allow use of any LLM for groundedness
- Remove need for summarize provider in groundedness measure

### DIFF
--- a/trulens_eval/trulens_eval/feedback/groundedness.py
+++ b/trulens_eval/trulens_eval/feedback/groundedness.py
@@ -9,6 +9,8 @@ from trulens_eval.feedback.provider import Provider
 from trulens_eval.feedback.provider.hugs import Huggingface
 from trulens_eval.feedback.provider.openai import AzureOpenAI
 from trulens_eval.feedback.provider.openai import OpenAI
+from trulens_eval.feedback.provider.litellm import LiteLLM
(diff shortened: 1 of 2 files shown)


trulens_eval/trulens_eval/feedback/groundedness.py
```diff
@@ -97,9 +90,9 @@ def groundedness_measure(self, source: str, statement: str) -> float:
         )
 
         groundedness_scores = {}
-        if isinstance(self.groundedness_provider, (AzureOpenAI, OpenAI)):
+        if isinstance(self.groundedness_provider, (AzureOpenAI, OpenAI, LiteLLM, Bedrock)):
```
ejisoo, line 93: Are we missing Hugging Face here? Or do we want LiteLLM to handle it?

joshreini1: LiteLLM will handle LLMs from huggingface. If huggingface is set directly as the provider, we'll use the NLI model from hf.

ejisoo: Oops I was only looking at the file changes. LGTM.

ejisoo (approved):

PR #532, opened 2023-11-01
[MLNN-1077] Groundedness score extraction bug

daniel-huang-1230:
Issue:  our score extraction in the partial match cases is failing on any LLM-generated output with group of digits **appearing before** the final `"Score: 10"` text.

Before the fix:
image

After the fix:
image

### DIFF
--- a/trulens_eval/trulens_eval/utils/generated.py
+++ b/trulens_eval/trulens_eval/utils/generated.py
@@ -7,14 +7,17 @@ import re
 
 logger = logging.getLogger(__name__)
 
-pat_0_10 = re.compile(r"\s*([0-9][0-9]*)\s*")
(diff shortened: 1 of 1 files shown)


ejisoo:
@daniel-huang-1230 Thanks for catching & fixing this. Shouldn't we have a test case for utils like this?

joshreini1 (approved):

ejisoo (approved):

daniel-huang-1230:
> @daniel-huang-1230 Thanks for catching & fixing this. Shouldn't we have a test case for utils like this?

yes makes a lot of sense - I should've added some unit tests for this and some other similar utilities in this PR. Will create a jira ticket to track

PR #533, opened 2023-11-01
Fix small bug with feedback direction

joshreini1:
If "higher_is_better" not added to db, set the default on both Leaderboard and Evaluations page.


ejisoo (approved):
Not sure if there are currently other ways than wrapping a logic with `Feedback` to define a feedback function, but I think this is a good idea (e.g., old apps the db migration script happens to not handle).

PR #547, opened 2023-11-08
fix nav background color on mobile (MLNN-1095)

joshreini1:
Before:
Screen Shot 2023-11-08 at 8 14 57 AM

After:
Screen Shot 2023-11-08 at 9 54 47 AM

### DIFF
--- a/docs/stylesheets/style.css
+++ b/docs/stylesheets/style.css
@@ -632,13 +632,12 @@ p {
 @media only screen and (max-width: 768px) {
   .nav__items {
     flex-direction: column;
-    display: flex;
     position: fixed;
-    top: 0;
+    top: calc(calc(var(--gap-lg) * 2 + 32px) + 0.5rem);
     left: 0;
     width: 100%;
     height: min-content;
-    background-color: var(--color-navigation) !important;
+    background-color: var(--color-navigation);
     box-shadow: 0px 4px 8px rgba(0, 0, 0, 0.08);
     list-style-type: none;
     -webkit-font-smoothing: antialiased;
--- a/docs/stylesheets/style.css
+++ b/docs/stylesheets/style.css
@@ -632,11 +632,13 @@ p {
 @media only screen and (max-width: 768px) {
   .nav__items {
     flex-direction: column;
+    display: flex;
     position: fixed;
-    top: calc(calc(var(--gap-lg) * 2 + 32px) + 0.5rem);
+    top: 0;
     left: 0;
-    width: 100vw;
-    background-color: var(--color-navigation);
+    width: 100%;
+    height: min-content;
+    background-color: var(--color-navigation) !important;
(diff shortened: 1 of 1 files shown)


docs/stylesheets/style.css
```diff
@@ -632,11 +632,13 @@ p {
 @media only screen and (max-width: 768px) {
   .nav__items {
     flex-direction: column;
+    display: flex;
```
walnutdust, line 635: this is probably not needed - will be inherited from line 617 if not overwritten by default!

docs/stylesheets/style.css
```diff
@@ -632,11 +632,13 @@ p {
...
     position: fixed;
-    top: calc(calc(var(--gap-lg) * 2 + 32px) + 0.5rem);
+    top: 0;
```
walnutdust, line 637: @joshreini1 this variable should be controlling where the first nav item starts to allow our logo and the close icon to be shown - I think we might want to keep it, but wanted to confirm with you

docs/stylesheets/style.css
```diff
@@ -632,11 +632,13 @@ p {
...
     left: 0;
-    width: 100vw;
-    background-color: var(--color-navigation);
+    width: 100%;
```
walnutdust, line 639: in this context there isn't a difference in 100vw (100 % of viewport width) vs 100%, so it's okay to keep it as before

docs/stylesheets/style.css
```diff
@@ -632,11 +632,13 @@ p {
...
+    height: min-content;
+    background-color: var(--color-navigation) !important;
```
walnutdust, line 641: generally avoid !important in CSS if possible, this causes headaches when trying to debug and figuring out the precedence of the lines

walnutdust (changes requested):
@joshreini1  thanks for taking a stab at this - it looks like the only change that needs to be done is adding `height: min-content` to achieve our intended goal, could we see if the result of that change would meet our needs?

docs/stylesheets/style.css
joshreini1: With the viewport still there, I'm seeing this width in my mobile simulator (https://www.webmobilefirst.com/en/). 
Screen Shot 2023-11-08 at 11 38 49 AM

docs/stylesheets/style.css
joshreini1: Tested and removed - thanks!

docs/stylesheets/style.css
joshreini1: Thanks! Tested and removed this change.

docs/stylesheets/style.css
joshreini1: Thanks! Tested and removed this change

docs/stylesheets/style.css
walnutdust: oh interesting - then yes let's keep it

walnutdust (approved):

Issue #546, opened 2023-11-07
OpenAI error on BasicApp


lopezjuanma96:
I'm testing out monitoring using the TruBasicApp.

I've already implemented a pretty complex call on my Windows PC, but when trying to upload it to a Linux-based server on AWS it is raising an error with the `OpenAIEndpoint`, that raises at the *costs* section.

This is the code I'm trying to run:

```py
import os
import dotenv

from trulens_eval import Tru, TruBasicApp, Feedback
from trulens_eval.schema import Cost, Record

SCRIPT_DIR = os.path.abspath(os.path.dirname(__file__))
dotenv.load_dotenv(os.path.join(SCRIPT_DIR, "my.env"))

tru = Tru(database_url=os.environ.get("database_url"))

def llm_standalone(prompt):
    return prompt

recorder = TruBasicApp(llm_standalone, app="default")

with recorder as recording:
    response = recorder.app("Hola pa")

print(response)
```

As I said, I tried keeping it as simple as I could, and it was working for me on my Windows PC, but on the AWS setup I got this error:

```
Traceback (most recent call last):
  File "/path/to/my/file/trulens.py", line 125, in <module>
    tru.add_record(
  File "/opt/conda/lib/python3.9/site-packages/trulens_eval/app.py", line 700, in __exit__
    raise exc_value
  File "/path/to/my/file/trulens.py", line 115, in <module>
    response = recorder.app(prompt, max_tokens, stop_sequences)
  File "/opt/conda/lib/python3.9/site-packages/trulens_eval/tru_basic_app.py", line 51, in __call__
    return self._call(*args, **kwargs)
  File "/opt/conda/lib/python3.9/site-packages/trulens_eval/instruments.py", line 783, in wrapper
    ctx.app._on_add_record(
  File "/opt/conda/lib/python3.9/site-packages/trulens_eval/app.py", line 748, in _on_add_record
    raise error
  File "/opt/conda/lib/python3.9/site-packages/trulens_eval/instruments.py", line 734, in wrapper
    rets, cost = Endpoint.track_all_costs_tally(
  File "/opt/conda/lib/python3.9/site-packages/trulens_eval/feedback/provider/endpoint/base.py", line 374, in track_all_costs_tally
    result, cbs = Endpoint.track_all_costs(
  File "/opt/conda/lib/python3.9/site-packages/trulens_eval/feedback/provider/endpoint/base.py", line 265, in track_all_costs
    e = OpenAIEndpoint()
[...]

PR #555, opened 2023-11-10
OpenAI v1.x migraation + robust hasattr for instrumentation issues

piotrm0:
Some objects/classes we instrument override getattr or related which causes some instrumentation issues on our end. This is an attempt at making our process more robust to this. Might address https://github.com/truera/trulens/issues/546 .

Also includes small bugfixes found during the above work.


trulens_eval/trulens_eval/feedback/provider/endpoint/base.py
```diff
@@ -224,22 +225,29 @@ def _instrument_module(self, mod: ModuleType, method_name: str) -> None:
...
             )
             func = getattr(cls, method_name)
             w = self.wrap_function(func)
             setattr(cls, method_name, w)
 
-    def _instrument_module_members(self, mod: ModuleType, method_name: str):
-        logger.debug(
-            f"Instrumenting {mod.__package__}.*.{method_name} for {self.name}"
-        )
+    # TODO: Remove
```
ejisoo, line 236: Do we need this comment block?

trulens_eval/trulens_eval/feedback/provider/openai.py
```diff
@@ -3,12 +3,14 @@
 from typing import Dict, Mapping, Optional, Sequence
 
 import openai
+from openai import OpenAI
+
+client = OpenAI()
```
ejisoo, line 8: Is there a reason we do this here? What if the user has multiple API keys they want to use for different apps?

ejisoo: May be a little more involved to make changes, but it may be better if something (not sure what's the best class to implement it among OpenAI, OpenAI endpoint, etc.) accepted the OpenAI client instance.

trulens_eval/trulens_eval/database/sqlalchemy_db.py
```diff
@@ -171,6 +171,10 @@ def insert_app(self, app: schema.AppDefinition) -> schema.AppID:
         with self.Session.begin() as session:
             if _app := session.query(orm.AppDefinition
                                     ).filter_by(app_id=app.app_id).first():
+                
+                j = app.json()
+                print("inserting json: ", j)
```
shayaks, line 176: Is this print statement supposed to be here?

shayaks (approved):

trulens_eval/trulens_eval/feedback/provider/endpoint/base.py
joshreini1: We can remove. This was a dev method we ended up not using. Nice catch.

trulens_eval/trulens_eval/feedback/provider/openai.py
joshreini1: I think we can also remove this.

joshreini1 (changes requested):

joshreini1 (approved):

PR #601, opened 2023-11-29
Add `quickstart.ipynb` and other notebooks to notebooks to test.

DavidKurokawa:


joshreini1:
We claim (I guess incorrectly) that we're maintaining all the notebooks in quickstart/ are maintained with every release. To support that, let's symlink the rest of them for testing here too.

joshreini1 (approved):

DavidKurokawa:
> We claim (I guess incorrectly) that we're maintaining all the notebooks in quickstart/ are maintained with every release. To support that, let's symlink the rest of them for testing here too.

Done.

PR #623, opened 2023-12-01
Fix > character in prompt

NoahVl:
This PR changes the `.` at the end of  `LLM_GROUNDEDNESS_FULL_SYSTEM` to a `>`, this is more in line with other template prompts such as `COT_REASONS_TEMPLATE`.

Please let me know if the full stop was there on purpose!


joshreini1 (approved):
Thanks for fixing @NoahVl !

PR #637, opened 2023-12-05
Fix llama index agents and multimodal notebooks

shayaks:


piotrm0:
Are you including notebook outputs intentionally?

shayaks:
> Are you including notebook outputs intentionally?

For these expositional notebooks, it might be good to have some output content showing what the output of the app looks like. I can remove if you'd like.

PR #640, opened 2023-12-05
Fix: AzureOpenAI

schmidtseb:
The module currently doesn't work with AzureOpenAI. With the new `openai` version (>1), there is no requirement for providing `deployment_id`.
This also fixes an issue by providing the `deployment_name` to the `openai.AzureOpenAI`-object.
Provide credentials using environment variables or `kwargs`, i.e.:

```
AZURE_OPENAI_API_KEY
OPENAI_API_VERSION
AZURE_OPENAI_ENDPOINT
```

Solution feels a bit hacky, but doing it in a proper way would probably require a refactoring


daniel-huang-1230 (approved):
LGTM 👍 thanks for putting up the quick fix!

PR #663, opened 2023-12-09
required packages and imports cleanup

piotrm0:
- Removed optional packages from installation requirements. The required list is now much shorter.
- Made sure all requirements have a lower bound.
- Moved required packages out of `setup.py` and into `trulens_eval/requirements.txt` and made this file be included in an installed package folder.
- Moved all optional packages to `trulens_eval/requirements.optional.txt` and made this be included in installation as well.
- Organized the two requirement files. The optional one still has some unorganized packages which I don't remember the purpose of.
- Changed our optional importing scheme to refer to package requirements from these files instead of having to hard code the constants.
- Adjusted CI pipelines to install the optional packages from the ...optional.txt file instead of hardcoded.


joshreini1 (approved):
Hugely impactful. Thanks for making this change

PR #681, opened 2023-12-15
Josh/release nits

joshreini1:
Remove cohere as a provider - this was a very early feature that has proved to be not useful. 

Fix header on Huggingface docs page.


daniel-huang-1230 (approved):
lgtm

PR #698, opened 2023-12-19
exposed AzureOpenAI provider

epinzur:
Exposed the AzureOpenAI provider so I can use it to power feedback functions.

### DIFF
--- a/trulens_eval/trulens_eval/feedback/provider/openai.py
+++ b/trulens_eval/trulens_eval/feedback/provider/openai.py
@@ -366,26 +366,25 @@ class AzureOpenAI(OpenAI):
     Has the same functionality as OpenAI out of the box feedback functions.
     """
 
-    def __init__(self, endpoint=None, deployment_name="gpt-35-turbo", **kwargs):
+    def __init__(self, deployment_name: str, endpoint=None, **kwargs):
         # NOTE(piotrm): pydantic adds endpoint to the signature of this
         # constructor if we don't include it explicitly, even though we set it
         # down below. Adding it as None here as a temporary hack.
         """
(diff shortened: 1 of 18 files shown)


epinzur:
I guess I should create docs for this too.  I'll work on that.

joshreini1:
Thanks @epinzur !

trulens_eval/trulens_eval/feedback/provider/openai.py
```diff
@@ -366,26 +366,25 @@ class AzureOpenAI(OpenAI):
     Has the same functionality as OpenAI out of the box feedback functions.
     """
 
-    def __init__(self, endpoint=None, deployment_name="gpt-35-turbo", **kwargs):
+    def __init__(self, deployment_name: str, endpoint=None, **kwargs):
```
epinzur, line 369: I dropped the default here since model deployment names in Azure can be anything. I can deploy `gpt-3.5-turbo` as `open-ai`.

epinzur:
I've tested this locally by just copying the AzureOpenAI class into my notebook.

trulens_eval/trulens_eval/feedback/provider/openai.py
piotrm0: Is there a typical default that we can still use?

epinzur: I don't think so. There are no models deployed by default. You need to pick and choose which to deploy and choose the name at deployment time. The deployment name often doesn't match the model name due to Azure naming restrictions.

joshreini1 (approved):

PR #701, opened 2023-12-20
[MLNN-1121] Context relevance benchmark as recommendation task 

daniel-huang-1230:
TODO: 
1.~~figure rate limiting issue with multiple model-as-service providers and run the actual eval with some backoff~~
2. add recall metrics

follow-up: implement similar benchmark for answer relevance


trulens_eval/trulens_eval/tests/context_relevance_benchmark.ipynb
```diff
@@ -0,0 +1,297 @@
+{
```
joshreini1, line 1: <pre class="ql-syntax" spellcheck="false">Line #5.    from benchmark_frameworks.eval_as_recommendation import score_passages, compute_ndcg, compute_ece, precision_at_k
</pre><p>Can you add some info here on why these metrics are chosen and short explainers on how they are computed?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/truera/trulens/pull/701/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

PR #706, opened 2023-12-20
Add shortcut to select_context()

joshreini1:
Motivation:

Make it easier for users building and evaluating RAGs to select context.

Langchain and Llama-Index users follow the same context selection path:

```python
from trulens_eval.app import App
context = App.select_context(app)
```

This is useful when doing evals such as context relevance, groundedness, etc. by dramatically shortening the selector required + has the added bonus of unification across app frameworks.


trulens_eval/examples/quickstart/langchain_quickstart.ipynb
```diff
@@ -189,7 +189,7 @@
     "# Define a groundedness feedback function\n",
     "f_groundedness = (\n",
     "    Feedback(grounded.groundedness_measure_with_cot_reasons)\n",
-    "    .on(Select.RecordCalls.first.invoke.rets.context)\n",
+    "    .on(TruChain.select_context().collect())\n",
```
epinzur, line 192: why does this one use `...collect()` and the next not, even though the old versions of both are the same?

epinzur:
I think this is great!

trulens_eval/trulens_eval/tru_chain.py
```diff
@@ -210,6 +213,13 @@ def __init__(self, app: Chain, **kwargs):
 
         super().__init__(**kwargs)
 
+    @classmethod
+    def select_context(cls) -> Lens:
+        """
+        Get the path to the context in the query output.
+        """
+        return Select.RecordCalls.first.invoke.rets.context
```
piotrm0, line 221: "first" is specific to `RunnableSequence`.

trulens_eval/trulens_eval/tru_llama.py
```diff
@@ -251,6 +251,13 @@ def select_source_nodes(cls) -> Lens:
         """
         return cls.select_outputs().source_nodes[:]
 
+    @classmethod
+    def select_context(cls) -> Lens:
+        """
+        Get the path to the context in the query output.
+        """
+        return cls.select_outputs().source_nodes[:].node.text
```
piotrm0, line 259: reuse `select_source_nodes` here?

joshreini1:
> As noted above, the selector for langchain select_context is specific to the langchain app. One option to fix this is to require the user to provide the app when calling select_context which is a bit annoying but doable. I made a PR to merge into this PR's branch the example of how to do this but had to make some bug fixes in other code to make sure the `VectorStoreRetriever` calls get captured in the new examples. Please take a look and test. See #711 .

Can we get out of passing in the app by using the context from all get_relevant_documents() methods instrumented?

piotrm0:
> Can we get out of passing in the app by using the context from all get_relevant_documents() methods instrumented?

I don't think there is a way right now to write a selector to do this given the structure of records. I'm thinking of some additions to selector syntax/semantics that would make it possible.

piotrm0:
Added `App.select_context` and `Select.context` to be able to use the same method regardless of use of langchain or llama index. Please review.

joshreini1:
> @joshreini1 Added `App.select_context` and `Select.context` to be able to use the same method regardless of use of langchain or llama index. Please review.

This is good - updated the quickstarts to show usage. The symmetry between langchain/llama-index context selection will go along way in reducing confusion and friction.

trulens_eval/examples/quickstart/langchain_quickstart.ipynb
joshreini1: `.collect()` is used fo groundedness to return all of the context here as an array. This is needed for groundedness because we need to identify supporting evidence from all of the context available.

This contrasts with context_relevance where we want to evaluate the relevance of each chunk individually.

trulens_eval/trulens_eval/tru_llama.py
joshreini1: Is there any benefit of keeping this? Or just remove it in place of app.select_context()

piotrm0 (approved):

Issue #707, opened 2023-12-21
Example google collab cell has a error when running


kleenkanteen:
In the TruLens quickstart here: https://www.trulens.org/trulens_eval/quickstart/

For the collab:

![image](https://github.com/truera/trulens/assets/67878058/1199ebf7-e6fc-4eb4-8567-564b54a70a21)

I uncommented and ran the first line: 
![image](https://github.com/truera/trulens/assets/67878058/ea2f538b-bfbf-4cd3-bf13-af640e57b602)

Ran all the cells until this cell, and then got this error:

![image](https://github.com/truera/trulens/assets/67878058/559d1c3e-81cc-4724-95e3-f884fe20fa93)

The solution is to add llama_index to the very first cell: 
![image](https://github.com/truera/trulens/assets/67878058/0d556f3d-aa80-451b-b28d-3ef1fbeac303)

Now the cell runs fine:
![image](https://github.com/truera/trulens/assets/67878058/9479ee03-0906-463f-bdca-f6853b8fe174)

joshreini1:
We should not require llama-index to import trulens here. Will take a look at the dependencies.

joshreini1:
Added PR #714 that should fix this.

thisiskeithkwan:
> Added PR #714 that should fix this.

Same problem different
Today on the latest version of trulens_eval

<img width="860" alt="image" src="https://github.com/truera/trulens/assets/26019476/c69788eb-34b6-480c-9c66-56fc17bd6f3f">

ingridstevens:
This is still a problem. The fix I'm using is: 
`pip install llama-index==0.9.21`

joshreini1:
Same issue as https://github.com/truera/trulens/issues/731

joshreini1:
Please upgrade your trulens version if you are still experiencing this issue. Thanks!

joshreini1:
Let me create a new issue for this litellm requirement. Will tag you

PR #708, opened 2023-12-21
Add llama_index dependency for quickstart collab example

kleenkanteen:
Fixes the problem in #707


joshreini1:
Fixing with #714 instead. Llama-Index should not be a requirement.

PR #709, opened 2023-12-21
More secure way of storing API key by using getpass standard library

kleenkanteen:
In the case someone makes a copy of the tutorial, they don't have to hard code the api key


joshreini1:
Thanks @kleenkanteen !

We intentionally are having this key setting in the tutorial to make it easy for people to get started in different environments (colab, etc.). Btw, see: 

`from trulens_eval.keys import check_or_set_keys`

PR #710, opened 2023-12-21
Add code to show IP address of the collab instance so dashboard can be viewed

kleenkanteen:
Without this, when you run the very last cell, the output says to input the IP but you do not know where the IP is from:

image

My change prints it like this:

image


joshreini1:
Thanks @kleenkanteen !

To confirm, are you experiencing this issue in google colab?

kleenkanteen:
Yes

piotrm0:
Not showing the external IP was a bug that was introduced in last release and will be fixed in the next release.

PR #714, opened 2023-12-21
import llama only if needed

joshreini1:
Address: https://github.com/truera/trulens/issues/707


piotrm0:
@joshreini1 I adjusted this branch to use OptionalImports; can you check whether this works for your purposes?

joshreini1:
> @joshreini1 I adjusted this branch to use OptionalImports; can you check whether this works for your purposes?

Seems to be a few other things in json.py that are requiring optional packages when importing Tru. Seeing httpx, openai (Timeout) so far.

shayaks (approved):

Issue #716, opened 2023-12-21
Getting While Running provider.summary_with_cot_reasons feedback with Bedrock as Provider


SanjuMLGeek:
Hello,

I am trying to evaluate Summary response using below code. Prompt is reduced to keep small intentionally to keep this description short. I tried getting into the Bedrock provider code and debug but I don't understand why it throws this error message from an internal assertion? 

Appreciate any help you can extend.

**Error:**

`_
> bedrock request failed <class 'AssertionError'>=Bedrock can only operate on 'prompt', not 'messages'.. Retries remaining=3.
> bedrock request failed <class 'AssertionError'>=Bedrock can only operate on 'prompt', not 'messages'.. Retries remaining=2.
> bedrock request failed <class 'AssertionError'>=Bedrock can only operate on 'prompt', not 'messages'.. Retries remaining=1.
> bedrock request failed <class 'AssertionError'>=Bedrock can only operate on 'prompt', not 'messages'.. Retries remaining=0.
> Feedback Function exception caught: Traceback (most recent call last):
>   File "/opt/conda/lib/python3.10/site-packages/trulens_eval/feedback/feedback.py", line 501, in run
>     result_and_meta, part_cost = Endpoint.track_all_costs_tally(
>   File "/opt/conda/lib/python3.10/site-packages/trulens_eval/feedback/provider/endpoint/base.py", line 447, in track_all_costs_tally
>     result, cbs = Endpoint.track_all_costs(
>   File "/opt/conda/lib/python3.10/site-packages/trulens_eval/feedback/provider/endpoint/base.py", line 396, in track_all_costs
>     return Endpoint._track_costs(thunk, with_endpoints=endpoints)
>   File "/opt/conda/lib/python3.10/site-packages/trulens_eval/feedback/provider/endpoint/base.py", line 553, in _track_costs
>     result: T = thunk()
>   File "/opt/conda/lib/python3.10/site-packages/trulens_eval/feedback/feedback.py", line 502, in <lambda>
>     lambda: self.imp(**ins)
>   File "/opt/conda/lib/python3.10/site-packages/trulens_eval/feedback/provider/base.py", line 1013, in summary_with_cot_reasons
>     return self._extract_score_and_reasons_from_response(system_prompt)
>   File "/opt/conda/lib/python3.10/site-packages/trulens_eval/feedback/provider/base.py", line 148, in _extract_score_and_reasons_from_response
>     response = self.endpoint.run_me(
[...]

Issue #609, opened 2023-11-30
question: Bedrock


arm-diaz:
How can i run Bedrock with the latest version of trulens? The jupyter notebook example does not work.

Code:
```
# Initialize Huggingface-based feedback function collection class:
bedrock = Bedrock(model_engine="Bedrock", model_id = "anthropic.claude-v2", region_name="us-west-2")

# Define a language match feedback function using HuggingFace.
f_relevance = Feedback(bedrock.relevance).on_input_output()
# By default this will check language match on the main app input and main app
# output.

tru_recorder = TruChain(chain,
    app_id='Chain1_ChatApplication',
    feedbacks=[f_relevance])


with tru_recorder as recording:
    llm_response = chain.run("What's the capital of the USA?")
display(llm_response)
```

Error:
```
OpenAIError: The api_key client option must be set either by passing api_key to the client or by setting the OPENAI_API_KEY environment variable
```

piotrm0:
I believe recent fixes and additions to BedrockEndpoint should address this problem. @arm-diaz : can you try using the trulens from github and see if you are still seeing the issue? (a release with the fixes should be available on pip later this week).

arm-diaz:
Hey, how can I install the library without the release?

I usually execute:

pip install git+https://github.com/truera/trulens.git

But i am getting an error that says,` https://github.com/truera/trulens.git does not appear to be a Python project: neither 'setup.py' nor 'pyproject.toml' found.`

joshreini1:
@arm-diaz You'll need to clone trulens and install from the trulens/trulens_eval directory.

```
git clone https://github.com/truera/trulens.git
cd trulens/trulens_eval
pip install -e .
```

arm-diaz:
Hi,

Is there a good reason for the bedrock endpoint name to be setup as "bedrock"?
Check link below:
https://github.com/truera/trulens/blob/6a41f2a1d201ded7c5d8add31d7cade2c2d51ed8/trulens_eval/trulens_eval/feedback/provider/endpoint/bedrock.py#L139

if I change it to bedrock-runtime, it works fine. otherwise, it gives me the following error:

```
bedrock request failed <class 'AttributeError'>='NoneType' object has no attribute 'invoke_model'. Retries remaining=3.
bedrock request failed <class 'AttributeError'>='NoneType' object has no attribute 'invoke_model'. Retries remaining=2.
[...]

PR #732, opened 2023-12-27
Update extract_score_and_reasons to work across providers

joshreini1:
Update structured extraction and COT prompting for better results across providers e.g. Bedrock, Ollama

Fixes: https://github.com/truera/trulens/issues/716 and open email with amazon team, possibly: https://github.com/truera/trulens/issues/609

### DIFF
--- a/trulens_eval/trulens_eval/feedback/provider/bedrock.py
+++ b/trulens_eval/trulens_eval/feedback/provider/bedrock.py
@@ -1,12 +1,10 @@
 import json
 import logging
 import os
-from typing import Dict, Optional, Sequence
+from typing import Dict, Optional, Sequence, Union, Tuple
 
-from trulens_eval.feedback import prompts
 from trulens_eval.feedback.provider.base import LLMProvider
 from trulens_eval.feedback.provider.endpoint import BedrockEndpoint
-from trulens_eval.feedback.provider.endpoint.base import Endpoint
 from trulens_eval.utils.generated import re_0_10_rating
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


piotrm0:
I think the more structured messages should be preferable if available so I'm not sure about just using prompts always.

joshreini1:
> I think the more structured messages should be preferable if available so I'm not sure about just using prompts always.

Good point - updated to use prompt only for bedrock.

trulens_eval/trulens_eval/feedback/provider/bedrock.py
```diff
@@ -63,17 +61,64 @@ def _create_chat_completion(
         assert self.endpoint is not None
         assert prompt is not None, "Bedrock can only operate on `prompt`, not `messages`."
 
-        # NOTE(joshr): only tested with sso auth
         import json
 
-        body = json.dumps({"inputText": prompt})
+        body = json.dumps({"inputText": prompt, "textGenerationConfig": {
+                "maxTokenCount": 4096,
```
piotrm0, line 67: Make a task to make these parameters configurable in a future PR?

piotrm0 (approved):

trulens_eval/trulens_eval/feedback/provider/bedrock.py
piotrm0: @joshreini1 approved but see this note.

joshreini1: Added a TODO for this

Issue #764, opened 2024-01-05
TruLlama with query engine argument not working


danielaguilera-aily:
```
from llama_index.indices.postprocessor import SentenceTransformerRerank
from llama_index.retrievers import AutoMergingRetriever
from llama_index.query_engine import RetrieverQueryEngine

automerging_retriever = automerging_index.as_retriever(
    similarity_top_k=12,
)

retriever = AutoMergingRetriever(
    automerging_retriever, 
    automerging_index.storage_context, 
    verbose=True,
)

rerank = SentenceTransformerRerank(top_n=6, model="BAAI/bge-reranker-base")

from llama_index.prompts import PromptTemplate


qa_template = PromptTemplate(prompt_temp, template_var_mappings={"query_str": "question", "context_str": "context"})


auto_merging_engine = RetrieverQueryEngine.from_args(
    retriever=retriever, service_context=auto_merging_context, node_postprocessors=[rerank]
)
```

```
def get_prebuilt_trulens_recorder(query_engine, app_id):
    from trulens_eval import TruLlama
    
    bedrock = get_trulens_bedrock()
    
    qa_relevance = Feedback(bedrock.relevance_with_cot_reasons, name="Answer Relevance").on_input_output()

    feedbacks = [qa_relevance]
    
    tru_recorder = TruLlama(
        query_engine,
        app_id=app_id,
        feedbacks=feedbacks
    )
    return tru_recorder
```

```
from trulens_eval import Feedback, Tru

tru = Tru()

Tru().reset_database()

tru_recorder = get_prebuilt_trulens_recorder(
    auto_merging_engine,
    app_id ='app_0'
)
```

Output:

TypeError                                 Traceback (most recent call last)
Cell In[34], line 7
      3 tru = Tru()
      5 Tru().reset_database()
----> 7 tru_recorder = get_prebuilt_trulens_recorder(
      8     auto_merging_engine,
      9     app_id ='app_0'
     10 )

Cell In[33], line 10, in get_prebuilt_trulens_recorder(query_engine, app_id)
      6 qa_relevance = Feedback(bedrock.relevance_with_cot_reasons, name="Answer Relevance").on_input_output()
      8 feedbacks = [qa_relevance]
---> 10 tru_recorder = TruLlama(
     11     query_engine,
     12     app_id=app_id,
     13     feedbacks=feedbacks
     14 )
     15 return tru_recorder

[...]

PR #769, opened 2024-01-05
Fix subclass/pydantic error

joshreini1:
Fixes https://github.com/truera/trulens/issues/764

### DIFF
--- a/trulens_eval/trulens_eval/utils/pyschema.py
+++ b/trulens_eval/trulens_eval/utils/pyschema.py
@@ -627,7 +627,7 @@ class WithClassInfo(pydantic.BaseModel):
                 val = obj[k]
 
             if isinstance(typ, type) \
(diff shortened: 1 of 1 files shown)


trulens_eval/trulens_eval/utils/pyschema.py
```diff
@@ -627,7 +627,7 @@ def load(obj, *args, **kwargs):
                 val = obj[k]
 
             if isinstance(typ, type) \
-            and issubclass(typ, WithClassInfo) \
+            and isinstance(typ, WithClassInfo) \
```
piotrm0, line 630: issubclass is needed as `typ` is a type, there is some issue with the check above. Have another PR that helps.

piotrm0:
https://github.com/truera/trulens/pull/770

PR #770, opened 2024-01-05
more optional annotations and some bugfixes

piotrm0:
- Wrapped openai imports in optional contexts as it is not installed with trulens_eval . 
- Wrapped jupyter notebook related imports similarly.
- Bugfix to `issubclass` not a class issue.


trulens_eval/trulens_eval/feedback/provider/__init__.py
```diff
@@ -3,8 +3,11 @@
 from trulens_eval.feedback.provider.hugs import Huggingface
 from trulens_eval.feedback.provider.langchain import Langchain
 from trulens_eval.feedback.provider.litellm import LiteLLM
-from trulens_eval.feedback.provider.openai import AzureOpenAI
-from trulens_eval.feedback.provider.openai import OpenAI
+from trulens_eval.utils.imports import REQUIREMENT_OPENAI, OptionalImports
+
+with OptionalImports(messages=REQUIREMENT_OPENAI):
```
joshreini1, line 8: Why use the optional imports for OpenAI/Azure but not other providers here?

piotrm0: Didn't realize some of the others were also optional, added similar for bedrock and litellm. The rest is not optional.

joshreini1 (approved):

PR #771, opened 2024-01-05
documentation and fix to weakref usage

piotrm0:
Documenting outstanding hacks and other tech debt.

Bugfix for one outstanding issue where paths to components are wrong if an app gets garbage collected.


joshreini1:
Can we call this something more user-friendly than bit-rot?

piotrm0:
> Can we call this something more user-friendly than bit-rot?

Changed to "tech debt" both in filename and text.

joshreini1 (approved):

PR #778, opened 2024-01-08
update langchain_agents notebook

piotrm0:


joshreini1 (approved):
Made some changes, but approved. Please take a second look since I made substantial udpates.

PR #787, opened 2024-01-09
optionals readme

piotrm0:


joshreini1 (changes requested):
- README changes should go in intro.md and gh_top_intro.md otherwise they will be overwritten.
- Should we put OPTIONAL.md directly in docs somewhere?

piotrm0:
> * README changes should go in intro.md and gh_top_intro.md otherwise they will be overwritten.
> * Should we put OPTIONAL.md directly in docs somewhere?

Fixed source of README and don't need doc OPTIONAL for now.

joshreini1 (approved):

PR #790, opened 2024-01-10
Fix missing f-strings

andrewisplinghoff:
I noticed the plain string "App {app_id} was not present in database. Adding it." without "{app_id}" replaced by the actual app id in the logs. This fixes the issue.


joshreini1:
Thanks @andrewisplinghoff  - LGTM

PR #803, opened 2024-01-17
error on deprecated passthrough methods

piotrm0:
Methods of TruChain, TruLlama, others that wrap methods of the enclosed app now throw an error instead of printing the deprecation message. The message in the exception is the same.


joshreini1 (approved):
Can you also note any implications of the sync/async changes in the PR summary?

piotrm0:
> Can you also note any implications of the sync/async changes in the PR summary?

Those are in another PR. I just based this PR off of that one but it hasn't been merged yet so it looks like this PR has more than it does.

PR #823, opened 2024-01-24
Refactor Evaluation Docs

joshreini1:
Target Structure:

### **Feedback Function Overview**
- What is a feedback function?
- Anatomy of a feedback function

### **Running Feedback Functions**
- Run on existing data
    - Run in isolation
    - Run with record
    - Run on dataframe
- Run with your app
  - Feedback modes

### **Feedback Function Selectors**

- Overview
- Finding components to select
- Available shortcuts

### **Feedback Providers**

- LLMProvider
- Provider-specific methods: openai moderation, Huggingface methods


docs/trulens_eval/feedback_functions_existing_data.md
```diff
@@ -0,0 +1,177 @@
...
+    record = chain(""What is langchain?")
+
+tru.run_feedbacks(record, feedbacks=[f_lang_match, f_qa_relevance, f_context_relevance])
+```
+
+### TruVirtual
+
+If your application was run (and logged) outside of TruLens, `TruVirtual` can be used to ingest and evluate the logs.
+
+To incorporate your data into TruLens, you have two options. You can either create a `Record` directly, or you can use the `VirtualRecord` class, which is designed to help you build records for virtual models.
```
yuvneshtruera, line 32: I don't know if this will confuse users, use something like, "designed help you build records without actually running or calling the generative model API"

yuvneshtruera (approved):

PR #832, opened 2024-01-26
factor out common error message from app types

piotrm0:


joshreini1:
Not quite following, can you give an example of the behavior that would produce this error?

piotrm0:
Example: user puts their langchain chain inside truchain and then tries to call a chain method on truchain instead, they will get that error message.

Issue #854, opened 2024-02-04
Custom LangChain Retriever is not logged


q2wxec:
In my Chain,I find that only LangChain Retriever can be logged,Custom Retriever is not logged.

##  LangChain Retriever 
![image](https://github.com/truera/trulens/assets/24437317/7611c1dc-e149-4142-a833-6f0aa1ff229e)
## Custom Retriever
```
class ContextExpandRetriver(BaseRetriever):
    def _get_relevant_documents(
        self,
        query: str,
        *,
        run_manager: CallbackManagerForRetrieverRun,
        **kwargs: Any,
    ) -> List[Document]:
        source_documents = []
        # do something to ContextExpandRetriver
        return source_documents
```
![image](https://github.com/truera/trulens/assets/24437317/fb37d0fc-cef4-4467-8ff6-5b44f751d403)

Just replace retriever in Example blow will reproduce the problem！
https://www.trulens.org/trulens_eval/langchain_quickstart/#load-documents

Because of the bug,the lens App.select_context returns , is not log to the db, and thus can not be dealed by the Feedback
![image](https://github.com/truera/trulens/assets/24437317/5868fd26-c842-4817-a70e-4b202435201c)

joshreini1:
Hey @q2wxec ! Thanks for creating the issue.

I’m not quite following the issue here. Is it that the expected component is not showing in the timeline (not being logged at all), or is it just not being selected by `select_context`

joshreini1:
@q2wxec this may be fixed by #873 - can you give it a try?

q2wxec:
> @q2wxec this may be fixed by #873 - can you install from that branch and give it a try?

ok,i will take a shoot and reply

q2wxec:
> @q2wxec this may be fixed by #873 - can you install from that branch and give it a try?

I think this problem has been solved.However some problem occour while open the dashbord.
## This problem solved by looking into the json
![1707534707713](https://github.com/truera/trulens/assets/24437317/7ca486ec-b756-481d-bbac-001d4484d9a2)
## dashbord problem 
![image](https://github.com/truera/trulens/assets/24437317/5df3f3f3-061b-4d5b-ba42-4c43b68026d6)

joshreini1:
Awesome news. Re: the dashboard error, this is fixed in main (and 0.22.1) but wasn’t in this branch yet. Just pulled it in to the branch

q2wxec:
[...]

PR #873, opened 2024-02-07
Deeper Instrumentation for Hybrid Retrievers + examples

joshreini1:
addresses: https://github.com/truera/trulens/issues/854 +slack requests

Instrumentation updates:

- Add instrumentation for postprocessor and retriever, query_engine components
- Recursively instrument sub-components. This is critical for hybrid/ensemble retrievers.

This results in much better instrumentation (and evals!). Previously, none of the subcomponents of `RetrieverQueryEngine.retrieve` were captured.

Screenshot 2024-02-08 at 7 10 16 PM

Examples:
Add user requested example for evaluating ensemble retrievers (langchain) and hybrid (llama-index)

Cool result:
Screenshot 2024-02-07 at 6 07 02 PM


joshreini1:
Note: this does slow down the wrap application step. Trade-off seems worth it to get better instrumentation.

trulens_eval/trulens_eval/instruments.py
```diff
@@ -743,6 +743,13 @@ def instrument_object(
         # https://github.com/pydantic/pydantic/blob/11079e7e9c458c610860a5776dc398a4764d538d/pydantic/main.py#LL370C13-L370C13
         # .
 
+        # Recursively instrument inner components
```
piotrm0, line 746: This should already be done by what is on line 835. Do you have example of that not recurring in correctly? If so, can you fix it there?

joshreini1: The classes are already included in instrumentation list in tru_llama. What do you mean by "fix it there"?

piotrm0: I mean the subcomponent instrumentation is done around line 835 so I was suggesting that your fix to it not correctly recurring down should be put in the same place.

joshreini1:
User validation that this change solves the issue: https://github.com/truera/trulens/issues/854

trulens_eval/trulens_eval/instruments.py
joshreini1: L835 seems to cover unrecognized container types? I don't think that's what's happening here

joshreini1: Would like to merge this soon, got customer ask for this today.

piotrm0 (approved):

PR #874, opened 2024-02-08
non-threaded pacer

piotrm0:
The endpoint pacing mechanism used threads meaning one was allocated for each endpoint which induced a bunch of overhead. Replaced the threaded mechanism that does not use a separate thread.

Also replaced `run_me` that operated on thunks with `run_in_pace` that operates functions and its args. This means we don't have to wrap everything with lambdas and reduces the number of frames in the call stack.


joshreini1:
If you have any latency benchmarks on this, that’d be pretty useful to tell folks about!

joshreini1 (approved):

PR #935, opened 2024-02-24
Add Vectara Hallucination Detection Model

Josephrp:
### Issue

we're adding the vectara hallucination detection model & demo in context for RAG using huggingface.

### Why 

the more specialized evaluation the better , and this one is great.


joshreini1:
Hey @Josephrp @MN-Noor - looks like there's a bunch of secrets detected in this branch. It might be better to start with a fresh branch to avoid committing all of these secrets to the repo. Also if you haven't already, I'd highly recommend rotating these secrets.

Josephrp:
yes understood with all my appologies :-)

PR #938, opened 2024-02-26
Better selection of main input/output

joshreini1:
This change better extracts the content from main_input and main_output to both display in the TruLens UI and use for helper methods such as `.on_input()` and `.on_output()`. Particularly, this change looks for content following the OpenAI pattern, e.g.

Inputs such as:

```
'{
     "model": "gpt-3.5-turbo",
     "messages": [{"role": "user", "content": "Say this is a test!"}],
     "temperature": 0.7
   }'
```

And outputs such as:

```
{
    "id": "chatcmpl-abc123",
    "object": "chat.completion",
    "created": 1677858242,
    "model": "gpt-3.5-turbo-0613",
    "usage": {
        "prompt_tokens": 13,
        "completion_tokens": 7,
        "total_tokens": 20
    },
    "choices": [
        {
            "message": {
                "role": "assistant",
                "content": "\n\nThis is a test!"
            },
            "logprobs": null,
            "finish_reason": "stop",
            "index": 0
        }
    ]
}
```

Notice in both cases, the text we actually care about is located as the value for the key, `content`. This pattern is widely used and adopted by frameworks such as Pinecone's `Canopy` and middleware such as `LiteLLM`.

See an example of this change below:

**Example 1: Pinecone Canopy**

Before:
This warning in notebook:
```
Unsure what the main input string is for the call to chat with args [[UserMessage(role=<Role.USER: 'user'>, content='How can you get started with Pinecone and TruLens?')]].
Unsure what the main output string is for the call to chat with return type <class 'canopy.models.api_models.ChatResponse'>.
```
And
Screenshot 2024-02-26 at 5 13 55 PM

After:
Screenshot 2024-02-26 at 5 12 05 PM

**Example 2: OpenAI**

Before:
Screenshot 2024-02-27 at 2 47 18 PM

After:

Screenshot 2024-02-27 at 2 49 28 PM

### DIFF
--- a/trulens_eval/trulens_eval/app.py
+++ b/trulens_eval/trulens_eval/app.py
@@ -666,8 +666,24 @@ class App(AppDefinition, WithInstrumentCallbacks, Hashable):
         raise NotImplementedError()
 
     def _extract_content(self, value):
+        """
+        Extracts the 'content' from various data types commonly used by libraries
+        like OpenAI, Canopy, LiteLLM, etc. This method navigates nested data
+        structures (pydantic models, dictionaries, lists) to retrieve the
+        'content' field. If 'content' is not directly available, it attempts to
+        extract from known structures like 'choices' in a ChatResponse. This
+        standardizes extracting relevant text or data from complex API responses
+        or internal data representations.
+        
+        Args:
+            value: The input data to extract content from. Can be a pydantic
+                   model, dictionary, list, or basic data type.
+        
+        Returns:
+            The extracted content, which may be a single value, a list of values,
+            or a nested structure with content extracted from all levels.
+        """
         if isinstance(value, pydantic.BaseModel):
-            # Check for 'content' attribute directly
             content = getattr(value, 'content', None)
             if content is not None:
                 return content
--- a/trulens_eval/trulens_eval/app.py
+++ b/trulens_eval/trulens_eval/app.py
@@ -761,11 +761,11 @@ class App(AppDefinition, WithInstrumentCallbacks, Hashable):
             return str(focus)
 
         logger.warning(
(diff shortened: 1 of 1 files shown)


trulens_eval/trulens_eval/app.py
```diff
@@ -731,7 +749,7 @@ def main_input(
             callable_name(func), all_args
         )
 
-        return None
+        return ""
```
piotrm0, line 752: Perhaps this can fallback to just returning str(all_args) maybe with a prefix "Could not determine main input/output of {str(all_args)}." ?

trulens_eval/trulens_eval/app.py
```diff
@@ -665,6 +665,36 @@ async def main_acall(self, human: str) -> str:
 
         raise NotImplementedError()
 
+    def _extract_content(self, value):
```
piotrm0, line 668: Can you add a note of the libs/apps that produce the types of objects that you are checking for here?

joshreini1: Will add a note. Also updated the PR description with this information.

piotrm0 (approved):

trulens_eval/trulens_eval/app.py
```diff
@@ -685,16 +731,10 @@ def main_input(
...
-                focus = list(focus.model_dump().values())
-                continue
-            
-            if isinstance(focus, Dict):
-                focus = list(focus.values())
-                continue
+            focus = self._extract_content(focus)
 
             if not isinstance(focus, Sequence):
+                print(f"focus {focus} is not a sequence")
```
piotrm0, line 737: Remove debug print.

PR #953, opened 2024-03-04
instrumentation notebook updates and fixes

piotrm0:
- Update symlinks to instrumentation notebook in `notebooks_to_test`.
- [x] Adjusted how methods to instrument are specified so we can check for them on classes as well as instantiations.
- [x] Add method to print out information about what can be instrumented (as opposed to what was instrumented after the fact).
- [x] Replace listings of what can be instrumented in instrumentation notebooks with the above method.
- [x] Include notebook outputs in these notebooks so that an example output is included in the docs since the docs integrate these notebooks.
- [x] Add static tests to make sure the all of the instrumentation code can be loaded without missing names, etc.

----

| <a href="https://ellipsis.dev" target="_blank">Ellipsis</a> | :rocket: This PR description was created by [Ellipsis](https://www.ellipsis.dev) for commit 9a66f14328be675929a7099c515452b8f74d4044.  | 
|--------|--------|

### Summary:
This PR introduces updates and fixes to the instrumentation notebook, including modifications to how methods are specified for instrumentation, addition of a method to print out information about what can be instrumented, replacement of listings in notebooks with this new method, inclusion of notebook outputs in the docs, and addition of static tests for code loading.

**Key points**:
- Updated symlinks to instrumentation notebook in `notebooks_to_test`.
- Modified how methods to instrument are specified, allowing checks on classes as well as instantiations.
- Added a method to print out information about what can be instrumented.
- Replaced listings of what can be instrumented in the notebooks with the new method.
- Included notebook outputs in these notebooks for better integration with the docs.
- Added static tests to ensure all instrumentation code can be loaded without missing names.

----
Generated with :heart: by [ellipsis.dev](https://www.ellipsis.dev)

### DIFF
--- a/docs/trulens_eval/tracking/instrumentation/index.ipynb
+++ b/docs/trulens_eval/tracking/instrumentation/index.ipynb
@@ -122,7 +122,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 2,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -140,7 +140,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 3,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -158,7 +158,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 4,
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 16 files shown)


joshreini1:
It'd be useful to tell the user at the point they set their feedback definition that a specified instrument/method/path does not exist. Is that in scope for this change?

Currently the user does not find out until they run the recorder/app which makes it confusing to debug

piotrm0:
> It'd be useful to tell the user at the point they set their feedback definition that a specified instrument/method/path does not exist. Is that in scope for this change?
> 
> Currently the user does not find out until they run the recorder/app which makes it confusing to debug

We cannot do this when Feedback is defined because app is not know then but we could do some checks after the app is known. This is not part of this PR though.

joshreini1 (approved):

PR #967, opened 2024-03-07
add missing virtual app setup and redirects from old core concept links

joshreini1:


joshreini1:
Note: this change is already deployed to trulens.org

yuvneshtruera (approved):

PR #970, opened 2024-03-08
[DRAFT] lamini provider and endpoint improvements

piotrm0:
- Added `WithOutputType` mixin for `LLMProvider`s to indicate their completion generation can accept `output_type`.
- Added Lamini provider with the above mixin.
- Used output_type in said provider to force completion to generate integer scores.
- [ ] Use output_type to force correct reasons outputs.
- Added `EndpointDelayError` to be raised when an endpoint request produces a result suggesting a need to wait. Catching this in `Endpoint.run_in_pace` to incorporate the suggested delay.
- Updated trubot example to new pinecone and langchain usage.
- Added `pace` argument to `Endpoint` constructor to be able to set the two pacing parameters instead of just `rpm`.
- Added option to trubot to use a specified `LLMProvider` instead of openai.
- Added "with_cot_reasons" to `feedback.rag_triad` to use the COT reasons variant of groundedness.
- [ ] Tests.
- [ ] Example notebooks.


sfc-gh-jreini:
Closing: this PR is 700+ days old and targets a codebase architecture that no longer exists (pre-OTEL, pre-monorepo). Relevant concepts have been captured in newer issues.

PR #1003, opened 2024-03-14
Allow Feedback.run with args even if they had selectors specified.

piotrm0:
- Small adjustments to litellm provider. Updated version lowerbound and removed upper bound as the python 3.8 issue seems to have been addressed on their end.

- Adjusted Feedback.run so one can run it without records/apps and even if the Feedback had selectors provided:
```python
f = Feedback(...).on(...)

f.run(text1="something", text2="something")
```
Previously this would not work if text1 or text2 had selectors set but now one can provide values to .run itself and if all values are provided, no selector will be used and one can run the feedback without the data that would otherwise be needed for the selector (record, app).


joshreini1 (approved):

joshreini1 (changes requested):
This is breaking the release pipeline - why is this needed?

piotrm0:
Release pipeline is broken since this doesn't include the recent fix that was in a different PR.

joshreini1:
Got it - still curious in what cases you would want to provide values to f.run itself?

joshreini1 (approved):

piotrm0:
> Got it - still curious in what cases you would want to provide values to f.run itself?

For simpler testing of feedback functions. I saw people unsure how to run feedback functions after they define them. We explained you can get f.imp but with f.run(...), it may be more direct for them. Also, this allows you to run partially on selected values and partially on other values. Like (f.run(record=record, app=app, text2="something else")) so that the value for one arg (text1) is retrieved from record, but value for a different input (text2) is provided explicitly. May be used for testing.

joshreini1:
Very helpful - thanks!

PR #1038, opened 2024-03-28
Add if_missing.

piotrm0:
- **Feedback.if_missing**: configure feedback functions to do something other than raise an error if a selector is wrong or whatever it wants to select does not exist. You can use this to let feedback ignore records that do not have the data that was called for. Supported are error as before (default), warn (issue warning and mark feedback as skipped), or ignore (just mark as skipped).
- **No -1 results**: if a feedback function implementation fails to find a 0-10 score in some generated text, the result will no longer be -10 (or normalized score -1). Instead the result will remain None and an exception will be raised.

### DIFF
--- a/docs/trulens_eval/api/feedback.md
+++ b/docs/trulens_eval/api/feedback.md
@@ -22,6 +22,8 @@ non-serializable instantiations.
 
 ::: trulens_eval.schema.FeedbackCombinations
 
+::: trulens_eval.schema.FeedbackOnMissingParameters
+
 ::: trulens_eval.schema.FeedbackMode
 
 ::: trulens_eval.schema.FeedbackResult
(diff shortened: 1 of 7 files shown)


joshreini1:
@ebotwick - please give this a try. Thanks!

ebotwick:
I was able to validate that this allows us to evaluate all records regardless of whether certain contexts are missing. 

I did notice however that we assign a value of -1 to a missing feedback, and this is causing some issues with aggregate scores. I.e. some metrics now have an average score of -0.125. Is there a way to put a None value or something here that will not impact the aggregate metrics?

piotrm0:
> I was able to validate that this allows us to evaluate all records regardless of whether certain contexts are missing.
> 
> I did notice however that we assign a value of -1 to a missing feedback, and this is causing some issues with aggregate scores. I.e. some metrics now have an average score of -0.125. Is there a way to put a None value or something here that will not impact the aggregate metrics?

I could not find how -1 could be the result of missing feedback but I do know that -1 is what gets assigned as a result of a feedback that produces a piece of text from which a 0-10 score cannot be extracted. Please double check where the -1 are coming from.

joshreini1 (approved):
-1 scores not coming from this PR/separate issue. This change looks good!

PR #1049, opened 2024-04-03
Fix import and favicon

arn-tru:
- Added `psutil` to imports (As it was giving "module missing" error on a fresh environment)
- Added favicon to the streamlit app
- Refactored `add_logo_and_style_overrides()` function to handle page config in a common function


joshreini1 (approved):
Awesome stuff :) Thanks @arn-tru !

PR #1051, opened 2024-04-03
check if a selector names an empty property

joshreini1:
- Check also if a selector names an empty property


piotrm0:
I'm unsure whether this is a good idea. Empty strings are indicative of the same problem as missing fields in general. 

This kind of logic perhaps can be closer to the feedback implementations. I.e. check for special values at the feedback if applicable to that feedback function.

PR #1074, opened 2024-04-16
feature(Improved Trace Display): [MLNN-1342] Improved Trace Display

walnutdust:
- **Improved Trace Display**: Consolidated metadata, app/record JSONs, and timeline in a tree-like display for ease of use and navigation. See the spans in a trace, as well as the call stack, inputs, and outputs of the different spans. 

https://github.com/truera/trulens/assets/36098015/a14b7626-460d-488c-92a4-70f867915d8a

- **Consolidated feedback function display for trace**: Use pills for feedback function score display, with a helpful icon determining if it passed/failed the user set thresholds.

- Deprecated the old trace timeline view. 
- Deprecated the old trace component/subcomponent view for the python object of a stack call.


piotrm0:
> * Deprecated the old trace timeline view.

There doesn't seem to be a replacement for this included. Is this intentional?

walnutdust:
@piotrm0 yes, they were omitted from the designs after looking into tools like LangSmith, BrainTrust, OpenEvals, LangFuse, Arize, and others and noting that they don't have them either. That said, happy to have a proper discussion regarding the use cases of this functionality to decide before moving forward. Will convert this to a draft PR.

piotrm0 (approved):

walnutdust:
@piotrm0 - updated the trace display to retain the timeline view as an alternative to the tree view, please take a look and let me know what you think

PR #1090, opened 2024-04-19
split schema.py

piotrm0:
- Split `schema.py` into `schema/{__init__, base, app, record, feedback, types}.py`.
- In process of the above, edited a lot of files to follow https://google.github.io/styleguide/pyguide.html#22-imports in order to avoid circular import issues.
- This will make room for new Span classes.


walnutdust (approved):
Looks good

PR #1103, opened 2024-04-26
span types -> feature/traces branch

piotrm0:

- Spans categorize record calls across a variety of app and component types.


walnutdust (approved):
LGTM!

What's the recommended approach for serializing `Span`s to JSON? Streamlit uses it when going from python -> react boundary, and I'm currently working around it by using `jsonify` from `trulens_eval.utils.json`, but that messes slightly with `context`, so I'm getting something like:

```
{
    "name": "retrieve_chunks",
    "kind": "INTERNAL",
    "status": "UNSET",
    "status_description": null,
    "start_timestamp": 1714162347092273000,
    "end_timestamp": null,
    "context": [
        2.9879168338455205e+38,
        10115842930225359000,
        false,
        0,
        {
            "__tru_non_serialized_object": {
                "cls": {
                    "name": "TraceState",
                    "module": {
                        "package_name": "opentelemetry.trace",
                        "module_name": "opentelemetry.trace.span"
                    },
                    "bases": null
                },
                "id": 5458539152,
                "init_bindings": null
            }
        },
        true
    ],
    "events": [],
    "links": {},
    "attributes": {
        "trulens_eval@span_type": "SpanRetriever",
        "trulens_eval@input_text": "hello",
        "trulens_eval@retrieved_contexts": [
            "Relevant chunk: HELLO",
            "Relevant chunk: olleh",
            "Relevant chunk: I allocated 56 bytes to pretend I'm doing something."
        ]
    },
    "attributes_metadata": {}
}
```

and the `context` field is slightly messed up

piotrm0:
> What's the recommended approach for serializing `Span`s to JSON?

They are pydantic models so you can use span.model_dump() or model_dump_json() . You can then reload using span.model_validate(_json) . But jsonify is ok as the same info that is missing in model_dump is missing from jsonify.

piotrm0:
Ohh, I didn't realize the large int got dumped as a float. Might have to compensate for that if json doesn't like 128 bit ints.

walnutdust:
will give that a shot, thanks!

PR #1111, opened 2024-04-30
feature/traces: span contents and examples

piotrm0:
Added more span type field definitions and examples of filling them in based on new dummy additions to the custom app example. @walnutdust : The UI is showing all of the spans as other which might happen if the conversion to spans is done in a different process than where the app is ran. See span example notebook to see their actual types. 

Added ability to include `self`, the instances being instrumented, as argument to the decorated span-filling methods.


trulens_eval/trulens_eval/trace/span.py
```diff
@@ -282,27 +294,96 @@ class SpanRetriever(SpanTyped):
...
+    model_name = Span.attribute_property("model_name", str)
+    """The model name of the reranker."""
+
+    top_n = Span.attribute_property("top_n", int)
+    """The number of contexts to rerank."""
+
+    input_context_texts = Span.attribute_property("input_context_texts", List[str])
+    """The contexts being reranked."""
+
+    input_context_scores = Span.attribute_property("input_score_scores", Optional[List[float]])
```
walnutdust, line 321: should the attribute property be `input_context_scores`?

walnutdust (approved):
LGTM 👍

Issue #1101, opened 2024-04-25
[BUG] TruChain - no values returned for Groundedness feedback function for Bedrock


mahesh-hv-appmod:
**Bug Description**
TruEval returns an empty array for Groundedness values

**To Reproduce**
from langchain.chains import LLMChain
from langchain.schema import StrOutputParser
from langchain_core.runnables import RunnablePassthrough

retriever = vectorstore.as_retriever()

retrieval_chain = (
    {"context": retriever, "query": RunnablePassthrough()}
    | prompt_template
    | llm
    | StrOutputParser()
)

retrieval_chain.invoke(question)

from trulens_eval.app import App
from trulens_eval.feedback import Groundedness
from trulens_eval import TruChain, Feedback, Tru, Select, Bedrock

tru = Tru()
tru.reset_database()

context = App.select_context(retrieval_chain)

# Initialize model for feedback function collection class:
bedrock = Bedrock(model_id = "amazon.titan-tg1-large", region_name="us-east-1")

# Define a groundedness feedback function
grounded = Groundedness(groundedness_provider=bedrock)

f_groundedness = (
    Feedback(grounded.groundedness_measure_with_cot_reasons, name="Groundedness")
    .on(context.collect()) # collect context chunks into a list
    .on_output()
    .aggregate(grounded.grounded_statements_aggregator)
)
    
f_relevance = Feedback(bedrock.relevance, name="Relevance").on_input_output()

tru_recorder = TruChain(retrieval_chain,
    app_id='Base LLM',
    selectors_check_warning=True,                   
    feedbacks=[f_groundedness, f_relevance])

questions = [
    "What is Acme Co.?", 
    "Who are the competetiors?",
    "What’s the financial profile?"
  ]

with tru_recorder as recording:
    for question in questions:
        llm_response = retrieval_chain.invoke(question)

**Expected behavior**
expect to see some values for the Groundedness metric. Relevance metrics are being returned

**Environment:**
Python 3.10.13
langchain: 0.1.16
trulens_eval: 0.27.0

**Additional context**

mahesh-hv-appmod:
small edit

the calls to the groundedness functions pass in empty arrays '[]'. see below:

![groundedness](https://github.com/truera/trulens/assets/77741042/b7d328b2-22cc-4d4e-ac9f-d5f81c89867f)

vasokb:
Hi! I have exactly the same issue, both with context relevance as well as groundedness. I am using hugging face as the provider.
Any update on that?

[...]

PR #1112, opened 2024-04-30
update groundedness prompt

bpmcgough:
TL;DR I was having issues with the groundedness feedback function when passing an object in as the source, and I determined the issue was that the instructions specify to return the "exact sentence" in the source where the data comes from. If the source is an object, there may be no sentence at all, so this was returning nothing. 

After making this change, my groundedness function works as expected, as shown in the video below.

Video walkthrough
https://www.loom.com/share/afdb7a3ac28e4bed9f31458804c357e1?sid=8a310c28-361d-4df4-8fd3-920c87236992

Could be related to https://github.com/truera/trulens/issues/1101


joshreini1:
LGTM - will run some testing later this week to confirm. Thanks for the change and explanation!

joshreini1 (approved):

PR #1116, opened 2024-04-30
Remove references to running moderation endpoint on AzureOpenAI

joshreini1:
- Moderation endpoint is not available on Azure. Instead, the moderation functions are run as content filters with every api call. Make this clear in docs.

### DIFF
--- a/docs/trulens_eval/api/provider/openai/azureopenai.md
+++ b/docs/trulens_eval/api/provider/openai/azureopenai.md
@@ -5,9 +5,6 @@ Below is how you can instantiate _AzureOpenAI_ as a provider.
 All feedback functions listed in the base [LLMProvider
 class][trulens_eval.feedback.provider.base.LLMProvider] can be run with the _AzureOpenAI_ Provider.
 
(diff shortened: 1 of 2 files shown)


docs/trulens_eval/api/provider/openai/azureopenai.md
```diff
@@ -1,10 +1,12 @@
 # AzureOpenAI Provider
 
-Below is how you can instantiate AzureOpenAI as a provider.
+Below is how you can instantiate _AzureOpenAI_ as a provider.
```
piotrm0, line 3: is AzureOpenAI a proper name or just reference to a class?

joshreini1: Proper name would be _Azure OpenAI_ (with space). Perhaps here I'll add the space to reference the proper name rather then the class?

piotrm0: Yes, and if it is a class, you can use the reference [AzureOpenAi][...full path to class] instead.

piotrm0 (approved):

PR #1119, opened 2024-05-01
[trurails] update __getattr__ in select_context

joshreini1:
Fixes issue in slack: https://aiqualityforum.slack.com/archives/C03S0PDJ57G/p1714405839410009


piotrm0:
I think there might be an underlying issue that TruRails had __name__ asked for. We shouldn't be doing that so need to find the actual bug.

piotrm0 (approved):

piotrm0:
Ok to merge. Original source of bug was fixed but this PR can help in the future.

PR #1127, opened 2024-05-06
Remove placeholder feedback for asynchronous responses

arn-tru:
- Removed placeholder feedback that was stored when asynchronous process starts. (This stops multiple feedbacks from being recorded.)

### DIFF
--- a/trulens_eval/trulens_eval/instruments.py
+++ b/trulens_eval/trulens_eval/instruments.py
@@ -578,27 +578,11 @@ class Instrument(object):
                 return records
 
             if isinstance(rets, Awaitable):
-                # If method produced an awaitable, add a placeholder record
-                # stating that results are not ready and return an awaitable
-                # that will capture the results when they are ready.
+                # If method produced an awaitable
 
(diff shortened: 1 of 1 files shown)


piotrm0:
The issue being solved by this might have been solved by https://github.com/truera/trulens/pull/1110 .

piotrm0:
> @piotrm0 why were we writing a placeholder before? Is there some reason we're overlooking why this is necessary?

This message was added to let a user know that if they used some (langchain/other) method that was asynchronous and they didn't await it, their feedbacks might not run properly.

joshreini1 (approved):

piotrm0:
> > The issue being solved by this might have been solved by #1110 . @arn-tru , can you check whether the bugs you are solving with this PR are still bugs with 1110? If they are solved by 1110, we could preserve the awaitable placeholder.
> 
> The issue still persists on #1110

Was the issue related to `__name__` attribute? If so, can you check again? (I added a fix to this to the mentioned PR).

PR #1156, opened 2024-05-24
❄️ Snowflake Arctic quickstart

joshreini1:
- Arctic Quickstart


piotrm0:
Is access of these models through SentenceTransformers and LiteLLM the typical use case?

sfc-gh-jreini:
W

> Is access of these models through SentenceTransformers and LiteLLM the typical use case?

Typical for now but we can wait on this and sub with Cortex provider when it's ready

sfc-gh-jreini:
Let's stick with replicate arctic for now. Can always add cortex arctic in the future

sfc-gh-jreini (approved):

PR #1168, opened 2024-05-31
[DRAFT] otel-like recording

piotrm0:
This work has moved to another PR: https://github.com/truera/trulens/pull/1346 . Keeping this around until the new PR is merged.

- Changed implementation of recording to resemble that of OpenTelemetry. This is to make further work on otel compatibility easier. Wrapped functions now produce spans but do not serialize anything. Serialization is done only after a recording context is closed.
- Reconstructs `Record`s from spans collected in the process of executing a recording context. These records are available in the recording context as before and sent to the same places as before (i.e. to database, for feedback evaluation).
- EXPERIMENTAL: Added `Tru` class argument `otel_exporter` to specify where to export spans. If specified, the spans that make up a record will be exported there whenever a record is created (upon recording context closure).
- Removed `ComponentView` and related code as that is not being used and is being replaced by spans in the feature/traces work.
- Factored out function wrapping that was done in `trulens_eval.instruments` (for tracing) and in `trulens_eval.feedback.provider.endpoint.base` (for cost tracking) into a common solution found in a new utility module `trulens_eval.utils.wrap`.
- Adjusted `EndpointCallback` used for cost tracking to extent the generalized callable wrapping callbacks found in the new `wrap` module.
- Updated existing endpoints to use the new cost tracking methodology.


sfc-gh-jreini:
Closing: superseded by current OTEL-based architecture. 500+ days stale.

PR #1186, opened 2024-06-06
Allow transforms to produce main input, output

sfc-gh-chu:
- **Heading**: delete this list if this PR does not introduce any changes that need announcing.
- Optional args_transform and ret_transform arguments to `Instrument.tracked_method_wrapper` that will be called on arguments and return values to determine main input string and output string respectively

- There should be something here at least.
- No decided interface for this method yet. Can be a decorator argument

### DIFF
--- a/trulens_eval/trulens_eval/app.py
+++ b/trulens_eval/trulens_eval/app.py
@@ -1104,8 +1104,19 @@ class App(mod_app_schema.AppDefinition, mod_instruments.WithInstrumentCallbacks,
 
             assert len(calls) > 0, "No information recorded in call."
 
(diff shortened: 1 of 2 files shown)


sfc-gh-jreini:
Please add a usage example for this

sfc-gh-jreini:
> Please add a usage example for this

or rather, is this change a user-facing change? how can folks take advantage of this

sfc-gh-chu:
@joshreini1 haven't had a chance to update here. This is an internal arg change for now. Will need to discuss with @piotrm0 how this looks in practice.

PR #1200, opened 2024-06-12
SkipEval exception

sfc-gh-pmardziel:
- **SkipEval**: Added exception to be raised by a feedback function to indicate that it should be skipped and not aggregated. This is for use cases were the feedback function is evaluated multiple times on different data and the determination of whether an evaluation on one piece of data should be skipped is done in the feedback implementation itself. Note the related existing feature `if_exists` to skip the entire feedback function for a whole record.


trulens_eval/trulens_eval/feedback/feedback.py
```diff
@@ -860,13 +889,24 @@ def run(
...
                     result_and_meta, part_cost = mod_base_endpoint.Endpoint.track_all_costs_tally(
                         self.imp, **ins
                     )
 
                     cost += part_cost
+
+                except SkipEval as e:
+                    e.feedback = self
+                    e.ins = ins
+                    skipped_exceptions.append(e)
```
sfc-gh-chu, line 906: any reason we track every skipped exception instance w/ metadata as opposed to (for example) a count?

trulens_eval/trulens_eval/feedback/feedback.py
```diff
@@ -917,6 +957,15 @@ def run(
                     UserWarning,
                     stacklevel=1
                 )
+                # Log any SkipCombination exceptions in case that was the reason
+                # why no results were collected.
+                if len(skipped_exceptions) > 0:
+                    warnings.warn(
+                        "One or more evaluations were skipped because they raised SkipEval (see earlier warnings).",
```
sfc-gh-chu, line 964: can we report the % or count of SkipEvals raised as well?

sfc-gh-chu (approved):
lgtm w/ some small feedback

trulens_eval/trulens_eval/feedback/feedback.py
sfc-gh-pmardziel: felt simpler to write and more future proof if we want to do more with those exceptions

PR #1223, opened 2024-06-18
Enable regex to extract floats in score generation

sfc-gh-dhuang:
- **Heading**: 
During benchmarking of various feedback providers, I notice there are models (i.e. a finetuned `mixtral-8x7b` that tend to give `10.0` instead of `10` in their feedback scoring before normalization. In the current implementation, `PATTERN_INTEGER` will extract `0` and `10` from `10.0` and eventually pick the lesser value. 

Failing example when testing groundedness feedback functions - where the score accompanying COT was interpreted as `0`, instead of the expected `10`:
```
0.0,
 {'reasons': 'STATEMENT 0:\nCriteria: I am a man and I love fish,\nSupporting Evidence: The source states "All men love fish", and the statement contains "I am a man and I love fish". The source contains the information that a man loves fish, and the statement contains the information that the speaker is a man and loves fish.\nScore: 10.0\n'}
```

I'm switching to `PATTERN_NUMBER` to unblock for now.

- This is only a stopgap solution and I might be missing some contexts here as in why `PATTERN_INTEGER` was used over `PATTERN_NUMBER` in the previous PR. cc @sfc-gh-pmardziel to add more background if I'm missing sth obvious. 

 I do believe we should move toward  structured and systematic feedback score generation mechanisms with some self-refining prompt iterations (i.e. via DSPy) ASAP for more robust score generation, ideally before integrating w/ the monitoring stack, even at the cost of slightly higher token usage/cost/latency (which can also be alleviated via better prompts and instruction tuning).


sfc-gh-jreini (approved):

trulens_eval/trulens_eval/utils/generated.py
```diff
@@ -65,14 +65,16 @@ def re_0_10_rating(s: str) -> int:
...
     if not matches:
-        raise ParseError("int or float number", s, pattern=PATTERN_INTEGER)
+        raise ParseError("int or float number", s, pattern=PATTERN_NUMBER)
 
     vals = set()
     for match in matches:
         try:
-            vals.add(validate_rating(int(match)))
+            vals.add(
+                validate_rating(int(float(match)))
```
sfc-gh-dkurokawa, line 76: kinda late to the game, but:
1. Why are we constrained to ints?
2. If we are constrained to ints, shouldn't we round it instead of flooring it?
3. The doc on L54 says "If the string does not match an integer ... raises an error ...", but this won't do that.

sfc-gh-dhuang: 1. No particular reason beyond easier intrepretability for the end users AFAIK.
2. Make sense - I can make a change to this
3. Same as 1. - I do recall we went back and forth a bit on this and the doc just became outdated b/c of my change. Will update

sfc-gh-dhuang: https://github.com/truera/trulens/pull/1244 fix PR

PR #1224, opened 2024-06-18
TruLens Streamlit components

sfc-gh-jreini:
- Add TruLens components that can be added to streamlit dashboards

https://github.com/truera/trulens/assets/171636836/3a5a0b53-27ed-442d-9139-6f118d332400


sfc-gh-dhuang:
FYI you can open the PR as a Draft or still hit `Convert to draft` near the reviewer section even after PR is opened :)

trulens_eval/trulens_eval/streamlit.py
```diff
@@ -0,0 +1,274 @@
...
+tru = Tru()
+
+
+class FeedbackDisplay(BaseModel):
+    score: float = 0
+    calls: List[FeedbackCall]
+    icon: str
+
+
+def trulens_leaderboard(app_ids: List = None):
```
sfc-gh-dkurokawa, line 35: `List[str]`?

trulens_eval/trulens_eval/streamlit.py
```diff
@@ -0,0 +1,274 @@
...
+from trulens_eval.schema.feedback import FeedbackCall
+from trulens_eval.schema.feedback import FeedbackDefinition
+from trulens_eval.schema.record import Record
+from trulens_eval.utils import display
+from trulens_eval.utils.python import Future
+from trulens_eval.ux import styles
+from trulens_eval.ux.components import draw_metadata
+from trulens_eval.ux.styles import CATEGORY
+
+tru = Tru()
```
sfc-gh-dkurokawa, line 26: I don't think you need this given L51?

trulens_eval/trulens_eval/streamlit.py
```diff
@@ -0,0 +1,274 @@
...
+
+        ```python
+        from trulens_eval import streamlit as trulens_st
+
+        trulens_st.trulens_leaderboard()
+        ```
+    """
+    tru = Tru()
+
+    lms = tru.db
```
sfc-gh-dkurokawa, line 53: for my edification, what's `lms` stand for?

trulens_eval/trulens_eval/streamlit.py
```diff
@@ -0,0 +1,274 @@
...
+            row.feedback_json["implementation"]["name"]
+        ): row.feedback_json.get("higher_is_better", True)
+        for _, row in feedback_defs.iterrows()
+    }
+
+    if df.empty:
+        st.write("No records yet...")
+        return
+
+    df = df.sort_values(by="app_id")
```
sfc-gh-dkurokawa, line 68: probably should just do `df.sort_values(by="app_id", inplace=True)` in case the `df` is big and you don't want to duplicate for no reason.

trulens_eval/trulens_eval/streamlit.py
```diff
@@ -0,0 +1,274 @@
...
+
+    if df.empty:
+        st.write("No records yet...")
+
+    apps = list(df.app_id.unique()) if app_ids is None else app_ids
```
sfc-gh-dkurokawa, line 73: probably should just replace this with:
```
if app_ids is None:
    app_ids = list(df.app_id.unique())
```
and then use `app_ids` instead of `apps` since you are dealing with app ids and not apps. It won't change the variable for the caller of this function don't worry.

trulens_eval/trulens_eval/streamlit.py
```diff
@@ -0,0 +1,274 @@
...
+
+            st.write(
+                styles.stmetricdelta_hidearrow,
+                unsafe_allow_html=True,
+            )
+
+            higher_is_better = feedback_directions.get(col_name, True)
+
+            if "distance" in col_name:
+                feedback_cols[i].metric(
```
sfc-gh-dkurokawa, line 129: why don't you care about the color and stuff if it's distances?

trulens_eval/trulens_eval/streamlit.py
```diff
@@ -0,0 +1,274 @@
...
+        record, response = recording.get()
+
+        trulens_st.trulens_leaderboard()
+        ```
+    """
+    feedback_cols = []
+    feedbacks = {}
+    icons = []
+    for feedback, feedback_result in record.wait_for_feedback_results().items():
+        print(feedback.name, feedback_result.result)
```
sfc-gh-dkurokawa, line 176: why are we printing this out directly?

trulens_eval/trulens_eval/streamlit.py
```diff
@@ -0,0 +1,274 @@
...
+    selected_feedback = pills(
+        "Feedback functions",
+        feedback_cols,
+        index=None,
+        format_func=lambda fcol: f"{fcol} {feedbacks[fcol].score:.4f}",
+        label_visibility=
+        "collapsed",  # Hiding because we can't format the label here.
+        icons=icons,
+        key=
+        f"{call_data['feedback_name']}_{len(feedbacks)}"  # Important! Otherwise streamlit sometimes lazily skips update even with st.exprimental_fragment
```
sfc-gh-dkurokawa, line 200: nit: s/exprimental/experimental/

trulens_eval/trulens_eval/utils/display.py
```diff
@@ -0,0 +1,20 @@
+import pandas as pd
+
+
+def get_feedback_result(tru_record, feedback_name):
```
sfc-gh-dkurokawa, line 4: put the types in. Also you should probably put the return types in in a lot of the functions above.

sfc-gh-dkurokawa (commented):
One overall thing that confuses me is that is there a reason we're branding this stuff as `trulens_st`? Like is it important that it's `streamlit`? Why not like `ui` or something that way if we change it later it's not a misnomer or something.

trulens_eval/trulens_eval/streamlit.py
sfc-gh-jreini: Distances are not bounded 0-1 and same thresholding doesn't apply. Also copied from `Leaderboard.py`

trulens_eval/trulens_eval/streamlit.py
sfc-gh-jreini: not sure really. this is mostly copied code from `leaderboard.py`

sfc-gh-jreini:
> One overall thing that confuses me is that is there a reason we're branding this stuff as `trulens_st`? Like is it important that it's `streamlit`? Why not like `ui` or something that way if we change it later it's not a misnomer or something.

Yeah, these are streamlit components targeted at addressing people already building with Streamlit. This is useful because:

- Streamlit is oriented as a frontend for GenAI
- We already use streamlit as the TruLens frontend, so this is relatively easy to do. However the current TruLens front end is not malleable, and users often rebuild portions of it into their own streamlit dashboards.
- This makes it easier to integrate into the large set of streamlit apps used by Snowflake users

trulens_eval/trulens_eval/streamlit.py
```diff
@@ -65,21 +65,22 @@ def trulens_leaderboard(app_ids: List = None):
         st.write("No records yet...")
         return
 
-    df = df.sort_values(by="app_id")
+    df = df.sort_values(by="app_id", inplace=True)
```
sfc-gh-dkurokawa, line 68: nit: you don't need to say `df = ` since it changes it in place.

trulens_eval/examples/expositional/end2end_apps/streamlit_app/app.py
```diff
@@ -43,4 +44,7 @@ def generate_response(input_text):
 
 if submitted:
     trulens_st.trulens_feedback(record=record)
-    trulens_st.trulens_trace(record=record)
+    app = tru.get_app(app_id=record.app_id)
+    records, feedback = tru.get_records_and_feedback()
+    record_json = json.loads(records.loc[records['record_id'] == record.record_id]['record_json'].values[0])
+    trulens_st.trulens_trace(record_json=record_json, app=app)
```
sfc-gh-dkurokawa, line 50: is there a reason you changed to this call pattern over the simpler `trulens_st.trulens_trace(record=record)` you had before?

trulens_eval/trulens_eval/streamlit.py
```diff
@@ -0,0 +1,255 @@
...
+                    delta_color=(
+                        "normal" if cat.compare(
+                            mean, CATEGORY.PASS[cat.direction].threshold
+                        ) else "inverse"
+                    ),
+                )
+
+        st.markdown("""---""")
+
+@st.experimental_fragment(run_every=2)
```
sfc-gh-gtokernliang, line 150: @sfc-gh-jreini `experimental_fragment` only came out in v1.33.0, while it looks like we have `streamlit          >= 1.32.2` - we might want to bump the version

trulens_eval/examples/expositional/end2end_apps/streamlit_app/app.py
sfc-gh-jreini: Yes - before I was calling Tru internally which was not ideal since the user may have different Tru specification. Better if they get the data themselves from their Tru connection.

trulens_eval/trulens_eval/streamlit.py
```diff
@@ -0,0 +1,255 @@
...
+
+    !!! example
+
+        ```python
+        from trulens_eval import streamlit as trulens_st
+
+        trulens_st.trulens_leaderboard()
+        ```
+    """
+    tru = Tru()
```
sfc-gh-gtokernliang, line 51: nice - this is one scenario where the singleton nature of it is helpful

trulens_eval/trulens_eval/streamlit.py
```diff
@@ -0,0 +1,255 @@
...
+
+    lms = tru.db
+    df, feedback_col_names = lms.get_records_and_feedback([])
```
sfc-gh-gtokernliang, line 54: should we not pass in `app_ids` here?

trulens_eval/trulens_eval/streamlit.py
```diff
@@ -0,0 +1,255 @@
...
+        str: The icon for the feedback
+    """
+    cat = CATEGORY.of_score(
+        result or 0,
+        higher_is_better=fdef.higher_is_better
+        if fdef.higher_is_better is not None else True
+    )
+    return cat.icon
+
+def trulens_trace(record_json: dict, app: AppDefinition):
```
sfc-gh-gtokernliang, line 233: should this take a Record and transform it here?

trulens_eval/trulens_eval/streamlit.py
```diff
@@ -0,0 +1,255 @@
...
+
+    Returns:
+        str: The icon for the feedback
+    """
+    cat = CATEGORY.of_score(
+        result or 0,
+        higher_is_better=fdef.higher_is_better
+        if fdef.higher_is_better is not None else True
+    )
+    return cat.icon
```
sfc-gh-gtokernliang, line 231: nit: what will be our criteria for items that are here versus in `utils`?

sfc-gh-gtokernliang: also non-blocker: we might be able to reuse this code for the normal record table components

sfc-gh-gtokernliang (commented):
we might want to come up with some internal rules of thumb with regards to UI, for "thick" components that handle connection to the database + logic, versus "thin" components that handle just the display. In addition - for the thin components, what is the expected input for a certain class of inputs (e.g. to pass in a record, should users pass in a `Record`, or a `record_json`, etc)

trulens_eval/examples/expositional/end2end_apps/streamlit_app/app.py
sfc-gh-dkurokawa: In that case, shouldn't the `trulens_leaderboard` also do that?

Also it may be easier for the user to just take in the `tru` and the `record` since what's going on in L49 is a bit complex for the user I'd say.

trulens_eval/trulens_eval/streamlit.py
sfc-gh-jreini: thanks for catching

trulens_eval/trulens_eval/streamlit.py
sfc-gh-jreini: changed this back to just using `Tru` singleton. Think this is actually okay because of singleton nature

trulens_eval/trulens_eval/streamlit.py
sfc-gh-jreini: Moved `get_icon` to `utils/display.py`

sfc-gh-jreini:
@sfc-gh-dkurokawa seems like your comments got squashed, but I think they should be addressed. I moved trulens_trace back to just calling tru - this should actually be okay because its a singleton.

sfc-gh-dkurokawa:
> @sfc-gh-dkurokawa seems like your comments got squashed, but I think they should be addressed. I moved trulens_trace back to just calling tru - this should actually be okay because its a singleton.

AFAICT, it's not technically a singleton, like if you give it a name you can have more then one. That is you can have:
```
tru_app1 = Tru(name="app1")
tru_app2 = Tru(name="app2)
```
as per the code it seems (I haven't tested it though).

sfc-gh-jreini:
> > @sfc-gh-dkurokawa seems like your comments got squashed, but I think they should be addressed. I moved trulens_trace back to just calling tru - this should actually be okay because its a singleton.
> 
> AFAICT, it's not technically a singleton, like if you give it a name you can have more then one. That is you can have:
> 
> ```
> tru_app1 = Tru(name="app1")
> tru_app2 = Tru(name="app2)
> ```
> 
> as per the code it seems (I haven't tested it though).

Ran the following test that seems to indicate it is a singleton/this would be okay. Let me know if you disagree

Set up two instances with different parameters and check fi they're equal:

```python
from trulens_eval import Tru

tru_app1 = Tru(database_url='sqlite:///memory')
tru_app2 = Tru()
print(tru_app1 is tru_app2)
```

```
True
```

Manually check arguments from app 1.

```python
args_tru_app1 = tru_app1.__dict__
print("Arguments used to initialize tru_app1:", args_tru_app1)
```

```
Arguments used to initialize tru_app1: {'db': SQLAlchemyDB(redact_keys=False, table_prefix='trulens_', engine_params={'url': 'sqlite:///memory',
```

Manually check arguments from app 2.

```python
args_tru_app2 = tru_app2.__dict__
print("Arguments used to initialize tru_app2:", args_tru_app2)
```

```
Arguments used to initialize tru_app2: {'db': SQLAlchemyDB(redact_keys=False, table_prefix='trulens_', engine_params={'url': 'sqlite:///memory', ...
```

Arguments from app 2 are actually from app 1

trulens_eval/trulens_eval/streamlit.py
```diff
@@ -0,0 +1,242 @@
...
+    """
+
+    tru = Tru()
+
+    app = tru.get_app(app_id=record.app_id)
+    records, feedback = tru.get_records_and_feedback()
+    record_json = json.loads(
+        records.loc[records['record_id'] == record.record_id]
+        ['record_json'].values[0]
+    )
```
sfc-gh-gtokernliang, line 241: @sfc-gh-jreini curious - could this not just be `json_str_of_obj(record)`? Looking at https://github.com/truera/trulens/blob/b95d5ed954a5e4566fa505fa883eb23d94e97a7a/trulens_eval/trulens_eval/database/orm.py#L242

sfc-gh-jreini: Thanks - this works. Way cleaner :)

trulens_eval/trulens_eval/streamlit.py
```diff
@@ -0,0 +1,237 @@
...
+        ```python
+        from trulens_eval import streamlit as trulens_st
+
+        trulens_st.trulens_leaderboard()
+        ```
+    """
+    tru = Tru()
+
+    lms = tru.db
+    df, feedback_col_names = lms.get_records_and_feedback(app_ids=[app_ids])
```
sfc-gh-gtokernliang, line 54: this looks like we're passing in a list of list of strings, is that accurate?

sfc-gh-gtokernliang (approved):
LGTM

trulens_eval/trulens_eval/streamlit.py
sfc-gh-jreini: fixed -thanks

PR #1226, opened 2024-06-19
Demo Context Relevance Guardrailing using small models

sfc-gh-chu:
- Filters retrieved contexts using a small local model finetuned to score context relevance before generating a LLM response.
 - Adds a streamlit UI toggle to enable guardrailing

### DIFF
--- a/trulens_eval/examples/snowflake_demo/llm.py
+++ b/trulens_eval/examples/snowflake_demo/llm.py
@@ -10,6 +10,7 @@ import streamlit as st
 from streamlit.delta_generator import DeltaGenerator
 
 # replicate key for running model
+from feedback import small_local_model_provider
 from trulens_eval.tru_custom_app import instrument
 
 FRIENDLY_MAPPING = {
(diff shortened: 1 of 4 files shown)


trulens_eval/examples/snowflake_demo/llm.py
```diff
@@ -155,13 +156,24 @@ def generate_response(
                 final_response += chunk
 
         return _reencode_outputs(final_response)
+    
+    @instrument
+    def filter_contexts(self, query: str, contexts: list[str], threshold: float = .5):
```
sfc-gh-jreini, line 161: Can we use the native TruLens filter here?

sfc-gh-jreini:
In `retrieve.py` there is existing filtering (set to 0 threshold). Would be good to remove it there for simplicity

trulens_eval/examples/snowflake_demo/llm.py
sfc-gh-chu: Saw your other PR. Let's use the built in guardrailing now that it's checked in. Will close this one

PR #1230, opened 2024-06-20
adjust optional imports assert_installed and optional imports docs.

sfc-gh-pmardziel:
- Simplified `assert_installed` usage and moved the optional imports doc into alongside the contributing section.


trulens_eval/trulens_eval/Example_TruBot.py
```diff
@@ -16,16 +16,14 @@
 from trulens_eval.utils.imports import OptionalImports
 from trulens_eval.utils.imports import REQUIREMENT_PINECONE
 
-with OptionalImports(messages=REQUIREMENT_PINECONE):
+with OptionalImports(messages=REQUIREMENT_PINECONE) as opt:
```
sfc-gh-jreini, line 19: what's up with aliasing as opt? why do this?

trulens_eval/trulens_eval/appui.py
```diff
@@ -11,19 +11,22 @@
...
+    import ipywidgets
     from ipywidgets import widgets
+
+with opt:
     import traitlets
     from traitlets import HasTraits
     from traitlets import Unicode
 
-OptionalImports(messages=REQUIREMENT_NOTEBOOK).assert_installed(ipywidgets)
+opt.assert_installed(ipywidgets).assert_installed(traitlets)
```
sfc-gh-jreini, line 29: not following what this change is here, can you explain? why assert install traitlets also?

sfc-gh-jreini (approved):

trulens_eval/trulens_eval/appui.py
sfc-gh-pmardziel: Just in case it is not installed. It should get installed with ipywidgets but in case someone's installation is broken, this might produce a useful message about traitlets.

trulens_eval/trulens_eval/Example_TruBot.py
sfc-gh-pmardziel: to not have to recreate an identical `OptionalImports` instance again when doing `assert_installed` later

PR #1244, opened 2024-06-22
Use float instead of flooring in feedback score extraction + update docstring

sfc-gh-dhuang:
# Description

Use rounding instead of flooring in feedback score extraction + update docstring

Please include any other details of this change useful for TruLens developers.

## Type of change


sfc-gh-jreini:
Seems better to just keep the float rather than rounding or flooring

sfc-gh-dhuang:
> Seems better to just keep the float rather than rounding or flooring

sg - no objection from my side. (prompting the LLM to output integers only does not work that well in my experience)

Regardless of what we do now, I think the current state is not ideal and we are hoping to fast follow with proposals and POCs to achieve more constrained / structured outputs (scores)

sfc-gh-jreini (approved):

PR #1251, opened 2024-06-24
Demo Provider Toggle

sfc-gh-chu:
# Description

Adds a selector for LLM Provider, with values for Cortex and Replicate. This affects traffic for both app generation and feedback evaluation.


trulens_eval/examples/snowflake_demo/.env.template
```diff
@@ -2,11 +2,11 @@ REPLICATE_API_TOKEN=
 PINECONE_API_KEY=
 
 # Snowflake details
-SF_USER=
```
sfc-gh-chu, line 5: Renamed env vars since trulens cortex provider uses `SNOWFLAKE_*` by default and didn't want to duplicate these keys

trulens_eval/trulens_eval/feedback/provider/cortex.py
```diff
@@ -101,7 +101,7 @@ def _create_chat_completion(
             kwargs['messages'] = messages
 
         elif prompt is not None:
-            kwargs['messages'] = [{"role": "system", "content": prompt}]
+            kwargs['messages'] = [{"role": "user", "content": prompt}]
```
sfc-gh-chu, line 104: @sfc-gh-dhuang fyi this used to throw an error on my end that the last message must use the user role

trulens_eval/trulens_eval/guardrails/base.py
```diff
@@ -34,10 +34,17 @@ def __call__(self, func):
...
+            # TODO: this is a hack
+            main_input = None
+            for arg in args + tuple(kwargs.values()):
+                if isinstance(arg, str):
+                    main_input = arg
+                    break
             with ThreadPoolExecutor(max_workers=max(1, len(contexts))) as ex:
                 future_to_context = {
                     ex.submit(
-                        lambda context=context: self.feedback(args[1], context)
```
sfc-gh-chu, line 40: @sfc-gh-jreini `args[1]` may not exist when only kwargs are given. also we may want to think of a better way of selecting the prompt string

trulens_eval/examples/snowflake_demo/llm.py
```diff
@@ -1,26 +1,34 @@
...
-from feedback import f_small_local_models_context_relevance
 
-FRIENDLY_MAPPING = {
-    "Snowflake Arctic": "snowflake/snowflake-arctic-instruct",
-    "LLaMa 3 8B": "meta/meta-llama-3-8b",
-    "Mistral 7B": "mistralai/mistral-7b-instruct-v0.2",
+PROVIDER_MODELS = {
+    "Replicate":
+        {
+            "Snowflake Arctic": "snowflake/snowflake-arctic-instruct",
```
sfc-gh-dhuang, line 21: could we name this sth like `"Snowflake Arctic Instruct"` just for extra clarity

sfc-gh-jreini (approved):

trulens_eval/examples/snowflake_demo/llm.py
```diff
@@ -1,26 +1,34 @@
...
+            "LLaMa 3 8B": "meta/meta-llama-3-8b",
+            "Mistral 7B": "mistralai/mistral-7b-instruct-v0.2",
```
sfc-gh-dhuang, line 23: same here could we name this "Mistral 7B Instruct (v0.2)"?  Main reason I think of is to avoid user from confusing they are the same model (checkpoint)

sfc-gh-dhuang (approved):
LGTM with a few comments - not-merge-blocker though

trulens_eval/examples/snowflake_demo/llm.py
```diff
@@ -109,6 +116,14 @@ def _generate_stream_with_replicate(
         for t in stream_iter:
             yield str(t)
 
+    def _generate_with_cortex(self, model_name: str, model_input: dict):
+        response = get_provider("Cortex")._create_chat_completion(
+            prompt=model_input['prompt'],
+            temperature=model_input['temperature'],
+            model=model_name
+        )
+        yield str(response)
```
sfc-gh-dhuang, line 125: This still isn't streaming-when-first-token-is-available right? (or perhaps something changes over the last few days?)

sfc-gh-chu: Not streaming, but made this method a generator to match replicate. It will yield the full response just once

PR #1256, opened 2024-06-25
Fix small model scoring

sfc-gh-chu:
# Description

Fixes an issue with the small context relevance model consistently scoring around .5
tokenizer.convert_tokens_to_ids(tokenizer.tokenize)) does not include the SOS/EOS special tokens. Also includes the premise/hypothesis prefixes.


sfc-gh-dkurokawa (approved):

trulens_eval/examples/snowflake_demo/custom_feedback/small_local_models.py
```diff
@@ -32,13 +32,9 @@ def context_relevance(
...
-                    tokenizer.convert_tokens_to_ids(
-                        tokenizer.tokenize(f"{question} [SEP] {context}")
-                    )
-                ).reshape(1, -1)
-            ).logits.numpy()
-            if logit.size != 1:
-                raise ValueError("Unexpected number of results from model!")
-            logit = float(logit[0, 0])
+                **tokenizer(
+                    f"premise: {question}[SEP] hypothesis: {context}",
```
sfc-gh-jreini, line 36: late to the party here, but why are premise/hypothesis prefixes the right ones to use for a relevance model?

sfc-gh-dkurokawa: It's how Corey trained the model so that's why he needs to do it for inference.

PR #1257, opened 2024-06-25
Cache TruCustomApp

sfc-gh-chu:
# Description

Avoids creating redundant TruCustomApps, and redundant records in the trulens_app table. SF tables don't respect primary keys being unique so multiple records sharing the same app_id will cause downstream issues.

### DIFF
--- a/trulens_eval/examples/snowflake_demo/common_ui.py
+++ b/trulens_eval/examples/snowflake_demo/common_ui.py
@@ -1,7 +1,7 @@
 import json
 import pathlib
 import threading
-from typing import Dict
+from typing import Dict, Any
 
 from conversation_manager import ConversationManager
 # feedback functions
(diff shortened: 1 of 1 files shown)


trulens_eval/examples/snowflake_demo/common_ui.py
```diff
@@ -140,6 +140,14 @@ def get_tru_app_id(
     return f"app-prod-{model}{'-' + retriever if use_rag else ''}{('-retrieval-filter-' + str(retrieval_filter)) if use_rag else ''} (provider-{provider}-temp-{temperature}-topp-{top_p}-maxtokens-{max_new_tokens})"
 
 
+@st.cache_resource
```
sfc-gh-jreini, line 143: Will there be any issue here if the trulens app is cached when we change a configuration?

sfc-gh-dkurokawa: Corey can correct me if I'm wrong, but the docs seem to suggest this is caching based off the args as well and memoizing the results and the configuration stuff is in `metadata` so I imagine it's fine.

That said, I'm somewhat concerned by the use of `generator` in here. Mainly in that from what I can tell it doesn't seem to have any state, but doing this would effectively mean if we did store any state in there in the future (or if I'm wrong with my reading of `StreamGenerator`) it could be problematic across threads. I'll leave it to you Corey to decide whether that's critical enough given this is just a demo which we'll presumably chuck away on Thursday.

sfc-gh-dkurokawa (approved):

trulens_eval/examples/snowflake_demo/common_ui.py
sfc-gh-chu: Yup this will react to config changes. For this demo, if we were to make `StreamGenerator` stateful, we'd also have to modify the streamlit state anyway since streamlit will rerender/reinitialize `generator` on every streamlit state change. In that case, we could initialize `generator` using values from the streamlit state, similar to what we do with `TruCustomApp` here. 
Probably won't need to worry about that for this week though.

PR #1266, opened 2024-06-27
virtual record should use supplied perf, cost

sfc-gh-chu:
# Description

When supplied, VirtualRecord should use supplied perf and cost args

Please include any other details of this change useful for TruLens developers.

## Type of change


trulens_eval/trulens_eval/tru_virtual.py
```diff
@@ -407,12 +407,11 @@ def __init__(
         if (end_time - start_time).total_seconds() == 0.0:
             end_time += datetime.timedelta(microseconds=1)
 
-        if "cost" not in kwargs:
```
sfc-gh-chu, line 410: seems cost and perf were out of __init__'s kwargs into a named argument but this was not been updated

sfc-gh-pmardziel (approved):

PR #1277, opened 2024-07-03
Formatter pre-commit hook

sfc-gh-jreini:
# Description

Adding a precommit hook for formatting.

To add the change to your local development environment, run from root:
```bash
sh setup-hooks.sh
```

## Type of change


sfc-gh-dkurokawa (commented):
I only just read about the pre-commit stuff, but is there a reason you don't just change the `.git/hooks/pre-commit` file directly?

sfc-gh-jreini:
> I only just read about the pre-commit stuff, but is there a reason you don't just change the `.git/hooks/pre-commit` file directly?

My understanding is that the `.git` folder isn't versioned

sfc-gh-chu:
> > I only just read about the pre-commit stuff, but is there a reason you don't just change the `.git/hooks/pre-commit` file directly?
> 
> My understanding is that the `.git` folder isn't versioned. Also this way devs can choose to use or not this hook

Have we taken a look at [precommit](https://pre-commit.com/)? yapf and isort are both supported as a config there

sfc-gh-jreini:
> > > I only just read about the pre-commit stuff, but is there a reason you don't just change the `.git/hooks/pre-commit` file directly?
> > 
> > 
> > My understanding is that the `.git` folder isn't versioned. Also this way devs can choose to use or not this hook
> 
> Have we taken a look at [precommit](https://pre-commit.com/)? yapf and isort are both supported as a config there



> > > I only just read about the pre-commit stuff, but is there a reason you don't just change the `.git/hooks/pre-commit` file directly?
> > 
> > 
> > My understanding is that the `.git` folder isn't versioned. Also this way devs can choose to use or not this hook
> 
> Have we taken a look at [precommit](https://pre-commit.com/)? yapf and isort are both supported as a config there

Would that be easier? Since our format.sh is nonstandard, wouldn't we have to configure that anyway with precommit?

sfc-gh-jreini:
Closing, will by done by #1303

PR #1297, opened 2024-07-17
SDK: Group leaderboard results by metadata

sfc-gh-jreini:
# Description

Add parameter to `get_leaderboard` to display results grouped by metadata.

To be used in groundedness abstention testing.

## Type of change

  not work as expected)


trulens_eval/trulens_eval/tru.py
```diff
@@ -745,6 +746,42 @@ def get_leaderboard(
 
         return leaderboard
 
+    def get_leaderboard_grouped_by_metadata(
+        record_metadata_key: str = None,
```
sfc-gh-dhuang, line 750: how about we just allow an extra optional field sth like `sort_by_record_metadata_key` to the existing  `get_leaderboard` method above rather than adding a new method? 

`get_leaderboard_grouped_by_metadata` looks similar enough to me to the original `get_leaderboard()`

sfc-gh-jreini: Good point - done

sfc-gh-jreini: Good thinking - made this update!

trulens_eval/trulens_eval/tru.py
```diff
@@ -720,30 +721,46 @@ def get_records_and_feedback(
...
 
-        col_agg_list = feedback_cols + ['latency', 'total_cost']
+        df['meta'] = [
+            json.loads(df["record_json"][i])["meta"] for i in range(len(df))
+        ]
 
-        leaderboard = df.groupby('app_id')[col_agg_list].mean().sort_values(
-            by=feedback_cols, ascending=False
-        )
+        df[str(group_by_metadata_key)] = [
```
sfc-gh-dhuang, line 748: so if `group_by_metadata_key` is not specified, won't this create a column named `"None"` in the df? 

I am thinking it might be better to but this column creation under the condition only when  `group_by_metadata_key` is present

sfc-gh-jreini: It was creating the column but not using it - but you’re right it would be better to only create it at all if the key is present. Should be good now

sfc-gh-dhuang (approved):

PR #1298, opened 2024-07-17
Improve handling of abstention for groundedness

sfc-gh-jreini:
# Description

- [x] Create testing framework for abstentions
- [x] Improve groundedness handling of abstentions on tests for the case of rewarding all abstention
- [x] Improve groundedness handling of abstentions on tests for the case of considering answerability
- [x] Make docs clear on how abstentions are handled

Also fix bug with MAE calculation not triggering when the score was 0.

## Type of change

  not work as expected)


trulens_eval/trulens_eval/feedback/groundtruth.py
```diff
@@ -198,8 +198,8 @@ def mae(self, prompt: str, response: str, score: float) -> float:
         """
 
         expected_score = self._find_score(prompt, response)
-        if expected_score:
-            ret = abs(float(score) - expected_score)
+        if expected_score is not None:
```
sfc-gh-dhuang, line 201: good catch!

sfc-gh-dhuang: Since you are touching up / fixing this method, do you mind renaming it to something more accurate like `mae_term`? 

I noticed the method itself is actually calculating the absolute difference b/w ground truth and generated score and it's a misnomer / not MAE without the `aggregate(np.mean)` part.

sfc-gh-jreini: Renamed to `absolute_error` and updated the name in notebooks as well

sfc-gh-dhuang (approved):

PR #1309, opened 2024-07-29
API dump and deeper trulens_eval compatibility test

sfc-gh-pmardziel:
# Description

Some utilities to collect module contents and a dump of the current names in the `trulens_eval` package. This is to be used for backwards compatibility tests. Also added dump of the new `trulens` package to make sure we are aware of API changes when they happen.

Factored out module enumeration and related from static tests to import utilities. Making use of this for some backwards compatibility checks.

Added `api_test` that compares the qualified names and types of all public members of trulens modules and classes to a fixed golden set. In the future, will also add function signatures to this ground truth.

## Type of change

  not work as expected)
- [x] New Tests


sfc-gh-chu (approved):

trulens_eval/trulens_eval/utils/imports.py
```diff
@@ -682,3 +684,104 @@ def __exit__(self, exc_type, exc_value, exc_tb):
...
+
+        yield modinfo.name
+
+        if modinfo.ispkg:
+            for submod in get_modules(path / modinfo.name, startswith=None):
+                submodqualname = modinfo.name + "." + submod
+
+                if startswith is not None and not submodqualname.startswith(
+                        startswith):
+                    continue
```
sfc-gh-dkurokawa, line 715: how could this ever happen? It seems that given `submodqualname = modinfo.name + "." + submod` and we know from L704 if `startswith is not None` then `modinfo.name.startswith(startswith)` is `True`.

trulens_eval/trulens_eval/utils/imports.py
```diff
@@ -682,3 +684,104 @@ def __exit__(self, exc_type, exc_value, exc_tb):
...
+
+        startswith: If given, only modules that with this prefix are returned.
+
+    Returns:
+        Iterable of module names. These are fully qualified.
+    """
+
+    for modinfo in pkgutil.iter_modules([str(path)]):
+
+        if startswith is not None and not modinfo.name.startswith(startswith):
```
sfc-gh-dkurokawa, line 704: the way this is written, I guess you have to assume `startswith` isn't something like `a.b` since this will fail. Probably should state that in the pre-conditions and then check that `startswith` doesn't have any periods.

trulens_eval/trulens_eval/utils/imports.py
```diff
@@ -682,3 +684,104 @@ def __exit__(self, exc_type, exc_value, exc_tb):
...
+
+        if len(item) == 1:
+            # Type var names.
+            continue
+
+        qualname = modname + "." + item
+        yield (qualname, val)
+
+
+def get_definitions(
```
sfc-gh-dkurokawa, line 760: you don't actually use this function. What's it for?

sfc-gh-pmardziel: Moving this and related to the tests folder as it is only being used for tests (working on the test).

trulens_eval/trulens_eval/utils/imports.py
sfc-gh-pmardziel: Fixed.

trulens_eval/trulens_eval/utils/imports.py
sfc-gh-pmardziel: fixed

PR #1314, opened 2024-07-30
[SNOW-1565351] Allow connection through Snowflake.

sfc-gh-dkurokawa:
# Description

Instead of giving a `database_url` in `Tru` that maps to a Snowflake schema, allow use of the connection parameters. Ideally, we could use the `connections.toml` files as well, but this seems difficult to do because:
1. The Snowflake sqlalchemy connector doesn't accept this as a method of connection and requires knowing the password.
2. There are no tools that read the `connections.toml` files and expose the password in it. `snowflake.snowpark.Session` and `snowflake.connector.connect` don't expose this functionality for instance.
3. AFAICT, the `snowflake.snowpark.Session` and `snowflake.connector.connect` don't seem to expose how they read the `connections.toml` files and parse them at all. I don't want to duplicate this logic, especially since it could change.

I've also added a test to connect to Snowflake and set the proper secret variables in the CI-eval pipeline so that it should work on there. In the future we should test things out of the new test file I made.

## Type of change

  not work as expected)


sfc-gh-jreini:
Please also update the docs below to the extent they are impacted:

https://github.com/truera/trulens/blob/main/docs/trulens_eval/tracking/logging/where_to_log/log_in_snowflake.md

docs/trulens_eval/tracking/logging/where_to_log/log_in_snowflake.md
```diff
@@ -23,24 +23,23 @@ Snowflake SQLAlchemy can be installed from [PyPI](https://pypi.org/project/snowf
...
+    tru = Tru(
+        name="MyApp",
+        snowflake_connection_parameters={
+            account: "<account>",
+            user: "<user>",
+            password: "<password>",
+            database: "<database>",
+            warehouse: "<warehouse>",
+            role: "<role>",
+        },
```
sfc-gh-pdharmana, line 41: We need to support another workflow to go through snowsql config. Look into https://docs.snowflake.com/en/developer-guide/snowflake-python-api/snowflake-python-connecting-snowflake it supports two things. This would mean, we have to `import snowflake.connector`

trulens_eval/trulens_eval/tru.py
```diff
@@ -163,6 +168,8 @@ def __init__(
         database_prefix: Optional[str] = None,
         database_args: Optional[Dict[str, Any]] = None,
         database_check_revision: bool = True,
+        snowflake_connection_parameters: Optional[Dict[str, str]] = None,
+        name: Optional[str] = None,
```
sfc-gh-pdharmana, line 172: Is name optional/required?

trulens_eval/trulens_eval/tru.py
```diff
@@ -214,6 +239,51 @@ def __init__(
...
+        )
+
+    @staticmethod
+    def _create_snowflake_schema_if_not_exists(
+        snowflake_connection_parameters: Dict[str, str], schema_name: str
+    ):
+        from snowflake.core import CreateMode
+        from snowflake.core import Root
+        from snowflake.core.schema import Schema
+        from snowflake.snowpark import Session
```
sfc-gh-pdharmana, line 276: How do we evaluate between using snowpark vs python connector? https://docs.snowflake.com/en/developer-guide/python-connector/python-connector-connect

trulens_eval/trulens_eval/tru.py
```diff
@@ -214,6 +239,51 @@ def __init__(
...
+
+        Tru._create_snowflake_schema_if_not_exists(
+            snowflake_connection_parameters, schema_name
+        )
+        return URL(
+            account=snowflake_connection_parameters["account"],
+            user=snowflake_connection_parameters["user"],
+            password=snowflake_connection_parameters["password"],
+            database=snowflake_connection_parameters["database"],
+            schema=schema_name,
```
sfc-gh-dhuang, line 264: can you remind me of why we are always computing schema on the fly and not allowing specifying them in conn params?

docs/trulens_eval/tracking/logging/where_to_log/log_in_snowflake.md
sfc-gh-dkurokawa: yeah so I comment on this on the description of the PR. PTAL:
https://github.com/truera/trulens/pull/1314#:~:text=Instead%20of%20giving,it%20could%20change.

trulens_eval/trulens_eval/tru.py
sfc-gh-dkurokawa: Yeah for backward compatibility we should have it optional for the time being and still allow the old flow.

trulens_eval/trulens_eval/tru.py
sfc-gh-dkurokawa: Sorry, what do you mean by "evalulate"?

trulens_eval/trulens_eval/tru.py
sfc-gh-dkurokawa: because the schema will be decided based off the name of the "app" as per the eng doc.

sfc-gh-dhuang (approved):

trulens_eval/trulens_eval/tru.py
sfc-gh-dhuang: AFAICT both are needed cuz python connector, while it holds active session information and supports setting params on existing session, does not expose (at least I wasn't able to find a straightforward api) Session obj directly.

trulens_eval/trulens_eval/tru.py
```diff
@@ -163,6 +168,8 @@ def __init__(
         database_prefix: Optional[str] = None,
         database_args: Optional[Dict[str, Any]] = None,
         database_check_revision: bool = True,
+        snowflake_connection_parameters: Optional[Dict[str, str]] = None,
```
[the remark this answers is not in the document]
sfc-gh-dhuang: +1, lets make this a Pydantic model?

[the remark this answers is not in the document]
sfc-gh-dkurokawa: Is there some canonical snowflake connection object? I wasn't aware of one when I wrote this.

I did it this way because that's the way that Snowpark does it. See here:
https://docs.snowflake.com/en/developer-guide/snowpark/python/creating-session#connect-by-specifying-connection-parameters
I don't really want think we should do the `.toml` approach (i.e. the snowsql approach) for the reasons I listed in the description of the PR so that only really leaves this approach in that doc.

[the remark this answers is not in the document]
sfc-gh-chu: I was thinking of this: https://docs.snowflake.com/en/developer-guide/python-connector/python-connector-api#object-connection

Don't think we should read params from `.toml` but if users specify a connection object with their own params through snowpark I think we should adopt that standard

trulens_eval/trulens_eval/tru.py
sfc-gh-pdharmana: Just talked with sumit, seems like anything to create a stored proc or udf requires snowpark session. So this is good.

trulens_eval/trulens_eval/tru.py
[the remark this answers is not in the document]
sfc-gh-pdharmana: I think this is better. Dave's description covers this.

sfc-gh-pdharmana (approved):

trulens_eval/trulens_eval/tru.py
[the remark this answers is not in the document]
sfc-gh-dkurokawa: @coreyhu, unfortunately AFAICT, with that object I can't get the password out which I need to connect to Snowflake via sqlalchemy.

sfc-gh-chu (approved):

trulens_eval/trulens_eval/tru.py
sfc-gh-jreini: @sfc-gh-dkurokawa is it possible to add an option to use keypair auth instead of password? https://docs.snowflake.com/en/user-guide/key-pair-auth

PR #1332, opened 2024-08-09
[SNOW-1569937] App Versioning

sfc-gh-chu:
# Description
 - Fixes alembic DB work to support migrations to new app version schema
 - `Tru()` takes an app_name parameter which sets the schema name
 - Renames `app_id` > `app_version`

## Type of change

  not work as expected)


sfc-gh-pmardziel:
What is the effect of app name on non-snowflake databases?

tests/e2e/test_tru.py
```diff
@@ -328,7 +328,7 @@ def test_run_feedback_functions_wait(self):
 
         # Check that results were added to db.
         _, returned_feedback_names = tru.get_records_and_feedback(
-            app_ids=[tru_app.app_id]
+            app_versions=[tru_app.app_id]
```
sfc-gh-dhuang, line 331: shouldn't this be `tru_app.app_version` after the migration?

src/core/trulens/core/app/base.py
```diff
@@ -1475,7 +1475,7 @@ def dummy_record(
                 calls.append(sample_call)
 
         return mod_record_schema.Record(
-            app_id=self.app_id,
+            app_version=self.app_version,
```
sfc-gh-pdharmana, line 1478: How do we support the following in OSS?
App 1 -> v0, v1, v2, v3
App 2 -> v0, v1
App 3 -> v1, v2.

From snowflake we support this by saying, create a different schema for the app, and every version is in tables. But I don't see how the same will work with open source. Should we introduce app-id and app-version as two different concepts instead of this?

sfc-gh-chu: `Tru` now supports an `app_name` argument, which in OSS, would define the database filename (`default.sqlite` > 'app_1.sqlite'). 

The thought is that in OSS, different apps are stored in different DBs. Only in Snowflake would that distinction be made at the schema-level however.

sfc-gh-dhuang (approved):

sfc-gh-pmardziel:
Thoughts on generalizing this while being backwards compatible and not deprecating:

- Reinterpret "app_id" as a generic unique identifier not to be set by the user directly, only to link tables.
- Create a new column app_idents (or some thing else), of type JSON, and store things like name, version, deployment or anything else one wants to store there. We could even use the current tags column for this, as long as we expose to the database.
- Adjust methods to accept app_name, app_version and interpret that as app_idents={'name': app_name, 'version': app_version}. User can now query apps by either or both name and version, or anything else we decide to add to ident in the future.
- Store both app_idents and unique app_id in the AppDefinition. This is so that users can also query the database by exact id that they would have not otherwise set themselves.

src/core/trulens/core/database/migrations/versions/2_set_ff_id_not_null.py
```diff
@@ -0,0 +1,41 @@
+"""set feedback function id to not nullable.
+
+Revision ID: 2
+Revises: 1
+Create Date: 2023-08-10 23:11:37.405982
```
sfc-gh-dkurokawa, line 5: Can you update this date.

src/core/trulens/core/database/migrations/versions/3_rename_app_id_to_version_tag.py
```diff
@@ -0,0 +1,59 @@
...
+    # ### end Alembic commands ###
+
+
+def downgrade(config: Config) -> None:
+    prefix = config.get_main_option("trulens.table_prefix")
+
+    if prefix is None:
+        raise RuntimeError("trulens.table_prefix is not set")
+
+    # TODO: need to prepend prefix to all table names in the upgrades
```
sfc-gh-dkurokawa, line 47: get rid of TODO.

PR #1339, opened 2024-08-13
Set up a `run_location` that determines where a feedback function should be run if it's deferred.

sfc-gh-dkurokawa:
# Description
Set up a `run_location` that determines where a feedback function should be run if it's deferred.

This does introduce schema changes. AFAICT the way `trulens` is written, this isn't a problem but someone can correct me if I'm wrong.

## Type of change
  not work as expected)

### DIFF
--- a/src/core/trulens/core/database/migrations/versions/1_first_revision.py
+++ b/src/core/trulens/core/database/migrations/versions/1_first_revision.py
@@ -51,6 +51,7 @@ def upgrade(config) -> None:
         sa.Column("name", sa.Text(), nullable=False),
         sa.Column("cost_json", sa.Text(), nullable=False),
         sa.Column("multi_result", sa.Text(), nullable=True),
+        sa.Column("run_location", sa.Text(), nullable=True),
         sa.PrimaryKeyConstraint("feedback_result_id"),
     )
     op.create_table(
--- a/src/core/trulens/core/database/orm.py
+++ b/src/core/trulens/core/database/orm.py
@@ -270,6 +270,7 @@ def new_orm(base: Type[T]) -> Type[ORM[T]]:
             name = Column(Text, nullable=False)
             cost_json = Column(TYPE_JSON, nullable=False)
             multi_result = Column(TYPE_JSON)
+            run_location = Column(TYPE_ENUM, nullable=True)
 
             record = relationship(
                 "Record",
@@ -309,6 +310,9 @@ def new_orm(base: Type[T]) -> Type[ORM[T]]:
                         obj.cost, redact_keys=redact_keys
                     ),
                     multi_result=obj.multi_result,
+                    run_location=obj.run_location.value
+                    if obj.run_location is not None
+                    else None,
                 )
 
     configure_mappers()  # IMPORTANT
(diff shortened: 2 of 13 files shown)


src/core/trulens/core/schema/feedback.py
```diff
@@ -192,6 +202,9 @@ class FeedbackResult(serial.SerialModel):
         default_factory=datetime.datetime.now
     )
 
+    run_location: Optional[FeedbackRunLocation]
```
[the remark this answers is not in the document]
sfc-gh-pdharmana: That way we don't have to duplicate it in every row.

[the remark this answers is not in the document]
sfc-gh-dkurokawa: Yeah initially I did it this way since I wouldn't have to join the feedback definition table, but I guess that's not a big deal so I went ahead with your suggested approach. PTAL.

src/core/trulens/core/database/migrations/versions/1_first_revision.py
```diff
@@ -33,6 +33,7 @@ def upgrade(config) -> None:
         sa.Column(
             "feedback_definition_id", sa.VARCHAR(length=256), nullable=False
         ),
+        sa.Column("run_location", sa.Text(), nullable=True),
```
sfc-gh-pdharmana, line 36: General question for my understanding - how do we name the files 1_first_revision vs 2_second_revision? Also When is this script run?

src/core/trulens/core/database/orm.py
```diff
@@ -188,6 +189,9 @@ def parse(
             ) -> ORM.FeedbackDefinition:
                 return cls(
                     feedback_definition_id=obj.feedback_definition_id,
+                    run_location=None
```
sfc-gh-pdharmana, line 192: Formatting seems off here?

src/core/trulens/core/database/sqlalchemy.py
```diff
@@ -557,6 +561,27 @@ def _feedback_query(
...
+                )
+            )
+        else:
+            q = q.filter(
+                self.orm.FeedbackDefinition.run_location == run_location.value
+            )
+        q = q.filter(
+            self.orm.FeedbackResult.feedback_definition_id
+            == self.orm.FeedbackDefinition.feedback_definition_id
+        )
```
sfc-gh-pdharmana, line 583: I dont understand this filter.

src/core/trulens/core/schema/feedback.py
```diff
@@ -59,6 +59,16 @@ class FeedbackMode(str, Enum):
     `DEFERRED` mode using `tru.start_deferred_feedback_evaluator`."""
 
 
+class FeedbackRunLocation(str, Enum):
+    """Where the feedback evaluation takes place (e.g. locally, at a Snowflake server, etc)."""
+
+    LOCAL = "local"
```
sfc-gh-pdharmana, line 65: Rename this to `IN_APP` ?

sfc-gh-pdharmana (approved):
Overall looks good with some minor comments.

sfc-gh-pmardziel:
How is the snowflake run location configured? Is it all determined from the snowflake db connection set in Tru class?

sfc-gh-dkurokawa:
> How is the snowflake run location configured? Is it all determined from the snowflake db connection set in Tru class?

So the test case goes through it, but basically instead of wrapping your thing with `Feedback`, you wrap it with `SnowflakeFeedback`. It's also what was agreed on in the PRD/eng-doc.

PR #1347, opened 2024-08-16
Fix up migration script and docs.

sfc-gh-dkurokawa:
# Description
Fix up migration script and docs.

This should make it clearer on how to do DB migrations and also fix up some issues with it.

## Type of change
  not work as expected)
- [ ] New Tests

### DIFF
--- a/src/core/trulens/core/database/migrations/data.py
+++ b/src/core/trulens/core/database/migrations/data.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import json
 import traceback
-from typing import List
+from typing import Callable, Dict, List, Optional, Tuple
 
 from sqlalchemy import select
 from sqlalchemy.orm import Session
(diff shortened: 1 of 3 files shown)


src/core/trulens/core/database/migrations/script.py.mako
```diff
@@ -25,7 +25,7 @@ def upgrade(config) -> None:
     ${upgrades if upgrades else "pass"}
 
 
-def downgrade() -> None:
+def downgrade(config) -> None:
```
sfc-gh-chu, line 28: Can we make the TODO clearer?

src/core/trulens/core/database/migrations/data.py
```diff
@@ -16,41 +16,36 @@
 from trulens.core.schema import record as mod_record_schema
 from trulens.core.utils.pyschema import FunctionOrMethod
 
-sql_alchemy_migration_versions: List[str] = ["1"]
-"""DB versions that need data migration.
+sql_alchemy_migration_versions: List[int] = [1]
```
sfc-gh-chu, line 19: Does the typing play a role? Our migrations are ascending integers but are typed as strings and can be arbitrary strings

sfc-gh-chu (commented):
Can we also add a migration script for the feedback function ID being not nullable? It's currently out of sync with the ORM

src/core/trulens/core/database/migrations/data.py
sfc-gh-dkurokawa: Yes, but the code was actually assuming they were all ints anyway so I just made it more explicit.

sfc-gh-dkurokawa:
> Can we also add a migration script for the feedback function ID being not nullable? It's currently out of sync with the ORM

Isn't this part of your app versioning PR?

sfc-gh-chu (approved):

sfc-gh-dhuang:
> Can we also add a migration script for the feedback function ID being not nullable? It's currently out of sync with the ORM

One thing to note here is that I don't believe SQLite support ALTER a originally nullable column to be not nullable. For all I know SQLite ALTER statement is very limited and only supports basic stuff like adding a new column, dropping columns,  renaming a table / column.  

cc @sfc-gh-dkurokawa @sfc-gh-chu

PR #1354, opened 2024-08-16
Regardless of `FeedbackMode`, always defer feedbacks that are run on Snowflake.

sfc-gh-dkurokawa:
# Description
Regardless of `FeedbackMode`, always defer feedbacks that are run on Snowflake.

## Type of change
  not work as expected)
- [ ] New Tests


src/core/trulens/core/app/base.py
```diff
@@ -1400,29 +1400,38 @@ def _handle_record(
         if len(self.feedbacks) == 0:
             return []
 
-        # Add empty (to run) feedback to db.
```
sfc-gh-pmardziel, line 1403: Can you add some documentation to what is happening in this code like the line that was removed?

src/core/trulens/core/app/base.py
```diff
@@ -712,8 +712,8 @@ def _tru_post_init(self):
                     "Feedback logging requires `tru` to be specified."
                 )
 
-        if self.feedback_mode == mod_feedback_schema.FeedbackMode.DEFERRED:
-            for f in self.feedbacks:
+        for f in self.feedbacks:
```
sfc-gh-chu, line 715: why this change? `self.feedback_mode` should be constant

sfc-gh-dkurokawa: oh thanks, I forgot to add the other constraint (if it's a SnowflakeFeedback).

src/core/trulens/core/app/base.py
sfc-gh-dkurokawa: Done.

sfc-gh-chu (approved):

PR #1355, opened 2024-08-17
Fix alembic migration env.py:  un-set default sqlalchemy_url in alembic.ini

sfc-gh-dhuang:
# Description

Unset `sqlalchemy.url` so that it can be properly overridden by users env variable in `env.py`
```
# Get `sqlalchemy.url` from the environment.
if config.get_main_option("sqlalchemy.url", None) is None:
    config.set_main_option(
        "sqlalchemy.url", os.environ.get("SQLALCHEMY_URL", "")
    )
```  

## Type of change

  not work as expected)
- [ ] New Tests


sfc-gh-dhuang:
@sfc-gh-dkurokawa I was following the new `migration.md` (the command ```SQLALCHEMY_URL="sqlite:///../../../../../../default.sqlite" alembic revision --autogenerate -m "<short_description>" --rev-id "<next_integer_version>" ```) and I think the SQL alchemy URL env variable override will only work if we modify the env.py to handle both `None` and empty string, as it's default to `sqlalchemy.url = ""` in the current `alembic.ini`

sfc-gh-chu:
Should we just not define `sqlalchemy.url` and expect `None`? Alembic seems to parse the value as a literal string already, so this ends up being `'""'` as opposed to `''`

sfc-gh-dhuang:
> Should we just not define `sqlalchemy.url` and expect `None`? Alembic seems to parse the value as a literal string already, so this ends up being `'""'` as opposed to `''`

makes sense - updated

sfc-gh-chu (approved):

PR #1359, opened 2024-08-19
Cookbook reorganization

sfc-gh-jreini:
# Description

Display examples nicely in docs cookbook + update some other docs examples to display nicely

## Type of change

  not work as expected)
- [ ] New Tests


docs/trulens/tracking/instrumentation/llama_index.md
```diff
@@ -0,0 +1,177 @@
+# 🦙 LlamaIndex Integration
```
sfc-gh-chu, line 1: just curious why replace the notebooks with markdown? I thought some jupyter mkdocs extension does that for us

sfc-gh-jreini: Did this for these docs pages because they're not intended to be runnable notebooks anyways, and the extension does it for us but it's still kind of ugly compared to normal markdown. Perhaps there's some changes we can make to the extension to make that nicer in general, I can look into that.

PR #1363, opened 2024-08-20
Better groundedness UI display

sfc-gh-jreini:
# Description

Problem: Groundedness reasons are hard to understand and parse because they're all lumped into one dataframe cell.

Solution: Separate each groundedness statement eval into its own row in the feedback display dataframe, and highlight low scores. This makes it easy to understand which statements are hallucinations.

This is a UI-side change for feedbacks with the groundedness name.

Also removing the index display for all feedbacks.

## Before:

Screenshot 2024-08-20 at 3 31 22 PM

## After:
Screenshot 2024-08-20 at 3 29 32 PM

## Type of change

  not work as expected)
- [ ] New Tests


src/dashboard/trulens/dashboard/pages/Evaluations.py
```diff
@@ -389,12 +403,45 @@ def highlight(s):
...
+                            reasons = df["reasons"].iloc[0]
+                            statements = reasons.split("STATEMENT ")
+                            data = []
+                            for statement in statements[1:]:
+                                parts = statement.split("\n")
+                                criteria = parts[1].split(": ", 1)[1]
+                                supporting_evidence = parts[2].split(": ", 1)[1]
+                                score = int(parts[3].split(": ", 1)[1])
+                                data.append({
+                                    "Statement": criteria,
```
sfc-gh-pdharmana, line 417: This would be easier to understand if there is a sample redacted example above as a comment.

sfc-gh-jreini: what would the example be?

sfc-gh-pdharmana: This example is similar to the one you added in the other Pr.

"Hi. I'm here to help. The university of Washington is a public research university. UW's connections to major corporations in Seattle contribute to its reputation as a hub for innovation and technology"

- but in this case, what are the statements, critera/split

src/dashboard/trulens/dashboard/pages/Evaluations.py
```diff
@@ -389,12 +403,45 @@ def highlight(s):
                             df.meta.apply(lambda m: pd.Series(m))
                         ).drop(columns="meta")
 
-                        st.dataframe(
-                            df.style.apply(highlight, axis=1).format(
-                                "{:.2f}", subset=["result"]
+                        if "groundedness" in feedback_name.lower():
```
sfc-gh-dhuang, line 406: This will not work in the case where user names the groundedness feedback function to str not containing `"groundedness"` in it right?

sfc-gh-jreini: That's true. Could also cause an issue if a non-groundedness feedback function for some reason was named groundedness.

I think I'm okay with this, but open to suggestions.

epinzur:
This is a great start to improving the readability of the groundedness reasoning. 

An additional improvement would be to indicate which chunk (or chunks) of context contain the "Supporting Evidence from Source" fact.

src/dashboard/trulens/dashboard/pages/Evaluations.py
sfc-gh-jreini: Added comments to explain. let me know if this still needs more explanation

sfc-gh-jreini:
> This is a great start to improving the readability of the groundedness reasoning.
> 
> An additional improvement would be to indicate which chunk (or chunks) of context contain the "Supporting Evidence from Source" fact.

Thanks @epinzur !

With the groundedness function we have now, the LLM does not explicitly link the original chunk to the reasons; we can consider this in future updates to the feedback function itself.

src/dashboard/trulens/dashboard/pages/Evaluations.py
sfc-gh-dhuang: Can we add a note / comment in the code and we can come back to this or mark as a TODO for the community?

sfc-gh-dhuang (approved):
Please add a comment to the groundedness name string exact match issue for future improvement. thanks

PR #1381, opened 2024-08-26
Update connection parameters typing for Cortex provider

sfc-gh-jreini:
# Description

Dict typing fails when RSA key in connection parameters. Fix by changing to Any.

## Type of change

  not work as expected)
- [ ] New Tests

### DIFF
--- a/src/providers/cortex/trulens/providers/cortex/provider.py
+++ b/src/providers/cortex/trulens/providers/cortex/provider.py
@@ -1,5 +1,5 @@
 import json
-from typing import ClassVar, Dict, Optional, Sequence
+from typing import Any, ClassVar, Dict, Optional, Sequence
 
 import snowflake
 import snowflake.connector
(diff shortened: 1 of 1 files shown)


src/providers/cortex/trulens/providers/cortex/provider.py
```diff
@@ -15,15 +15,15 @@ class Cortex(
 
     DEFAULT_MODEL_ENGINE: ClassVar[str] = "snowflake-arctic"
 
-    connection_parameters: Dict
+    connection_parameters: Any
```
sfc-gh-dhuang, line 18: How about Dict[str, Any] or the worst case `dict` (which is what I'm using in my onging PR to unblock)

sfc-gh-jreini: Pydantic is detecting a dict with the RSA type value as PydanticUndefined rather than Dict/dict, so either of these fail

sfc-gh-dhuang: ah I see :/

sfc-gh-dhuang (approved):

src/providers/cortex/trulens/providers/cortex/provider.py
sfc-gh-dhuang: Approving the PR now but tagging Piotr @piotrm0 to see if you have seen this issue and might have some suggestions here for the typing. (I saw similar pydantic issue on my branch recently)

PR #1408, opened 2024-08-28
Davidkurokawa/handle zips correctly

sfc-gh-dkurokawa:
# Description
Handle zipped wheels for Snowflake stage correctly in the package.

Previously, we weren't actually including it into the wheel and it just happened to work because I was in the source directory.

## Type of change
  not work as expected)
- [ ] New Tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -188,16 +188,24 @@ zip-wheels:
 
 ## Step: Upload wheels to pypi
 # Usage: TOKEN=... make upload-trulens-instrument-langchain
-upload-%: clean build zip-wheels
+# In all cases, we need to clean, build, zip-wheels, then build again. The reason is because we want the final build to have the zipped wheels.
+upload-%: clean build
+    make build
 	poetry run twine upload -u __token__ -p $(TOKEN) dist/$*/*
 
-upload-all: clean build zip-wheels
+upload-all: clean build
+    make zip-wheels
+    make build
 	poetry run twine upload --skip-existing -u __token__ -p $(TOKEN) dist/**/*.whl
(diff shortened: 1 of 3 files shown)


zip_wheels.sh
```diff
@@ -27,6 +27,7 @@ zip_wheel() {
     rm -f $2
     cd ./tmp_build_zip
     unzip trulens*-py3-none-any.whl
+    rm -rf `find ./ -name snowflake_stage_zips`
```
sfc-gh-chu, line 30: wouldn't this be handled on line 38? Here it would remove all `snowflake_stage_zips` each time we zip a new package

Makefile
```diff
@@ -188,16 +188,25 @@ zip-wheels:
...
 
-upload-testpypi-%: clean build zip-wheels
+upload-testpypi-%: clean build
+	make zip-wheels
+	make build
 	poetry run twine upload -r testpypi -u __token__ -p $(TOKEN) dist/$*/*
 
-upload-testpypi-all: clean build zip-wheels
+upload-testpypi-all: clean build
+	make zip-wheels
```
sfc-gh-chu, line 209: is this change due to the ordering the commands are run?

sfc-gh-chu (approved):

Makefile
sfc-gh-dkurokawa: I need to enforce the ordering, and also run build twice.

zip_wheels.sh
sfc-gh-dkurokawa: this is removing it from the wheel that was built in the first round of `make build`.

PR #1420, opened 2024-08-29
Fix how golden tests are specified.

sfc-gh-dkurokawa:
# Description
Fix how golden tests are specified. Previously the `write_golden` and `load_golden` sets were using the location of the caller function to determine where the golden sets were, but this didn't work because the actual tests weren't the direct callers of the method sometimes.

I changed it so it's just based off the base directory so it's less confusing in case you have a helper function call the `write_golden` or `load_golden` functions like we did.

The tests were failing before, but they still are now, but for a totally different reason.

## Type of change
  not work as expected)
- [ ] New Tests


tests/e2e/test_serial.py
```diff
@@ -105,13 +107,13 @@ def test_app_serial(self):
...
                     actual=fdef.model_dump(),
-                    golden_filename=f"customapp_{name}.def.json",
+                    golden_filename=f"tests/e2e/golden/customapp_{name}.def.json",
                     skips=set(["feedback_definition_id", "id"]),
                 )
             with self.subTest(f"feedback result {name} serialization"):
                 self.assertGoldenJSONEqual(
                     actual=fres.model_dump(),
-                    golden_filename=f"customapp_{name}.result.json",
+                    golden_filename=f"tests/e2e/golden/customapp_{name}.result.json",
```
sfc-gh-dhuang, line 116: Shall we just make the common path prefix `tests/e2e/golden/` a constant

tests/unit/static/test_api.py
```diff
@@ -176,7 +176,9 @@ def test_api_trulens_eval_compat(self):
         """
         # TODEP: Deprecate after trulens_eval is removed.
 
-        golden_file = f"api.trulens_eval.{self.pyversion}.yaml"
+        golden_file = (
+            f"tests/unit/static/golden/api.trulens_eval.{self.pyversion}.yaml"
```
sfc-gh-dhuang, line 180: nit, same here could we just a constant? nbd I know it's just 2 occurrences

sfc-gh-dhuang (approved):
nit comments about using constants for paths. not blocking the merge

PR #1427, opened 2024-08-30
experimental flags

sfc-gh-pmardziel:
# Description

Adds `trulens.core.experimental` module for configuring experimental features.

Adds draft experimental feature policy into POLICIES.md and by includes, the policies docs page.

## Type of change

  not work as expected)
- [ ] New Tests


src/core/trulens/core/experimental/__init__.py
```diff
@@ -0,0 +1,441 @@
...
+
+class Feature(str, Enum):
+    """Experimental feature flags.
+
+    Use [TruSession.experimental_enable_feature][trulens.core.session.TruSession.experimental_enable_feature] to enable
+    these features.
+    """
+
+    OTEL_TRACING = "otel_tracing"
+    """OTEL-like tracing.
```
sfc-gh-jreini, line 32: please add a code example showing how to turn on otel_tracing

src/core/trulens/core/experimental/__init__.py
```diff
@@ -0,0 +1,441 @@
...
+        """The stored value."""
+
+        self.locked_by: Set[FrameType] = set()
+        """Set of frames (not in trulens) that have locked this value.
+
+        If empty, it has not been locked and can be changed.
+        """
+
+    @property
+    def is_locked(self) -> bool:
```
sfc-gh-jreini, line 70: should this be an _internal method instead? Or if user-facing, when would someone us this?

src/core/trulens/core/experimental/__init__.py
```diff
@@ -0,0 +1,441 @@
...
+
+        if isinstance(flags, dict):
+            for flag, val in flags.items():
+                self.set(flag, value=val, lock=lock)
+        else:
+            for flag in flags:
+                self.set(flag, value=True, lock=lock)
+
+
+class WithExperimentalSettings(
```
sfc-gh-jreini, line 235: difference between this `WithExperimentalSettings` and `Feature` classes is unclear to me. Can you explain?

src/core/trulens/core/experimental/__init__.py
sfc-gh-pmardziel: Added example to Feature docstring.

src/core/trulens/core/experimental/__init__.py
sfc-gh-pmardziel: Made the whole class private with _.

src/core/trulens/core/experimental/__init__.py
sfc-gh-pmardziel: You mean difference between `WithExperimentalSettings` and `ExperimentalSettings`? Here are all the classes:

- `Feature` -- enum for experimental features, public.
- `Setting` (now `_Setting`) -- class to manage the value to assign to some `Feature` (a boolean True for enabled, False for disabled).
- `ExperimentalSettings` (now `_ExperimentalSettings`) -- class to manage assignments to multiple features.
- `WithExperimentalSettings` -- a mixin class to include in another class to give it capabilities of `ExperimentalSettings` class while printout out messages to the user.

Without a mixin, we could have added the methods in it directly into `TruSession` but that would make a large complex class already more complex. Instead we "mixin" `WithExperimentalSettings`.

Alternatively we could have not had a mixin class at all, and used just `ExperimentalSettings` and try to mix that into TruSession instead. The problem there would be collisions with attributes/names. I want to only include methods that start with `experimental_` into `TruSession` and wanted to seperate the user printouts side of things from the feature value logic. So the solution in the PR has a separate `ExperimentalSettings` which I will rename to `_Settings` which does not print anything and has its own namespace, and a mixin class `WithExperimentalSettings` that provides access to `_Settings` through specialized methods that have user facing printouts.

src/core/trulens/core/experimental/__init__.py
```diff
@@ -0,0 +1,452 @@
...
+
+        If the value is not provided, lock the flag to its current value.
+
+        Raises:
+            ValueError: If the flag has already been locked to a different value.
+        """
+
+        return self.set(flag, value=value, lock=True)
+
+    def get(self, flag: Union[str, Feature], *, lock: bool = False) -> bool:
```
sfc-gh-pdharmana, line 187: Why do we need lock param for get operation?

src/core/trulens/core/database/sqlalchemy.py
```diff
@@ -99,6 +99,13 @@ class SQLAlchemyDB(DB):
     [ORM][trulens.core.database.orm.ORM] upon initialization.
     """
 
+    def __str__(self) -> str:
+        return f"SQLAlchemyDB({self.engine.url})"
+
+    # for DB's WithIdentString mixin
+    def _ident_str(self) -> str:
+        return self.engine.url
```
sfc-gh-pdharmana, line 107: This is not good, sometimes engine url will contain credentials?

sfc-gh-pdharmana (commented):
Overall looks good except few minor comments.

POLICIES.md
```diff
@@ -48,3 +48,43 @@ will occur at the introduction of the warning period.
...
+previews. Such features are indicated by the prefix `experimental_`. For
+example, the OTEL exporter for `TruSession` is specified with the
+`experimental_otel_exporter` parameter. Some features require additionally
+setting a flag before they are enabled. This is controlled by the
+`TruSession.experimental_{enable,disable}_feature` method:
+
+```python
+from trulens.core.session import TruSession
+session = TruSession()
+session.experimental_enable_feature("otel_tracing")
```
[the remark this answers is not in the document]
sfc-gh-pmardziel: Wanted to have the prefix "experimental_" intentionally as everything to do with experimental features has this prefix.

src/core/trulens/core/experimental/__init__.py
sfc-gh-pmardziel: That is actually the more common use case of locking. If you need to configure something differently based on the value of a feature flag and if this configuration is permanent so it cannot be changed, you'd lookup the feature flag while locking it:
```python
if settings.get(Feature.OTEL_TRACING, lock=True):
   # do instrumentation one way
else:
   # do instrumentation in some other way
```

src/core/trulens/core/database/sqlalchemy.py
sfc-gh-pmardziel: fixed to only show the database name

POLICIES.md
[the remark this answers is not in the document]
sfc-gh-jreini: imo this makes it seem like enabling the feature is experimental, not that the feature is experimental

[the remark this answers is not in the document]
sfc-gh-pmardziel: Why not both? Experimental feature enabling can be experimental until stability.

sfc-gh-pdharmana (approved):

PR #1441, opened 2024-09-04
fix async handling and other release pipeline failures

sfc-gh-pmardziel:
# Description

  - Fixes handling of instrumented async methods. Added/refined tests for these.
  
  - Fixes cost tracking of instrumented async methods. Added/refined tests for these.
  
  - Fixes main output guess for streams. Added/refined tests for this.
  
  - Fixes cost tracking for openai streams. However, only chunk counter is available in streams. Added tests for these.
  
  - In the process of the above, removed the stack-walking-based instrumentation in favor of `contextvars`. There are still a few uses of stack-walking but intend on getting on the path to remove those hacks completely.
  
  - Added reference/garbage collection tests to app-type test files. Moved this to another PR here: https://github.com/truera/trulens/pull/1470

  - Fixed spurious reference in dummy app that prevented garbage collection.

## Type of change

  not work as expected)
- [x] New Tests


docs/trulens/contributing/techdebt.md
```diff
@@ -12,8 +12,9 @@ lives.
 
 See `instruments.py` docstring for discussion why these are done.
 
-- We inspect the call stack in process of tracking method invocation. It may be
-  possible to replace this with `contextvars`.
+- __Stack walking removed in favor of contextvars in 1.0.2__. We inspect the
```
sfc-gh-jreini, line 15: nit: released 1.0.2 yesterday, bump to either 1.0.3 or 1.1?

sfc-gh-jreini:
Can you also update (or remove if async is considered an experimental feature) the async notebooks in `/expositional/frameworks`?

sfc-gh-jreini (commented):
Bigger questions:

- I see the note that stream chunks are tracked in place of tokens/cost for streaming apps. Is this a TruLens limitation or a Langchain/Llama-Index limitation?
- When we track stream chunks, do we propogate this in the frontend anywhere so users can see/use this information?
- Is it possible to capture time to first token? This is more impactful than total latency for streaming cases.
- Is it possible to record async custom apps?

sfc-gh-pmardziel:
> Bigger questions:
> 
>     * I see the note that stream chunks are tracked in place of tokens/cost for streaming apps. Is this a TruLens limitation or a Langchain/Llama-Index limitation?

Limitation of OpenAI which langchain/llamaindex both use. I don't see any cost information being produced from openai when streaming informing is enabled.

>     * When we track stream chunks, do we propogate this in the frontend anywhere so users can see/use this information?

Would have to be another PR with UI help.

>     * Is it possible to capture time to first token? This is more impactful than total latency for streaming cases.

Not in the current design but will think about it. Given we are moving to otel-like tracing soon, it might be done there instead.

>     * Is it possible to record async custom apps?

I believe so, will try to add example.

docs/trulens/contributing/techdebt.md
sfc-gh-pmardziel: fixed

sfc-gh-jreini:
Not a blocker for this PR, but for tracking usage of openai streaming looks like they recently enabled this by adding a parameter `stream_options: {"include_usage": true} `. Perhaps we can recommend this gets set by the user in their app if they want usage tracking?

https://community.openai.com/t/usage-stats-now-available-when-using-streaming-with-the-chat-completions-api-or-completions-api/738156

sfc-gh-pmardziel:
> Can you also update (or remove if async is considered an experimental feature) the async notebooks in `/expositional/frameworks`?

Updated the async notebooks but as async is different and independent from streaming, I created also 2 new notebooks for showing streaming which is what the notebooks were after to begin with. Also, have to pause fixing llama index stream handling until later. Will need to add as known issue:

"""
Llamaindex streams may produce records with the output "Lazy value of type <sometype> is incomplete."
"""

sfc-gh-pmardziel:
> Not a blocker for this PR, but for tracking usage of openai streaming looks like they recently enabled this by adding a parameter `stream_options: {"include_usage": true} `. Perhaps we can recommend this gets set by the user in their app if they want usage tracking?
> 
> https://community.openai.com/t/usage-stats-now-available-when-using-streaming-with-the-chat-completions-api-or-completions-api/738156

Looks like llamaindex already does this and there we do get that info recorded in our costs but I see langchain does not. Either way, we have no control over how they use openai unless we try to hack it under them.

sfc-gh-pmardziel:
> * Is it possible to record async custom apps?

I assumed here you meant streaming custom apps. I did add a quickstart notebook `custom_stream.ipynb` for this.

sfc-gh-jreini (approved):

PR #1461, opened 2024-09-12
json decode error workaround

sfc-gh-chu:
# Description

Workaround to snowflake-ml error handling code for JsonDecodeError

[Source](https://github.com/snowflakedb/snowflake-ml-python/blob/main/snowflake/ml/_internal/exceptions/exceptions.py#L37)

snowflake-ml wraps the original exception, and attempts to recreate it and assumes only string inputs, which does not hold for json.decode()

## Type of change

  not work as expected)
- [ ] New Tests

### DIFF
--- a/src/providers/cortex/trulens/providers/cortex/endpoint.py
+++ b/src/providers/cortex/trulens/providers/cortex/endpoint.py
@@ -129,25 +129,25 @@ class CortexEndpoint(Endpoint):
         # response is a snowflake dataframe instance or a list if the response is from cursor.fetchall()
         try:
             if isinstance(response, DataFrame):
-                response: dict = json.loads(response.collect()[0][0])
+                response_dict = json.loads(response.collect()[0][0])
             elif isinstance(response, list):
-                response: dict = json.loads(response[0][0])
+                response_dict = json.loads(response[0][0])
+        except json.JSONDecodeError:
+            response_dict = response
         except Exception as e:
             logger.error(f"Error occurred while parsing response: {e}")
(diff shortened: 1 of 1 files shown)


sfc-gh-jreini (approved):

src/providers/cortex/trulens/providers/cortex/endpoint.py
```diff
@@ -129,23 +129,25 @@ def handle_wrapped_call(
         # response is a snowflake dataframe instance or a list if the response is from cursor.fetchall()
         try:
             if isinstance(response, DataFrame):
-                response: dict = json.loads(response.collect()[0][0])
+                response_dict = json.loads(response.collect()[0][0])
             elif isinstance(response, list):
-                response: dict = json.loads(response[0][0])
+                response_dict = json.loads(response[0][0])
+        except json.JSONDecodeError:
+            response_dict = response
```
sfc-gh-dhuang, line 136: I'm not sure I quite follow here - so we swallow the json.JSONDecodeError exception and just pass the response_dict to continue (which might be of string type at this moment)

What are we trying to achieve or solve here?

sfc-gh-chu: So we can't raise a JSONDecodeError because of how snowflake.ml._external.exceptions does their error handling. So as a workaround, we pass the string forward which will inevitably raise a logger warning (line 150). We could also raise another kind of error but figured the warning message captures what's going on anyway

sfc-gh-dhuang (approved):

PR #1470, opened 2024-09-17
memory, threads, and async leakage testing

sfc-gh-pmardziel:
# Description

- Added reference/garbage collection test to check for memory leakage. It is to be used like this:
  
  ```python
  ca = ... # some app type earlier in the test
  
  ca_ref = weakref.ref(ca)
  del ca, ta, recorder, rec # delete ca and everything else that was used in the test
  self.assertCollected(ca_ref) # runs garbage collection and asserts that `ca` was collected.
  ```
  
  This can also print out the path from test globals to the to-be-collected object if an env flag is set. It is the explanation of why the object was not garbage collected. It shows something like this:
  
  ```python
     GetItem ['DummyApp']
     GetReferent /@35ae4e640 dict {'__module__': 'examples.dev.dummy_app.app', '__doc__': 'Dummy app implementation.\n\n    Contains:\n        - A memory component of type\n          [DummyMemory][examples.dev.dummy_app.memory.DummyMemory].\n\n        - A retriever component of type\n
     GetItem ['__init__']
     GetReferent /@328c237c0 dict {'__name__': 'examples.dev.dummy_app.app', '__doc__': None, '__package__': 'examples.dev.dummy_app', '__loader__': <_frozen_importlib_external.SourceFileLoader object at 0x328bdacc0>, '__spec__': ModuleSpec(name='examples.dev.dummy_app.app', loader=<_froze
     GetItem ['DummyStackTool']
     GetReferent /@35aea6300 dict {'__module__': 'examples.dev.dummy_app.tool', '__annotations__': {'last_stacks': typing.List[typing.Any]}, '__doc__': 'A tool that returns a rendering of the call stack when it is invoked.', 'last_stacks': [[<frame at 0x362cebc40, file '/Users/pmardziel/re
     GetItem ['last_stacks']
     GetIndex [0]
     GetIndex [6]
     GetReferent /@35a041430 DummyApp <examples.dev.dummy_app.app.DummyApp object at 0x35a041430>
  ```
  
  This means that the test globals (a dictionary) has a value at key `"DummyApp"` which has a reference to some object of type `dict` which contains a value at key `"__init__"` which has a reference to another `dict`, which has a `dict` at key `"DummyStackTool"` which has a reference to another `dict` with the key `"last_stacks"` pointing to an array whose 0th element is an array whose 6th element has a reference to `DummyApp` instance that we asserted should have been garbage collected.
  
  Added one example usage but commented it out as it is currently failing. Fixes are in another PR.

- Added tests for non-cleaned up threads and tasks to the shared test class and renamed it as it is no longer about JSON. These can optionally result in test failures if some specific env flags are set.

## Type of change

  not work as expected)
- [x] New Tests


src/core/trulens/core/utils/serial.py
```diff
@@ -159,6 +158,15 @@ def set(self, obj: Any, val: Any) -> Any:
         """
         raise NotImplementedError()
 
+    def get_sole_item(self, obj: Any) -> Any:
+        all_objects = list(self.get(obj))
+
+        assert (
```
sfc-gh-dkurokawa, line 164: is there a reason you're not doing an `if`-`raise` kinda thing? Technically I think all the `assert` can be turned off.

src/core/trulens/core/utils/text.py
```diff
@@ -89,6 +89,31 @@ def format_quantity(quantity: float, precision: int = 2) -> str:
...
+    Returns:
+        str: The formatted quantity.
+    """
+
+    units = ["", "KB", "MB", "GB", "TB", "PB", "EB", "ZB", "YB"]
+    unit_idx = max(
+        0,
+        min(
+            len(units) - 1,
+            int(math.floor(0 if size == 0 else math.log10(abs(size)) / 3)),
```
sfc-gh-dkurokawa, line 109: haha are you expecting negative sizes or something?

src/core/trulens/core/utils/text.py
```diff
@@ -89,6 +89,31 @@ def format_quantity(quantity: float, precision: int = 2) -> str:
...
+    [millify](https://github.com/azaitsev/millify).
+
+    Args:
+        size: The quantity to format.
+
+    Returns:
+        str: The formatted quantity.
+    """
+
+    units = ["", "KB", "MB", "GB", "TB", "PB", "EB", "ZB", "YB"]
```
sfc-gh-dkurokawa, line 104: shouldn't the first one be B?

tests/test.py
```diff
@@ -475,8 +502,131 @@ def recur(j1, j2, path, unordered=False):
...
+
+        # Enable WITH_REF_PATH to see printout of why the given ref was not
+        # GC-ed.
+        if (
+            obj is not None
+            and os.environ.get(WITH_REF_PATH_VAR, None) is not None
+        ):
+            caller_globals = python_utils.caller_frame(offset=1).f_globals
+
+            with self.subTest(part="reference path"):
```
sfc-gh-dkurokawa, line 548: what's the point of this `subTest`? Can't we just move this whole `if` block above L537 then have no subtests?

tests/test.py
```diff
@@ -475,8 +502,131 @@ def recur(j1, j2, path, unordered=False):
...
+        obj = ref()
+
+        with self.subTest(part="garbage collection"):
+            self.assertTrue(obj is None, msg)
+
+        # Enable WITH_REF_PATH to see printout of why the given ref was not
+        # GC-ed.
+        if (
+            obj is not None
+            and os.environ.get(WITH_REF_PATH_VAR, None) is not None
```
sfc-gh-dkurokawa, line 544: here and a few places below you do this
```
os.environ.get(ENV_VAR, None) is not None
```
thing, but if `ENV_VAR` is `0` or `False` or something, this is still going to evaluate as `True`. The descriptions of the env variables say "Env var that when set to true..." so this doesn't seem like that.

tests/test.py
```diff
@@ -475,8 +502,131 @@ def recur(j1, j2, path, unordered=False):
...
+        if running_tasks:
+            if os.environ.get(TEST_TASKS_CLEANUP_VAR, None) is not None:
+                with self.subTest(part="running tasks"):
+                    raise AssertionError(
+                        f"Tasks still running: {running_tasks}"
+                    )
+            else:
+                print(f"Tasks still running: {running_tasks}")
+
+        non_main_threads = []
```
sfc-gh-dkurokawa, line 588:
```
non_main_threads = [thread for thread in threading.enumerate() if thread != threading.main_thread()]
```

tests/test.py
```diff
@@ -475,8 +502,131 @@ def recur(j1, j2, path, unordered=False):
...
+            caller_globals = python_utils.caller_frame(offset=1).f_globals
+
+            with self.subTest(part="reference path"):
+                # Show the reference path to the given ref.
+                print(f"Reference path from globals to {ref}:")
+                path = find_path(id(caller_globals), id(obj))
+                self.assertIsNotNone(path, "Couldn't find reference path.")
+                print_referent_lens(path)
+
+    def tearDown(self):
```
sfc-gh-dkurokawa, line 555: might want to call the super version too since it may do some good stuff? I have no idea.

tests/test.py
```diff
@@ -475,8 +502,131 @@ def recur(j1, j2, path, unordered=False):
...
+            if os.environ.get(TEST_THREADS_CLEANUP_VAR, None) is not None:
+                with self.subTest(part="non-main threads"):
+                    raise AssertionError(
+                        f"Non-main threads still running: {non_main_threads}"
+                    )
+            else:
+                print(f"Non-main threads still running: {non_main_threads}")
+
+    @classmethod
+    def tearDownClass(cls):
```
sfc-gh-dkurokawa, line 603: might want to call the super version too since it may do some good stuff? I have no idea.

tests/test.py
```diff
@@ -475,8 +502,131 @@ def recur(j1, j2, path, unordered=False):
...
+                running_tasks.append(task)
+                print("    " + str(task))
+                task.print_stack()
+
+        except Exception as e:
+            print(f"  No running loop? {e}")
 
-    pass
+        print("  Remaining non-main threads:")
+        non_main_threads = []
```
sfc-gh-dkurokawa, line 628: you don't use this at all

src/core/trulens/core/utils/text.py
sfc-gh-pmardziel: coped the code from the float one and didn't think enough about it

src/core/trulens/core/utils/text.py
sfc-gh-pmardziel: yes, good catch

src/core/trulens/core/utils/serial.py
sfc-gh-pmardziel: did not know this. changed to explicit raise

src/core/trulens/core/utils/text.py
sfc-gh-pmardziel: fixed

tests/test.py
sfc-gh-pmardziel: changed this to be testing for truth instead of just definition

tests/test.py
sfc-gh-pmardziel: Wanted the printout of the path to be under a seperate (sub)test than the GC collection (sub)test and only show up when the env var is set. There is also the possibility that there is something wrong with the path computation that may fail especially if the state of the interpreter is not settled at this point (i.e. other threads are still running and being activaly collected).

tests/test.py
sfc-gh-pmardziel: added at bottom

tests/test.py
sfc-gh-pmardziel: uncleaned copy, thanks

tests/test.py
sfc-gh-pmardziel: added at end

tests/test.py
```diff
@@ -91,9 +91,9 @@ def optional_test(testmethodorclass):
     all optional packages have been installed.
     """
 
-    return unittest.skipIf(not os.environ.get(OPTIONAL_VAR), "optional test")(
-        testmethodorclass
-    )
+    return unittest.skipIf(
```
sfc-gh-dkurokawa, line 94: I don't believe this fixes this. Like don't you want it to evaluate to `False` if you set the env variable to 0 or False or something? 
```
(trulens_3_11) dkurokawa@KJ6R9J3J4N trulens % cat cow_testing.py                 
import os

if os.environ.get("COW_ENV_VAR", False):
    print("True")
else:
    print("False")
(trulens_3_11) dkurokawa@KJ6R9J3J4N trulens % unset COW_ENV_VAR                 
(trulens_3_11) dkurokawa@KJ6R9J3J4N trulens % python cow_testing.py 
False
(trulens_3_11) dkurokawa@KJ6R9J3J4N trulens % COW_ENV_VAR=False python cow_testing.py
True
(trulens_3_11) dkurokawa@KJ6R9J3J4N trulens % COW_ENV_VAR=0 python cow_testing.py    
True
```

sfc-gh-pmardziel: It is using python's view of true and I believe any non-empty string is true. Changed it to cast to bool first to further reduce confusion.

sfc-gh-pmardziel: * changed to a test for several specific strings

tests/test.py
sfc-gh-pmardziel: did this

sfc-gh-dkurokawa (approved):

PR #1474, opened 2024-09-17
add_dataframe method + quickstart

sfc-gh-jreini:
# Description

- [x] Add new `add_dataframe` method for virtual apps to quickly ingest data from a dataframe.
- [x] Add helper method for accessing context in the structure stored by `add_dataframe`.

## Type of change

  not work as expected)
- [ ] New Tests


src/core/trulens/apps/virtual.py
```diff
@@ -570,5 +576,56 @@ def add_record(
...
+                        retriever.get_context: dict(
+                            args=[record["prompt"]], rets=context_rets
+                        )
+                    },
+                    main_input=record["prompt"],
+                    main_output=record["response"],
+                )
+            else:
+                rec = VirtualRecord(
+                    main_input=record["prompt"],
```
sfc-gh-pdharmana, line 619: Keep this similar to ragas, question, contexts, response. I think

sfc-gh-jreini: We shouldn't break Virtual Record here by changing the name, but I updated the add_dataframe method to take question/answer/contexts.

Noting that this is a departure from the ground truth data naming, which requires `query` column name. This means if I want to use the same dataframe with some ground truths for ground truth evaluation, I'd need to rename the column for the ground truth calculation (not ideal).

sfc-gh-jreini: IMO we should go with `query`, `response`, `contexts`. Are you good with this @sfc-gh-pdharmana ?

sfc-gh-pdharmana (approved):

PR #1475, opened 2024-09-17
use python complete instead of sql call in Arctic Quickstart

sfc-gh-jreini:
# Description

use python complete instead of sql call in Arctic Quickstart

## Type of change

  not work as expected)
- [ ] New Tests


sfc-gh-jreini:
closed because python version still does not return usage info needed for cost tracking

sfc-gh-dhuang:
> closed because python version still does not return usage info needed for cost tracking

https://snowflakecomputing.atlassian.net/browse/SNOW-1677642 I created a task to track so that we don't forget :)

PR #1490, opened 2024-09-20
small perf improvements

sfc-gh-chu:
# Description

 - Fix app caching and updates sqlalchemy bulk insert api.
 - Call stack related speedups

## Type of change

  not work as expected)
- [ ] New Tests

### DIFF
--- a/src/core/trulens/core/database/sqlalchemy.py
+++ b/src/core/trulens/core/database/sqlalchemy.py
@@ -378,7 +378,7 @@ class SQLAlchemyDB(DB):
                 self.orm.Record.parse(r, redact_keys=self.redact_keys)
                 for r in records
             ]
-            session.bulk_save_objects(records_list)
+            session.add_all(records_list)
             logger.info(f"{UNICODE_CHECK} added record batch")
             # return record ids from orm objects
             return [r.record_id for r in records_list]
@@ -583,7 +583,7 @@ class SQLAlchemyDB(DB):
                 self.orm.FeedbackResult.parse(f, redact_keys=self.redact_keys)
                 for f in feedback_results
             ]
-            session.bulk_save_objects(feedback_results_list)
+            session.add_all(feedback_results_list)
             return [f.feedback_result_id for f in feedback_results_list]
 
     def _feedback_query(
@@ -849,7 +849,7 @@ class SQLAlchemyDB(DB):
                     )
                     ground_truths_to_insert.append(new_ground_truth)
 
-            session.bulk_save_objects(ground_truths_to_insert)
+            session.add_all(ground_truths_to_insert)
             return [gt.ground_truth_id for gt in ground_truths]
 
     def get_ground_truth(
--- a/src/core/trulens/core/schema/app.py
+++ b/src/core/trulens/core/schema/app.py
@@ -352,6 +352,8 @@ class AppDefinition(pyschema.WithClassInfo, serial.SerialModel):
         tp = TP()
 
         for ffunc in feedback_functions:
+            ffunc.run(app=app, record=record)
+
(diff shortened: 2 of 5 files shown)


src/core/trulens/core/schema/app.py
```diff
@@ -352,6 +352,8 @@ def _submit_feedback_functions(
         tp = TP()
 
         for ffunc in feedback_functions:
+            ffunc.run(app=app, record=record)
```
sfc-gh-pdharmana, line 355: Why is this needed?

src/core/trulens/core/database/sqlalchemy.py
```diff
@@ -378,7 +378,7 @@ def batch_insert_record(
                 self.orm.Record.parse(r, redact_keys=self.redact_keys)
                 for r in records
             ]
-            session.bulk_save_objects(records_list)
+            session.add_all(records_list)
```
sfc-gh-pdharmana, line 381: Whats difference between add_all and bulk_save objects?

src/core/trulens/core/schema/app.py
sfc-gh-chu: Ah some working changes got checked in. Will revert this

src/core/trulens/core/database/sqlalchemy.py
sfc-gh-chu: Functionally the same thing. [bulk_save_objects](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session.bulk_save_objects) is the legacy API in SQLAlchemy 2.0 so updated this to use the `add_all` API.

src/core/trulens/core/utils/python.py
```diff
@@ -535,37 +543,43 @@ def merge_stacks(
...
             pass
 
     return ret
 
 
-def stack_with_tasks() -> Sequence[FrameType]:
+def stack_generator(offset: int = 0) -> Iterable[FrameType]:
+    frame = inspect.currentframe()
+    for _ in range(offset):
+        assert frame is not None
```
sfc-gh-dhuang, line 562: Do we have to use assert here? or can we just raise exception with error message

src/core/trulens/core/utils/python.py
```diff
@@ -437,10 +445,10 @@ def external_caller_frame(offset=0) -> FrameType:
     Raises:
         RuntimeError: If no such frame is found.
     """
-
-    for finfo in inspect.stack()[offset + 1 :]:
-        if not finfo.frame.f_globals["__name__"].startswith("trulens"):
-            return finfo.frame
+    gen = list(stack_generator(offset + 2))
```
sfc-gh-dhuang, line 448: casting to list shouldn't be necessary here right? and it's more memory efficient if you just iterate over the generator without forcing to evaluate the entire stack all at once

sfc-gh-dhuang:
left a few comments but minor. mostly LGTM

src/core/trulens/core/utils/python.py
sfc-gh-chu: I was originally thinking the same. Then realized that because the generator is called lazily, its call stack would change and reflect the generator callee's call stack, not the call stack of `external_caller_frame`

sfc-gh-dhuang: ah got it sg

sfc-gh-dhuang (approved):

src/core/trulens/core/utils/python.py
sfc-gh-chu: actually we may be able to get around this by passing in the current frame so that inspect.currentframe() is guarenteed to be invoked at the right time

PR #1498, opened 2024-09-23
Make nltk punkt_tab available for server-side eval

sfc-gh-dhuang:
# Description

The directory `nltk_data/tokenizers/punkt_tab` is then going to be zipped and uploaded to Snowflake stages for server-side eval, as it's part of the `trulens-feedback` zipped wheel

## Type of change

  not work as expected)
- [ ] New Tests


src/feedback/trulens/feedback/llm_provider.py
```diff
@@ -1446,6 +1447,15 @@ def _remove_trivial_statements(self, statements: List[str]) -> List[str]:
             )
         )
 
+    def _ensure_punkt_installed(self, download_dir: str):
+        from nltk.data import find
+
+        try:
+            # Try to locate the 'punkt' directory in the configured NLTK data path
+            find("tokenizers/punkt_tab")
+        except LookupError:
```
sfc-gh-pdharmana, line 1456: Maybe add a print statement?to know we're proceeding with this codepath. Also if nltk_data is now part of the code, would it mean this error should never happen?

sfc-gh-dhuang: yeah under normal cases it should no longer happen. 

added a log warning msg now

sfc-gh-dkurokawa (commented):
Few points/questions:
1. Can you add a test testing this with a Snowflake deployment?
2. Does the `punkt` downloaded stuff never change?
3. I'd think it'd be better to just have as part of the `zip_wheels.sh` script something that downloads the `nltk` data and zips it then. That way it's always up-to-date with the associated `nltk` version and also we won't have to store the code in our repo. Kinda like with the `snowflake-sqlalchemy` stuff.

src/feedback/trulens/feedback/llm_provider.py
```diff
@@ -1446,12 +1448,35 @@ def _remove_trivial_statements(self, statements: List[str]) -> List[str]:
             )
         )
 
+    def _ensure_punkt_installed(self):
+        from nltk.data import find
+
+        try:
+            # Try to locate the 'punkt' directory in the configured NLTK data path
+            home_path = os.path.expanduser("~")
+            nltk_data_dir = os.path.join(home_path, "nltk_data")
```
sfc-gh-chu, line 1457: Does this handle staged data?

sfc-gh-dhuang: good point, switching to use current path, rather than home path should work w/ staged files in internal stage.  In the process of updating e2e tests as well to ensure this works w/ SSE

sfc-gh-dhuang:
With some further testing, I don't think adding nltk punkt_tab to Snowflake internal stages is as trivial as making them available in internal user stages. 

This is because the stored proc would not have access to any local file system, but only staged data. While we could access and query staged data, the `find` method in nltk package invoked by the `sent_tokenizer` is always going to be looking for language files in the runtime's file system. ref: https://www.nltk.org/_modules/nltk/tokenize/punkt.html and see the line `lang_dir = find(f"tokenizers/punkt_tab/{lang}/")` in the Punkt constructor. 




For context, the original intention of this PR was to avoid LLM hallucination by performing sentence splitting with punkt in server-side eval. And after taking closer look and doing more meta-evaluation on the groundedness feedback function, we realized that a large number of hallucinated feedback runs actually did not come from the sentence splitting part, but sometimes from the trivial statements handling part. And the performance varies greatly between different backing LLM-as-judges.  `snowflake-arctic` does not perform particularly well in this regards (68% recall in our internal benchmark)

I'm closing this PR for now, and instead we will switch to `mistral-large` as default LLM in cortex provider, as snowflake 





cc @sfc-gh-pdharmana @sfc-gh-jreini @sfc-gh-dkurokawa @sfc-gh-chu @sfc-gh-pmardziel

PR #1523, opened 2024-09-30
Endpoints as Multitons and Singletons represented as metaclasses

sfc-gh-chu:
# Description
 - Endpoints are no longer singletons. This allows users to define 2 OpenAI endpoints with different base urls for example.
 - Define Singletons as metaclasses so __init__ is only invoked once

## Type of change

  not work as expected)
- [ ] New Tests

----

> [!IMPORTANT]
> Refactor endpoint and session management by replacing singleton pattern with metaclasses and allowing multiple endpoint instances.
> 
>   - **Behavior**:
>     - Endpoints are no longer singletons, allowing multiple instances with different configurations.
>     - Introduces `InstanceRefMixin` for managing instance references in `endpoint.py`.
>     - Replaces `SingletonPerName` with `InstanceRefMixin` in `Endpoint` and other classes.
>   - **Metaclasses**:
>     - Defines `Singleton` and `PydanticSingleton` metaclasses in `python.py` to manage singleton behavior without re-running `__init__`.
>     - Applies `PydanticSingleton` to `TruSession` in `session.py`.
>   - **Refactoring**:
>     - Removes `__new__` methods related to singleton logic in `endpoint.py`, `session.py`, and other endpoint files.
>     - Updates `handle_wrapped_call` and other methods to accommodate non-singleton endpoints.
>   - **Misc**:
>     - Minor logging and string formatting changes in `sqlalchemy.py` and other files.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=truera%2Ftrulens&utm_source=github&utm_medium=referral)<sup> for b63674a0aef19ae2b52748233edb56ce8eb7d3e3. It will automatically update as commits are pushed.</sup>


src/core/trulens/core/session.py
```diff
@@ -193,7 +188,6 @@ def __init__(
                 logger.warning(
                     "TruSession was already initialized. Cannot change database configuration after initialization."
                 )
-                self.warning()
             return
```
sfc-gh-pmardziel, line 191: Is the whole condition not required any more? It was for the singletonbyname to abort a second call to `__init__` but if I understand right, the metaclass will prevent the second init now?

sfc-gh-pmardziel: Condition starting `if python.safe_hasattr(self, "connector"):`.

src/core/trulens/core/utils/python.py
```diff
@@ -1276,24 +1191,50 @@ def delete_singleton_by_name(
...
+
+
+class PydanticSingleton(type(pydantic.BaseModel), SingletonPerName):
+    pass
+
+
+class InstanceRefMixin:
+    _instance_refs: Dict[
+        Type, List[weakref.ReferenceType[InstanceRefMixin]]
+    ] = defaultdict(list)
```
sfc-gh-pmardziel, line 1229: Change list to WeakSet so the list doesn't get clogged up with stale refs?

src/core/trulens/core/utils/python.py
```diff
@@ -1276,24 +1191,50 @@ def delete_singleton_by_name(
...
+        k = cls_name, name
+        if k in SingletonPerName._singleton_instances:
+            del SingletonPerName._singleton_instances[k]
         else:
-            logger.warning("Instance %s not found in our records.", self)
+            logger.warning("Instance %s not found:", obj)
+
+
+class PydanticSingleton(type(pydantic.BaseModel), SingletonPerName):
+    pass
```
sfc-gh-pmardziel, line 1223: Some documentation here as to why this is necessary.

src/core/trulens/core/utils/python.py
```diff
@@ -1148,118 +1150,31 @@ def rewrap(obj_):
...
-        *args,
-        name: Optional[str] = None,
-        **kwargs,
-    ) -> SingletonPerName:
+    def __call__(cls, *args, name: Optional[str] = None, **kwargs):
         """
         Create the singleton instance if it doesn't already exist and return it.
         """
-
         k = cls.__name__, name
```
sfc-gh-pmardziel, line 1166: This was a problem before but might as well fix here. Can you make the first part of the key be "cls.__module__ + "." + cls.__name__` so the same class name in different modules will not cause an issue.

src/core/trulens/core/utils/python.py
```diff
@@ -1148,118 +1150,31 @@ def rewrap(obj_):
 
 
 # Class utilities
-
-
-@dataclasses.dataclass
-class SingletonInfo(Generic[T]):
-    """
-    Information about a singleton instance.
+class SingletonPerName(type):
```
sfc-gh-pmardziel, line 1153: Change name so it is obvious that it needs to be set as metaclass instead of parent. Perhaps `MetaSingletonPerName` ?

src/core/trulens/core/utils/python.py
```diff
@@ -1148,118 +1150,31 @@ def rewrap(obj_):
...
-
-    def __init__(self, name: str, val: Any):
-        self.val = val
-        self.cls = val.__class__
-        self.name = name
-        self.frameinfo_codeline = code_line(
-            caller_frameinfo(offset=2), show_source=True
-        )
-
-    def warning(self):
```
sfc-gh-pmardziel, line 1188: Can the warning mechanism be integrated into meta `__call__`? Track the frame for debugging when singleton created and then add a special method for singleton children that checks whether a warning should be issued:

```python
def init_ignored(cls, frame, *args, **kwargs):
     if # some args are not none:
         print("Singleton already created for class .... # rest of message)
```

tests/utils.py
```diff
@@ -359,7 +359,9 @@ def get_class_members(
             (name, field.default, field.annotation)
             for name, field in class_.model_fields.items()
         ]
-    elif hasattr(class_, "__fields__"):  # pydantic.v1.BaseModel
+    elif hasattr(class_, "__fields__") and hasattr(
+        class_.__fields__, "items"
```
sfc-gh-pmardziel, line 363: Add a note as to what happened here to necessitate the additional check.

src/core/trulens/core/session.py
sfc-gh-pmardziel: Although it would be nice to have a warning in case someone wants to reconfigure the session. See another comment about the warnings.

src/core/trulens/core/utils/python.py
```diff
@@ -1148,118 +1150,31 @@ def rewrap(obj_):
...
-                f"Instance of singleton type/name {k} does not exist."
-            )
-
-    def __new__(
-        cls: Type[SingletonPerName],
-        *args,
-        name: Optional[str] = None,
-        **kwargs,
-    ) -> SingletonPerName:
+    def __call__(cls, *args, name: Optional[str] = None, **kwargs):
```
sfc-gh-pmardziel, line 1162: Is this a classmethod or is that implied by this being a Metaclass?

src/core/trulens/core/session.py
sfc-gh-chu: Yea this would only invoke `__init__` once. We can throw a warning though

src/core/trulens/core/utils/python.py
sfc-gh-chu: yea it's implied

src/core/trulens/core/utils/python.py
```diff
@@ -1223,85 +1144,55 @@ def delete_singleton_by_name(
...
-            info: SingletonInfo = SingletonInfo(val=instance)
-            Singleton._instances[k] = info
-        else:
-            info = Singleton._instances[k]
-
-        obj = info.val
-        assert isinstance(obj, cls)
-        return obj
+    _instance_refs: Dict[
+        Type, weakref.WeakSet[weakref.ReferenceType[InstanceRefMixin]]
```
sfc-gh-pmardziel, line 1185: Don't need to stack the refs. Store just the `InstanceRefMixin` inside the WeakSet.

src/core/trulens/core/utils/python.py
```diff
@@ -1223,85 +1144,55 @@ def delete_singleton_by_name(
...
 
-    def delete_singleton(self):
-        """
-        Delete the singleton instance. Can be used for testing to create another
-        singleton.
-        """
-        k = self.__class__.__name__
+    def __init__(self, register_instance: bool = True):
+        if register_instance:
+            self._instance_refs[self.__class__].add(weakref.ref(self))
```
sfc-gh-pmardziel, line 1190: See stacked refs comment above.

src/core/trulens/core/utils/python.py
```diff
@@ -1223,85 +1144,55 @@ def delete_singleton_by_name(
...
-            del Singleton._instances[k]
-        else:
-            logger.warning(
-                "Instance %s not found among existing singletons.", self
-            )
+    @classmethod
+    def get_instances(cls):
+        """Get all instances of the class."""
+        for inst_ref in cls._instance_refs[cls]:
+            inst = inst_ref()
```
sfc-gh-pmardziel, line 1196: See stacked ref comment above.

sfc-gh-pmardziel (approved):
See comment regarding Refs inside WeakSet being not necessary. Please fix.

PR #1565, opened 2024-10-16
warning and docpage for bad context

sfc-gh-pmardziel:
# Description

Added a warning if the `contextvars` used for tracing and cost tracking cannot be retrieved. The warning points to a new doc page about causes and possible solutions. 

Small things related to investigating the warning:

- Fixed async methods in `DummyApp` and `DummyEndpoint` to not use `time.sleep` and use `asyncio.sleep` instead.
- Removed use of `python_utils.superstack` in `DummyApp` as we no longer rely on preserving stack across contexts.
- Updated and added more examples inside `dummy_example.ipynb` that show parallel invocations with threads and async tasks.
- Fixed some blocking issues with `App.wait_for_feedback_results`.
- Added a non-blocking option to the container used to implement the prior.
- Added a property `Record.feedback_results_as_completed` which wraps `Record.feedback_results` with `as_completed`.
- Added a warning to `python.py:task_factory_with_stack` if user does something that would cause the no context error / warning to be produced. Also made this method work with python prior to 3.11 .

## Type of change

  not work as expected)
- [ ] New Tests


src/feedback/trulens/feedback/dummy/endpoint.py
```diff
@@ -119,8 +120,80 @@ async def apost(
...
+        # allocate some data to pretend we are doing hard work
+        temporary = np.empty(self.alloc, dtype=np.int8)  # noqa: F841
+
+        if self.delay > 0.0:
+            await asyncio.sleep(
+                max(0.0, self.ndt.np_random.normal(self.delay, self.delay / 2))
+            )
+
+        r = self.ndt.discrete_choice(
+            seq=["normal", "freeze", "error", "loading", "overloaded"],
```
sfc-gh-dhuang, line 139: small nit - can we make these constants on the top?

src/feedback/trulens/feedback/dummy/endpoint.py
```diff
@@ -119,8 +120,80 @@ async def apost(
...
+                - self.loading_prob
+                - self.overloaded_prob,
+                self.freeze_prob,
+                self.error_prob,
+                self.loading_prob,
+                self.overloaded_prob,
+            ],
+        )
+
+        if r == "freeze":
```
sfc-gh-dhuang, line 153: same nit / suggestion as above: could use some constants here

sfc-gh-dhuang (approved):
left small nit comments - mostly LGTM

sfc-gh-jreini:
The new warning docs page does not seem super actionable. What does the user actually need to do in each of these cases? Can we give them code examples they can run or use for troubleshooting rather than api references?

src/feedback/trulens/feedback/dummy/endpoint.py
sfc-gh-pmardziel: fixed

src/feedback/trulens/feedback/dummy/endpoint.py
sfc-gh-pmardziel: fixed

sfc-gh-pmardziel:
> The new warning docs page does not seem super actionable. What does the user actually need to do in each of these cases? Can we give them code examples they can run or use for troubleshooting rather than api references?

There are code samples with "before" (what they might have been doing before) and "after" (what they need to do to fix the problem).

PR #1577, opened 2024-10-19
Add tags to schema during snowflake app creation

sfc-gh-pdharmana:
# Description

and context.

## Type of change

  not work as expected)
- [ ] New Tests

----

> [!IMPORTANT]
> Adds `trulens_workspace_version` tag to Snowflake schema using database revision in `connector.py`.
> 
>   - **Behavior**:
>     - Adds `trulens_workspace_version` tag to Snowflake schema in `_init_with_snowpark_session()` in `connector.py` using `get_db_revision()`.
>   - **Database**:
>     - Adds `get_db_revision()` method to `DB` class in `base.py`.
>     - Implements `get_db_revision()` in `SQLAlchemyDB` in `sqlalchemy.py` to return current database revision.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=truera%2Ftrulens&utm_source=github&utm_medium=referral)<sup> for a42857d83b13957fa7290b1430715636293a5df9. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/connectors/snowflake/trulens/connectors/snowflake/connector.py
+++ b/src/connectors/snowflake/trulens/connectors/snowflake/connector.py
@@ -13,7 +13,6 @@ from typing import (
 from trulens.connectors.snowflake.utils.server_side_evaluation_artifacts import (
     ServerSideEvaluationArtifacts,
 )
-from trulens.core import __version__ as trulens_version
 from trulens.core.database import base as core_db
 from trulens.core.database.base import DB
 from trulens.core.database.connector.base import DBConnector
@@ -191,7 +190,10 @@ class SnowflakeConnector(DBConnector):
         ).collect()
         res = snowpark_session.sql(
             "ALTER schema {}.{} SET TAG {}='{}'".format(
-                db, schema, TRULENS_WORKSPACE_VERSION_TAG, trulens_version
+                db,
+                schema,
+                TRULENS_WORKSPACE_VERSION_TAG,
+                self.db.get_db_revision(),
             )
         ).collect()
 
--- a/src/connectors/snowflake/trulens/connectors/snowflake/connector.py
+++ b/src/connectors/snowflake/trulens/connectors/snowflake/connector.py
@@ -13,15 +13,13 @@ from typing import (
 from trulens.connectors.snowflake.utils.server_side_evaluation_artifacts import (
     ServerSideEvaluationArtifacts,
 )
+from trulens.core import __version__ as trulens_version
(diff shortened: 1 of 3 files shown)


src/connectors/snowflake/trulens/connectors/snowflake/connector.py
```diff
@@ -179,6 +180,12 @@ def _init_with_snowpark_session(
                 snowpark_session, database_args["database_prefix"]
             ).set_up_all()
 
+        # Add "trulens_workspace_version" tag to the current schema
+        schema = snowpark_session.get_current_schema()
+        snowpark_session.sql(
+            f"ALTER SCHEMA {schema} SET TAG trulens_workspace_version = {trulens_version}"
```
sfc-gh-dkurokawa, line 186: don't you have to put single quotes around the tag value?

src/connectors/snowflake/trulens/connectors/snowflake/connector.py
```diff
@@ -180,6 +181,22 @@ def _init_with_snowpark_session(
                 init_server_side_with_staged_packages,
             ).set_up_all()
 
+        # Add "trulens_workspace_version" tag to the current schema
```
sfc-gh-dkurokawa, line 184: note that this will change the tag whenever you connect. Is that what you want? That is, consider these two cases:

1. If two ppl have two different trulens versions and are working together on the project the version will keep flip-flopping between the two whenever they connect.
2. I think if you fail the db revision check this will still execute even though it's not even at the right version.

sfc-gh-pdharmana: At high level, we don't care about the tag value, we care if tags exist to point to objects. If you think we can use another tag that is more constant than versions, we could use that too. We don;t have a workspace name like "app_name" to use that instead.

We have the following options:
1. version (with the current problems) you mentioned 
2. Ignore value - just use a constant value for the tag. 
3. Include metadata of tables under the schema like (number of apps) to help reduce querytime. Complex implementation.

In this PR i chose (1), but I think we could also go with (3).

sfc-gh-dkurokawa: can we have the alembic version?

sfc-gh-pdharmana: Done

sfc-gh-dkurokawa (approved):

PR #1581, opened 2024-10-21
Have `run_leaderboard` should fail more clearly if it's unable to authenticate at Snowflake due to being created by a `snowpark_session`.

sfc-gh-dkurokawa:
# Description
Have `run_leaderboard` should fail more clearly if it's unable to authenticate at Snowflake due to being created by a `snowpark_session`.

Elliott and Josh both ran into this issue and it wasn't clear what the problem was.

Most of the changes in `src/connectors/snowflake/trulens/connectors/snowflake/connector.py` are just moving things around to be cleaner.

## Type of change
  not work as expected)
- [ ] New Tests

----

> [!IMPORTANT]
> Refactor `SnowflakeConnector` to improve error clarity for missing authentication and add test for password requirement.
> 
>   - **Behavior**:
>     - `run_dashboard` in `run.py` now raises a `ValueError` if `SnowflakeConnector` is created without a password when using a `snowpark_session`.
>     - Adds `_is_snowflake_connector` function in `run.py` to check connector type.
>   - **Refactoring**:
>     - Moves session creation logic to `_create_snowpark_session` in `connector.py`.
>     - Moves session validation logic to `_validate_snowpark_session_with_connection_parameters` in `connector.py`.
>   - **Testing**:
>     - Adds `test_run_leaderboard_without_password` in `test_snowflake_connection.py` to verify error handling when password is missing.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=truera%2Ftrulens&utm_source=github&utm_medium=referral)<sup> for 0532dafbed04485537aa7b9e32a18096d8acd84f. It will automatically update as commits are pushed.</sup>


src/connectors/snowflake/trulens/connectors/snowflake/connector.py
```diff
@@ -118,9 +74,80 @@ def __init__(
...
+                connection_parameters_to_set.append(k)
+        if connection_parameters_to_set:
+            raise ValueError(
+                f"If not supplying `snowpark_session` then must set `{connection_parameters_to_set}`!"
+            )
+        self.password_known = True
+        # Create snowpark session making sure to create schema if it doesn't
+        # already exist.
+        schema = connection_parameters["schema"]
+        del connection_parameters["schema"]
```
sfc-gh-dhuang, line 97: just curious - this is done so that you don't have to explicitly catch/except the connection error when the passed in schema from connection params is not valid or not recognized right?

sfc-gh-dhuang (approved):
left one question

src/connectors/snowflake/trulens/connectors/snowflake/connector.py
sfc-gh-dkurokawa: yes; since we want to be able to handle the situation where we just create the schema on the fly.

PR #1584, opened 2024-10-22
Blocking guardrails

sfc-gh-jreini:
# Description

Blocking guardrails so that apps can block unsafe input from reaching the app, and unsafe output from reaching the user.

## Type of change

  not work as expected)
- [x] New Tests

----

> [!IMPORTANT]
> Add `block_input` and `block_output` decorators to block unsafe inputs and outputs, with documentation updates and tests.
> 
>   - **New Features**:
>     - Add `block_input` and `block_output` decorators in `base.py` to block unsafe inputs and outputs based on feedback functions and thresholds.
>     - `block_input` prevents function execution if input feedback exceeds threshold, returning `None`.
>     - `block_output` blocks output if feedback exceeds threshold, returning `None`.
>   - **Documentation**:
>     - Update `index.md` with examples and explanations for `block_input` and `block_output` usage.
>   - **Tests**:
>     - Add `test_guardrails.py` to test `block_input` and `block_output` decorators.
>   - **Misc**:
>     - Change `logger.warn` to `logger.warning` in `base.py` for consistency.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=truera%2Ftrulens&utm_source=github&utm_medium=referral)<sup> for d50dc64d954553871d5902cc8bc9aca9921bba87. It will automatically update as commits are pushed.</sup>


sfc-gh-pmardziel:
Can you add to these decorators the value to be returned if something is blocked instead of returning None in all cases?

src/core/trulens/core/guardrails/base.py
```diff
@@ -98,4 +101,168 @@ def wrapper(*args, **kwargs):
         # note: the following information is manually written to the wrapper because @functools.wraps(func) causes breaking of the method.
         wrapper.__name__ = func.__name__
         wrapper.__doc__ = func.__doc__
+        wrapper.__signature__ = sig
+        return wrapper
+
+
+class block_input:
```
sfc-gh-dhuang, line 108: class name should be capitalized not snake case

src/core/trulens/core/guardrails/base.py
```diff
@@ -98,4 +101,168 @@ def wrapper(*args, **kwargs):
...
+            return func(*args, **kwargs)
+
+        # note: the following information is manually written to the wrapper because @functools.wraps(func) causes breaking of the method.
+        wrapper.__name__ = func.__name__
+        wrapper.__doc__ = func.__doc__
+        wrapper.__signature__ = sig
+        return wrapper
+
+
+class block_output:
```
sfc-gh-dhuang, line 199: cap the class name

src/core/trulens/core/guardrails/base.py
```diff
@@ -98,4 +101,168 @@ def wrapper(*args, **kwargs):
...
+        self.threshold = threshold
+        self.return_value = return_value
+
+    def __call__(self, func):
+        sig = inspect.signature(func)
+
+        def wrapper(*args, **kwargs):
+            output = func(*args, **kwargs)
+            result = self.feedback(output)
+            if not isinstance(result, float):
```
sfc-gh-dhuang, line 253: will this work with feedback results like 
{0.5, 'reasons': "COT reasons str..."}?

sfc-gh-jreini: No, must be score-only feedback. That disclaimer is in the docs and there is a dedicated error message for that case. We could allow reasons feedback but that would be extra tokens generated that are not seen/used, so it seems better to not allow it

src/core/trulens/core/guardrails/base.py
sfc-gh-jreini: Agree in other cases but here this will be the decorator name, matching other existing trulens decorators like `@instrument` and `@context_filter`

sfc-gh-pmardziel: Note that these are functions in spirit though classes in syntax. They should not be used like classes, hence the naming pattern.

src/core/trulens/core/guardrails/base.py
sfc-gh-pmardziel: Could also generalize to allow feedbacks with selectors. Perhaps add a jira issue for future work?

src/core/trulens/core/guardrails/base.py
```diff
@@ -98,4 +101,168 @@ def wrapper(*args, **kwargs):
...
+            if self.keyword_for_prompt not in sig.parameters:
+                raise TypeError(
+                    f"Keyword argument '{self.keyword_for_prompt}' not found in `{func.__name__}` signature."
+                )
+        else:
+            # For backwards compatibility, allow inference of keyword_for_prompt:
+            first_arg = list(k for k in sig.parameters.keys() if k != "self")[0]
+            self.keyword_for_prompt = first_arg
+            logger.warning(
+                f"Assuming `{self.keyword_for_prompt}` is the `{func.__name__}` arg to filter. "
```
sfc-gh-pmardziel, line 173: Update warning to this use case.

sfc-gh-pmardziel (approved):

src/core/trulens/core/guardrails/base.py
sfc-gh-dhuang: makes sense and thanks for the context

src/core/trulens/core/guardrails/base.py
sfc-gh-jreini: I think this does allow feedbacks with selectors, but I'll create a JIRA nonetheless

sfc-gh-jreini: Would be good also to allow feedbacks that take multiple args, such as a way to block ungrounded responses

PR #1589, opened 2024-10-25
Always ensure endpoint context variable is cleaned up.

sfc-gh-dkurokawa:
# Description
Always ensure endpoint context variable is cleaned up.

Previously, if we failed during app invocation the endpoint context variable remained set. This was a problem because there is different behaviors for when it's set (as to allow cost tracking). This ensures this doesn't happen.

There's two changes here:
1. I just put the context variable reset in a `finally` block.
2. A test to make sure this doesn't happen again.

## Type of change

  not work as expected)
- [ ] New Tests

----

> [!IMPORTANT]
> Ensure endpoint context variable is always reset in `_track_costs` and add a test to verify cleanup.
> 
>   - **Behavior**:
>     - Ensure endpoint context variable is reset in `_track_costs` in `endpoint.py` using a `finally` block.
>     - Prevents leftover context variables when app invocation fails.
>   - **Tests**:
>     - Add `test_context_variables.py` to verify context variable cleanup.
>     - Includes a test case `test_endpoint_contextvar_always_cleaned` that simulates a failure and checks cleanup.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=truera%2Ftrulens&utm_source=github&utm_medium=referral)<sup> for 3a4c02e3820b357189dc9524a6df162991a19229. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/core/trulens/core/feedback/endpoint.py
+++ b/src/core/trulens/core/feedback/endpoint.py
@@ -603,29 +603,30 @@ class Endpoint(
         # following call to retrieve.
         endpoints_token = Endpoint._context_endpoints.set(endpoints)  # noqa: F841
 
-        # context_vars = contextvars.copy_context()
-        context_vars = {
-            Endpoint._context_endpoints: Endpoint._context_endpoints.get()
-        }
+        try:
+            # context_vars = contextvars.copy_context()
+            context_vars = {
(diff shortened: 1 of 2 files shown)


sfc-gh-dhuang (approved):
👍 

Just curious if there's any documentation / references for the cause of the `snowpark_session.sql` failure

PR #1600, opened 2024-10-28
Don't check for user and account in snowpark sessions because Streamlit apps might hide them.

sfc-gh-dkurokawa:
# Description
Don't check for user and account in snowpark sessions because Streamlit apps might hide them.

Some users are running the snowflake connector in Streamlit and running into an issue where the user isn't set (in reality it is set, it's just not visible).

## Type of change
  not work as expected)
- [ ] New Tests

----

> [!IMPORTANT]
> Skip checking 'account' and 'user' in Snowpark sessions for Streamlit apps in `connector.py`.
> 
>   - **Behavior**:
>     - In `connector.py`, `_validate_snowpark_session_with_connection_parameters` now skips checking `account` and `user` if they are not visible, addressing issues in Streamlit apps.
>     - Ensures Snowpark sessions can proceed without visible `account` and `user` values, assuming they are set but hidden.
>   - **Misc**:
>     - Adds comments explaining the rationale for skipping these checks in Streamlit environments.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=truera%2Ftrulens&utm_source=github&utm_medium=referral)<sup> for c31e623467399b3700605f3f9203fb96535b08f6. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/connectors/snowflake/trulens/connectors/snowflake/connector.py
+++ b/src/connectors/snowflake/trulens/connectors/snowflake/connector.py
@@ -119,6 +119,12 @@ class SnowflakeConnector(DBConnector):
         missing_snowpark_session_parameters = []
         mismatched_parameters = []
         for k, v in snowpark_session_connection_parameters.items():
+            if k in ["account", "user"] and not v:
+                # Streamlit apps may hide these values so we don't check them.
+                # They are required for a Snowpark Session anyway so this isn't
+                # a problem (though we can't check consistency with
+                # `connection_parameters`).
+                continue
             if not v:
                 missing_snowpark_session_parameters.append(k)
             elif connection_parameters[k] not in [None, v]:


src/connectors/snowflake/trulens/connectors/snowflake/connector.py
```diff
@@ -119,6 +119,12 @@ def _validate_snowpark_session_with_connection_parameters(
         missing_snowpark_session_parameters = []
         mismatched_parameters = []
         for k, v in snowpark_session_connection_parameters.items():
+            if k in ["account", "user"] and not v:
+                # Streamlit apps may hide these values so we don't check them.
+                # They are required for a Snowpark Session anyway so this isn't
```
sfc-gh-chu, line 124: wouldn't this be "They aren't required"?

sfc-gh-chu (approved):

src/connectors/snowflake/trulens/connectors/snowflake/connector.py
sfc-gh-dkurokawa: no they are required.

Basically, AFAICT, you can't create a snowpark session without a user and account. But, in Streamlit apps, when we ask for them, they hide it from us unless you have a specific setting enabled in your Snowflake account.

PR #1627, opened 2024-11-02
Ensure `make clean` does indeed clean or it will fail out.

sfc-gh-dkurokawa:
# Description
Ensure `make clean` does indeed clean or it will fail out.

Also removed an extra dependency to `openai` in a test that wasn't necessary.

During a release, I noticed you could get away without cleaning things which we probably shouldn't allow. Also there was some issue with the test environment (hence my removing of the `openai` thing).

## Type of change
  not work as expected)
- [ ] New Tests

----

> [!IMPORTANT]
> Ensure `make clean` in `Makefile` performs a clean operation and remove unnecessary `openai` import from `test_context_variables.py`.
> 
>   - **Makefile**:
>     - Modify `clean` target to exit with a message if the user chooses not to clean.
>   - **Tests**:
>     - Remove unnecessary `openai` import from `test_context_variables.py`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=truera%2Ftrulens&utm_source=github&utm_medium=referral)<sup> for 34a8cb5077ba943df57e890d190d7e3d41a63f47. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/tests/e2e/test_context_variables.py
+++ b/tests/e2e/test_context_variables.py
@@ -5,7 +5,6 @@ Tests for context variable issues.
 import os
 import unittest
 
-import openai
 from snowflake.snowpark import Session
 from trulens.apps.custom import TruCustomApp
 from trulens.apps.custom import instrument
@@ -37,8 +36,6 @@ class TestContextVariables(unittest.TestCase):
     @optional_test
     def test_endpoint_contextvar_always_cleaned(self):
         class FailingRAG:
-            oai_client = openai.OpenAI()
-
             @instrument
             def retrieve(self, query: str) -> list:
                 return ["A", "B", "C"]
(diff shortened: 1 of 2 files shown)


tests/e2e/test_context_variables.py
```diff
@@ -5,7 +5,6 @@
 import os
 import unittest
 
-import openai
```
sfc-gh-chu, line 8: are these changes intended for this PR?

sfc-gh-chu (approved):

tests/e2e/test_context_variables.py
sfc-gh-chu: ah nvm read the pr description

PR #1652, opened 2024-11-15
Remove `poetry.lock` files.

sfc-gh-dkurokawa:
# Description
Remove `poetry.lock` files. I removed them for a few reasons:
1. I keep seeing them in PRs.
2. It makes it slightly less likely that if something changes with our (transitive) dependencies that break us we'll notice since we're all on the same versions with the `poetry.lock` files.
3. We're forced into updating them a lot.
4. It may not even work for a different OS AFAIU.
5. That's what's recommended for libraries as per my Googling.

That said, it does seem like it'll make the pipeline slower now since it has to resolve the dependencies (it seems to take 6-7 min longer in my sample size of 1). That's a trade-off I'm willing to accept but not sure of the reviewers.

TBH, the main issue I have with them is that I keep seeing them in so many PRs and it's annoying because of how many there are. AFAICT here are the ways we can go:
1. We do nothing.
2. We do what this PR suggests and remove them.
3. We only keep the `poetry.lock` in the root directory. I don't know why we have the other ones, it's not like they're used AFAIU? @coreyhu correct me if I'm wrong.
4. We remove the pre-commits for them which is what I'm assuming is causing ppl to update them all the time.

## Type of change
  not work as expected)
- [ ] New Tests

----

> [!IMPORTANT]
> Remove `poetry.lock` files and related configurations to improve dependency management for library development.
> 
>   - **Dependency Management**:
>     - Removed `poetry.lock` files to avoid version locking across different environments.
>     - Added `pypi-public` source to `pyproject.toml` for dependency resolution.
>   - **Pre-commit Configuration**:
>     - Removed `poetry-check` hooks from `.pre-commit-config.yaml`.
>   - **Makefile Adjustments**:
>     - Modified `env-tests-required` target to use `&&` for command chaining.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=truera%2Ftrulens&utm_source=github&utm_medium=referral)<sup> for 5f86486be4be1649871ddcf095fd40abb8f0ca51. It will automatically update as commits are pushed.</sup>


sfc-gh-chu:
I'm in favor of option 3 (keep only the root lockfile and add the others to gitignore), which requires some of option 4 for the removed lockfiles. The sub-lockfiles are only needed when building the package, but our dev environment only depends on the root one. The sub-lockfiles also much faster to create from scratch than the root one (30s vs 2-3 min).

sfc-gh-dkurokawa:
Closed in favor of https://github.com/truera/trulens/pull/1653/files

PR #1670, opened 2024-11-26
Fix `poetry` environment issues.

sfc-gh-dkurokawa:
# Description
It seems that the root `pyproject.toml` doesn't take into account the python version requirements of the dependent `pyproject.toml` files during a `poetry lock`. Specifically, in this case, there are restrictions on `scipy` depending on the python version but it's not taken into account unless we put it in the root `pyproject.toml`. This is causing an issue in another PR so I spun this out.

## Type of change
  not work as expected)
- [ ] New Tests

----

> [!IMPORTANT]
> Add `scipy` to root `pyproject.toml` to fix `poetry lock` Python version handling issue.
> 
>   - **Dependencies**:
>     - Add `scipy` dependency to root `pyproject.toml` with version constraints based on Python version to address `poetry lock` issue.
>     - Update `scipy` version constraint in `src/feedback/pyproject.toml` for Python >=3.9 to `>=1.11.1`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=truera%2Ftrulens&utm_source=github&utm_medium=referral)<sup> for 5fef00cefc49d9478cb9b388251558be64ab4b66. It will automatically update as commits are pushed.</sup>


pyproject.toml
```diff
@@ -50,6 +50,13 @@ mypy = "^1.11"
 build = "^1.2"
 twine = "^5.1"
 codespell = "^2.3"
+# scipy is here even though it's for trulens-feedback because `poetry lock`
+# doesn't incorporate the python version specific dependencies in
+# `./src/feedback/pyproject.toml` correctly for some reason.
+scipy = [
```
sfc-gh-chu, line 56: is it just scipy causing this issue right now? I added similar python-specific versioning for `tiktoken`, `onnxruntime`, `numpy`, and `torch` for different packages

sfc-gh-dkurokawa: yeah for some reason `scipy` is the only one causing an issue for me in another PR

1. `tiktoken` luckily has `0.7.0` satisfying all the requirements.
2. `onnxruntime` has `1.19.2` satisfying python 3.12 and below and 3.13 we don't actually test anything on so it's fine.
3. `numpy` does seem like it could be a problem as the version `poetry.lock` has chosen (i.e. `1.26.4` isn't supported for python 3.8 but the pipeline shows that it somehow installed `1.24.4` do to another dependency.
4. `torch` has `2.5.1`. satisfying everything.

sfc-gh-chu (approved):

PR #1674, opened 2024-12-03
Feedback customization (including few-shot examples)

sfc-gh-jreini:
# Description

- Add an option to pass user-defined examples to feedback functions.
- Allow output space, temperature and groundedness parameters to be used in Feedback class instance (not just directly)

## Type of change

  not work as expected)
- [ ] New Tests

----

> [!IMPORTANT]
> Add user-defined examples to feedback functions, fix prompt template bug, and enhance parameter usage in feedback classes.
> 
>   - **New Feature**:
>     - Add `examples` parameter to `Feedback` class in `feedback.py` and `FeedbackDefinition` in `schema/feedback.py` for user-defined examples.
>     - Introduce `FewShotExamples` class in `v2/feedback.py` to handle few-shot examples.
>   - **Bug Fix**:
>     - Fix incorrect filling of feedback function prompt templates for output space in `llm_provider.py`.
>   - **Enhancements**:
>     - Allow `output_space` and `temperature` parameters in `Feedback` class in `feedback.py`.
>     - Add `_determine_output_space_prompt` method in `llm_provider.py` to resolve output space prompts.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=truera%2Ftrulens&utm_source=github&utm_medium=referral)<sup> for 2822c4e1f926ba63124eb744db9765c554d2a6c0. It will automatically update as commits are pushed.</sup>


src/feedback/trulens/feedback/llm_provider.py
```diff
@@ -309,11 +309,24 @@ def _determine_output_space(
             f"Invalid score range: [{min_score_val}, {max_score_val}]. Must match one of the predefined output spaces."
         )
 
+    def _determine_output_space_prompt(self, output_space_name: str) -> str:
```
sfc-gh-dhuang, line 312: A bit confused here - does the existing output space validation logic using pydantic validator not work? 

We have a mixin dedicated to this and this is called within `generate_system_prompt`
https://github.com/truera/trulens/blob/main/src/feedback/trulens/feedback/v2/feedback.py#L246

sfc-gh-dhuang: Ah okay I saw you removed the logic in `CriteriaOutputSpaceMixin` below. could you elaborate a bit why doing so - my main concern is that now we are repeating `output_space_prompt = self._determine_output_space_prompt(output_space)` in almost every feedback function, and we are also forcing the dev to do the same in the future.

src/feedback/trulens/feedback/v2/feedback.py
```diff
@@ -475,6 +509,12 @@ class PromptResponseRelevance(Relevance, WithPrompt, CriteriaOutputSpaceMixin):
         """
     )
 
+    system_prompt: ClassVar[str] = cleandoc(
```
sfc-gh-dhuang, line 512: I guess the same question as above - we are already doing in `def relevance(...)` in llm_provider.py`   
```
system_prompt = (
            feedback_v2.PromptResponseRelevance.generate_system_prompt(
                min_score_val, max_score_val, criteria, output_space
            )
        )
```, do this not work anymore?

Comment (sfc-gh-jreini) [reply src/feedback/trulens/feedback/v2/feedback.py:512]: This was not working afaict. I ran some test prints of the prompts and the template was not filled in properly, for example the output space was left as "LIKERT_0_3" instead of the prompt itself and the min/max scores were not filled into the criteria.

File: src/feedback/trulens/feedback/llm_provider.py (hunk 1 above)
Comment (sfc-gh-jreini) [reply src/feedback/trulens/feedback/llm_provider.py:312]: Not sure if the validation works, but the output space prompt generation was not working. I moved it so I could call it directly and pass the output space prompt to the system message generator

File: src/feedback/trulens/feedback/v2/feedback.py (hunk 2 above)
Comment (sfc-gh-dhuang) [reply src/feedback/trulens/feedback/v2/feedback.py:512]: hmm this is odd. 

The code has been there for several months and RAG Triad has been using this logic without issue, at least with the default output space and criteria.


I tested below with a minimal example:
```

gpt_4o_mini = OpenAI(model_engine="gpt-4o-mini")

gpt_4o_mini.context_relevance_with_cot_reasons(question='test question', 
                              context='test context',
                              criteria="Daniel's test criteria",
                              min_score_val=0,
                              max_score_val=1)
```

could you share your code that fails? happy to talk synchronously over a call / huddle too. 
and the printed system prompt also shows the logic works properly (I inserted a print within `context_relevance_with_cot_reasons` method) 
```
system_prompt in context relevance w/ cot: You are a RELEVANCE grader; providing the relevance of the given CONTEXT to the given QUESTION.
Respond only as a number from 0 or 1, where 0 is lowest and negative (i.e. irrelevant or not grounded) and 1 is highest and positive (relevant, grounded, valid, etc.).

Criteria for evaluating relevance:
Daniel's test criteria

A few additional scoring guidelines:

- Long CONTEXTS should score equally well as short CONTEXTS.

- RELEVANCE score should increase as the CONTEXTS provides more RELEVANT context to the QUESTION.

- RELEVANCE score should increase as the CONTEXTS provides RELEVANT context to more parts of the QUESTION.

- Never elaborate.
```

File: src/core/trulens/core/feedback/feedback.py [hunk 3]
```diff
@@ -152,10 +152,36 @@ class Feedback(feedback_schema.FeedbackDefinition):
...
+    min_score_val: Optional[int] = pydantic.Field(None, exclude=True)
+    """Minimum score value for the feedback function."""
+
+    max_score_val: Optional[int] = pydantic.Field(None, exclude=True)
+    """Maximum score value for the feedback function."""
+
+    temperature: Optional[float] = pydantic.Field(None, exclude=True)
+    """Temperature parameter for the feedback function."""
+
+    use_sent_tokenize: Optional[bool] = pydantic.Field(None, exclude=True)
```
sfc-gh-dhuang, line 167: for `use_sent_tokenize` and `filter_trivial_statements`, wdyt if we can encapsulate them in a separate (pydantic) class called sth like GroundednessConfigs ?  this way we have composition and don't have to include these 2 (and potentially more) groundedness-specific attributes.

src/feedback/trulens/feedback/v2/feedback.py
```diff
@@ -178,6 +181,60 @@ class OutputSpace(Enum):
...
+             0)
+        ]
+
+        fewshot_relevance_examples = FewShotExamples.from_list(fewshot_relevance_examples_list)
+        ```
+    """
+
+    @classmethod
+    def from_list(
+        cls, examples_list: List[Tuple[dict, int]]
```
sfc-gh-dhuang, line 211: consider adding a new Pydantic class FewShotExamples to have more control over validating Tuple[dict, int] / feedback scores

sfc-gh-dhuang (commented):
2 small-ish comments and not meant to be blocking

src/feedback/trulens/feedback/v2/feedback.py
sfc-gh-jreini: put this in - thanks for the feedback!

sfc-gh-dhuang (approved):
LGTM - ty

PR #1692, opened 2024-12-11
Introduce Event table for ORM to prepare for OTEL traces

sfc-gh-gtokernliang:
# Description

- Update `docs/contributing/migration.md` to reflect current migration steps. Specifically - the database with new changes can only be done after the alembic revision script is run.
- Introduce `Event` in `orm.py` to reflect an Event (or span)

N/A

## Type of change

  not work as expected)
- [ ] New Tests

----

> [!IMPORTANT]
> Add `Event` table and class for OTEL traces, update migration steps, and define `EventID` type alias.
> 
>   - **Database Schema**:
>     - Add `Event` class in `orm.py` for OTEL traces.
>     - Create `10_create_event_table.py` migration script to add `events` table.
>     - Update `sql_alchemy_migration_versions` in `data.py` to include new version.
>   - **Documentation**:
>     - Update `migration.md` to reflect new migration steps and order.
>   - **Schema**:
>     - Add `Event` class in `event.py` to represent event data.
>     - Define `EventID` type alias in `types.py`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=truera%2Ftrulens&utm_source=github&utm_medium=referral)<sup> for 1149f46c0c3a720460a9c003d60933647e5b1a2b. It will automatically update as commits are pushed.</sup>


src/core/trulens/core/database/migrations/versions/10_create_event_table.py
```diff
@@ -0,0 +1,60 @@
+"""create event table
```
sfc-gh-gtokernliang, line 1: @sfc-gh-dkurokawa is there a way to make this optional for users? Though I don't foresee changes, this is still an incomplete feature so there might be unexpected changes to the db schema

src/core/trulens/core/database/migrations/versions/10_create_event_table.py
```diff
@@ -0,0 +1,60 @@
...
+depends_on = None
+
+
+def upgrade(config) -> None:
+    prefix = config.get_main_option("trulens.table_prefix")
+
+    if prefix is None:
+        raise RuntimeError("trulens.table_prefix is not set")
+
+    # TODO: The automatically generated code below likely references
```
sfc-gh-dkurokawa, line 24: can you do this for L32 and L59?

src/core/trulens/core/database/migrations/versions/10_create_event_table.py
```diff
@@ -0,0 +1,60 @@
...
+    #       tables such as "trulens_feedback_defs" or "trulens_records".
+    #       However, the common prefix for these tables "trulens_" is
+    #       actually configurable and so replace it with the variable
+    #       prefix.
+    #       e.g. replace "trulens_records" with prefix + "records".
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
```
sfc-gh-dkurokawa, line 31: I think it's possible to check the sqlalchemy dialect here and if so, in the case of Snowflake, can you just make an actual event table with the clustering? That way we're closer to what we want in testing.

src/core/trulens/core/database/migrations/versions/10_create_event_table.py
sfc-gh-dkurokawa: I think we can just have an experimental flag that sets the revision to upgrade to when we call `db_migrations.upgrade_db`. That is, if we're doing this OTEL stuff we can do this event table, otherwise we can do the regular thing for the time being.

src/core/trulens/core/database/orm.py
```diff
@@ -401,6 +404,86 @@ def parse(
                     dataset_json=obj.model_dump_json(redact_keys=redact_keys),
                 )
 
+        class Event(base):
```
sfc-gh-dkurokawa, line 407: I didn't compare this to the event table column defn, but it seems fine otherwise.

src/core/trulens/core/schema/event.py
```diff
@@ -0,0 +1,70 @@
...
+import logging
+from typing import Any, Dict, Hashable
+
+from trulens.core.schema import types as types_schema
+from trulens.core.utils import serial as serial_utils
+
+logger = logging.getLogger(__name__)
+
+
+class Event(serial_utils.SerialModel, Hashable):
```
sfc-gh-dkurokawa, line 15: same comment above here.

sfc-gh-dkurokawa (commented):
Aside from the import issues, LGTM.

sfc-gh-dkurokawa (approved):

PR #1703, opened 2024-12-19
Change recording to create a span of type "record_root"

sfc-gh-gtokernliang:
# Description

Updated the app context manager to be able to create a span of type "record_root"

The general approach is the following:
1. In `TruApp`'s context manager - we create a span of type `RECORD_ROOT`. This is necessary because:
  a. we're relying on the context manager to do things such as running/deferring the feedback function on record completion
  b. The user might already be actively tracing, so we need a point for which we identify the calls that belong to the record
2. We introduce a new span type - `MAIN`, intended for the main entrypoint into the app's calls. 
  a. This is because we do not have control over what happens within the recording context, so there might be scenarios where the user can call non-instrumented functions, that we have no way of tracking the input/outputs of. 

```
RECORD_ROOT (This holds the record)
|_ MAIN (this should be the main entrypoint into the app's calls)
   |_ call 1
   |_ call 2
   |_ ...
```

## Type of change

  not work as expected)
- [x] New Tests

----

> [!IMPORTANT]
> Add `RECORD_ROOT` and `MAIN` span types to `TruApp` context manager for improved recording and tracing.
> 
>   - **Behavior**:
>     - Add `RECORD_ROOT` span type in `TruApp` context manager in `app.py` to manage recording contexts.
>     - Introduce `MAIN` span type in `instrument.py` for main entry points.
>   - **Instrumentation**:
>     - Modify `instrument()` in `instrument.py` to handle `MAIN` span type.
>     - Update `set_main_span_attributes()` in `span.py` to set attributes for `MAIN` spans.
>   - **Testing**:
>     - Update `test_otel_instrument.py` to test new span types and attributes.
>     - Add new test cases for `MAIN` and `RECORD_ROOT` spans.
>   - **Misc**:
>     - Add `tokens` and `span_context` attributes to `App` class in `app.py` for context management.
>     - Import changes in `app.py` to use `OTELApp` from `instrument.py`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=truera%2Ftrulens&utm_source=github&utm_medium=referral)<sup> for f3c9e22f134d429f39adfadaff95ae7ad59da480. It will automatically update as commits are pushed.</sup>


src/core/trulens/experimental/otel_tracing/core/instrument.py
```diff
@@ -81,3 +92,57 @@ def wrapper(*args, **kwargs):
...
+        logging.debug("Entering the OTEL app context.")
+
+        # Note: This is not the same as the record_id in the core app since the OTEL
+        # tracing is currently separate from the old records behavior
+        otel_record_id = str(uuid.uuid4())
+
+        tracer = trace.get_tracer_provider().get_tracer(TRULENS_SERVICE_NAME)
+
+        # Calling set_baggage does not actually add the baggage to the current context, but returns a new one
+        # To avoid issues with remembering to add/remove the baggage, we attach it to the runtime context.
```
sfc-gh-dkurokawa, line 109: s/To/to

src/core/trulens/experimental/otel_tracing/core/instrument.py
```diff
@@ -81,3 +92,57 @@ def wrapper(*args, **kwargs):
...
+        root_span.set_attribute(
+            SpanAttributes.RECORD_ROOT.APP_NAME, self.app_name
+        )
+        root_span.set_attribute(
+            SpanAttributes.RECORD_ROOT.APP_VERSION, self.app_version
+        )
+        root_span.set_attribute(SpanAttributes.RECORD_ROOT.APP_ID, self.app_id)
+        root_span.set_attribute(
+            SpanAttributes.RECORD_ROOT.RECORD_ID, otel_record_id
+        )
```
sfc-gh-dkurokawa, line 134: the semantic convention that Piotr laid out also has the `MAIN_INPUT`/`MAIN_OUTPUT`/`MAIN_ERROR`, but I don't see how we can get that even during `__exit__`, so I'm inclined to say we can remove them. Though that raises the question of, how will the UI know what the input and output are to display. It's too close to the finish line of 2024 that I don't want to investigate how this is determined currently but can next year haha.

There's also `TOTAL_COST` but I'm not as worried about that.

src/core/trulens/experimental/otel_tracing/core/instrument.py
```diff
@@ -51,6 +58,10 @@ def wrapper(*args, **kwargs):
                 span.set_attribute(
                     "parent_span_id", parent_span.get_span_context().span_id
                 )
+                span.set_attribute(
```
sfc-gh-dkurokawa, line 61: we want app/run id in the baggage as well I guess, but given that it's not totally clear yet no point in doing it now I suppose.

src/core/trulens/experimental/otel_tracing/core/instrument.py
sfc-gh-dkurokawa: I also don't really get this token business, can you enlighten me?

sfc-gh-gtokernliang: from what I can tell - to set the baggage you need to:

1. call `set_baggage`, which returns a new, updated context
2. attach it to the current context with `context_api.attach` - this returns a token that can then be used to remove it.
3. Once the record is over, we want to remove it, and the way to do it via OTEL `context` is by telling it the baggage you want it to detach via `context.detach(token)`

src/core/trulens/experimental/otel_tracing/core/instrument.py
sfc-gh-gtokernliang: yeah we should chat about this - I think one idea I had while typing this up was to omit the record_root span type, and instead just store the relevant attributes (app name/version/id/record id) in the baggage. record root can just be identified as the span with no parent

src/core/trulens/experimental/otel_tracing/core/instrument.py
sfc-gh-gtokernliang: I'll put the app id in baggage for now I guess since that seems pretty good to me. For run_id it's a little more ambiguous what the shape of the API will look like so I'll leave that out for now.

src/core/trulens/experimental/otel_tracing/core/instrument.py
sfc-gh-dkurokawa: We still need the root span type though to denote when we start a record though since if you create a otel trace before you start calling your app a bunch of times they'll all be part of the same trace but we want them to be separate records.

But more importantly, I'm confused, how does this help us determine the `MAIN_INPUT`/`MAIN_OUTPUT`/`MAIN_ERROR`?

src/core/trulens/experimental/otel_tracing/core/instrument.py
```diff
@@ -137,3 +157,69 @@ def wrapper(*args, **kwargs):
...
+
+        return root_span
+
+    def __exit__(self, exc_type, exc_value, exc_tb):
+        remove_baggage(SpanAttributes.RECORD_ID)
+        remove_baggage(SpanAttributes.APP_ID)
+
+        logger.debug("Exiting the OTEL app context.")
+
+        while len(self.tokens) > 0:
```
sfc-gh-dkurokawa, line 219: nit: `while self.tokens:`

In python, if you put something in the while loop it will do a `bool` on it and a list returns `False` iff it's empty.

src/core/trulens/experimental/otel_tracing/core/instrument.py
```diff
@@ -137,3 +157,69 @@ def wrapper(*args, **kwargs):
...
+            # Clearing the context once we're done with this root span.
+            # See https://github.com/open-telemetry/opentelemetry-python/issues/2432#issuecomment-1593458684
+            context_api.detach(self.tokens.pop())
+
+        if self.span_context:
```
sfc-gh-dkurokawa, line 224: would this ever not be true?

src/core/trulens/core/app.py
```diff
@@ -421,6 +422,18 @@ def tru(self) -> core_connector.DBConnector:
         pydantic.PrivateAttr(default_factory=dict)
     )
 
+    tokens: list[object] = []
```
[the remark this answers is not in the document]
sfc-gh-gtokernliang: thanks! didn't know about the difference between the two before. 

`object` is chosen because that's the signature of the `attach` function in the OTEL context API

Screenshot 2024-12-24 at 10 03 17 AM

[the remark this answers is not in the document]
sfc-gh-gtokernliang: Screenshot 2024-12-24 at 10 03 59 AM

src/core/trulens/experimental/otel_tracing/core/instrument.py
```diff
@@ -84,6 +90,20 @@ def wrapper(*args, **kwargs):
                     # It's on the user to deal with None as a return value.
                     func_exception = e
 
+                span.set_attribute("name", func.__name__)
```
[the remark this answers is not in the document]
sfc-gh-gtokernliang: yeah I'm trying to follow the schema where possible - see https://docs.snowflake.com/en/developer-guide/logging-tracing/event-table-columns#for-span-record-type

src/core/trulens/experimental/otel_tracing/core/instrument.py
sfc-gh-gtokernliang: oh nice - JS does that too, wasn't sure if it worked here, good to know!

src/core/trulens/experimental/otel_tracing/core/instrument.py
sfc-gh-gtokernliang: I doubt it, but didn't think it would hurt to include it, if nothing, at least for the type checking haha

examples/experimental/otel_exporter.ipynb
```diff
@@ -53,9 +52,14 @@
    "outputs": [],
    "source": [
     "from trulens.experimental.otel_tracing.core.instrument import instrument\n",
+    "from trulens.otel.semconv.trace import SpanAttributes\n",
     "\n",
     "\n",
     "class TestApp:\n",
+    "    @instrument(span_type=SpanAttributes.SpanType.MAIN)\n",
```
sfc-gh-dkurokawa, line 59: out of curiosity, why didn't you just put this decorator on `respond_to_query`?

src/core/trulens/experimental/otel_tracing/core/instrument.py
```diff
@@ -1,16 +1,104 @@
...
+def validate_value_for_attribute(value):
+    """
+    Ensure that value is a valid attribute value type, and coerce it to a string if it is not.
+
+    This is helpful for lists/etc because if any single value is not a valid attribute value type, the entire list
+    will not be added as a span attribute.
+    """
+    arg_type = type(value)
+
+    # Coerge the argument to a string if it is not a valid attribute value type.
```
sfc-gh-dkurokawa, line 36: coerce

src/core/trulens/experimental/otel_tracing/core/instrument.py
```diff
@@ -1,16 +1,104 @@
...
+from opentelemetry.baggage import set_baggage
+import opentelemetry.context as context_api
+from trulens.core import app as core_app
 from trulens.experimental.otel_tracing.core.init import TRULENS_SERVICE_NAME
 from trulens.otel.semconv.trace import SpanAttributes
 
 logger = logging.getLogger(__name__)
 
 
+VALID_ATTR_VALUE_TYPES = (bool, str, int, float)
```
sfc-gh-dkurokawa, line 18: technically I think `int` isn't a signed 64 bit integer and it's like a class that can handle arbitrarily large ints so we may want to check if it can fit into a signed 64 bit int.

src/core/trulens/experimental/otel_tracing/core/instrument.py
```diff
@@ -1,16 +1,104 @@
...
+        return str(value)
+
+    return value
+
+
+def validate_list_of_values_for_attribute(arguments: list):
+    """
+    Ensure that all values in a list are valid attribute value types, and coerce them to strings if they are not.
+    """
+    return list(map(validate_value_for_attribute, arguments))
```
sfc-gh-dkurokawa, line 47: The OTEL spec that you copied (and I confirmed is in the link you have) says this:
"The array MUST be homogeneous, i.e., it MUST NOT contain values of different types."
This will allow something like [1, True, "2"] when it should be converted to ["1", "True", "2"]. I think you even have a test checking this incorrectly.

src/core/trulens/experimental/otel_tracing/core/instrument.py
```diff
@@ -1,16 +1,104 @@
...
+    if (
+        SpanAttributes.SELECTOR_NAME_KEY in result
+        and SpanAttributes.SELECTOR_NAME in result
+    ):
+        raise ValueError(
+            f"Both {SpanAttributes.SELECTOR_NAME_KEY} and {SpanAttributes.SELECTOR_NAME} cannot be set."
+        )
+
+    if SpanAttributes.SELECTOR_NAME in result:
+        # Transfer the trulens namespaced to the non-trulens namespaced key.
```
sfc-gh-dkurokawa, line 71: Transfer the trulens namespace

src/core/trulens/experimental/otel_tracing/core/instrument.py
```diff
@@ -77,13 +127,45 @@ def wrapper(*args, **kwargs):
...
+
                 try:
                     ret = func(*args, **kwargs)
                 except Exception as e:
                     # We want to get into the next clause to allow the users to still add attributes.
                     # It's on the user to deal with None as a return value.
                     func_exception = e
 
+                if span_type == SpanAttributes.SpanType.MAIN:
+                    all_args = list(args) + list(kwargs.values())
```
sfc-gh-dkurokawa, line 150: I thought about this for like 15 minutes just now because I don't love this for a few reasons, but I honestly can't think of a good solution to the things that bother me. My main issues are:
1. If there are kwargs we won't know what keyword they map to.
2. It's quite likely the first arg is self which I doubt anyone cares to see the string representation of.
3. I'm not sure if we should just error out if theres multiple inputs (e.g. if `all_args` is `[self, "hi", "bye"]).

Anyway if you can think of a nicer way I'm all ears but I'm not blocking this approach since it's not like I can see a good approach to replace it with.

src/core/trulens/experimental/otel_tracing/core/instrument.py
```diff
@@ -127,13 +209,77 @@ def wrapper(*args, **kwargs):
...
+
+        logger.debug("Exiting the OTEL app context.")
+
+        while self.tokens:
+            # Clearing the context once we're done with this root span.
+            # See https://github.com/open-telemetry/opentelemetry-python/issues/2432#issuecomment-1593458684
+            context_api.detach(self.tokens.pop())
+
+        if self.span_context:
+            self.span_context.__exit__(exc_type, exc_value, exc_tb)
```
sfc-gh-dkurokawa, line 285: can you put a TODO about adding in the feature function spans. I forgot to mention I appended it to this task:
https://snowflakecomputing.atlassian.net/browse/SNOW-1854360

tests/unit/test_otel_instrument.py
```diff
@@ -134,10 +145,105 @@ def test_instrument_decorator(self) -> None:
...
+        with self.subTest("dict"):
+            self.assertEqual(
+                validate_value_for_attribute({"key": "value"}),
+                "{'key': 'value'}",
+            )
+
+    def test_validate_list_of_values_for_attribute(self):
+        with self.subTest("list of primitives"):
+            self.assertEqual(
+                validate_list_of_values_for_attribute([1, "2", True]),
```
sfc-gh-dkurokawa, line 232: this is the test I was talking about above.

src/core/trulens/experimental/otel_tracing/core/instrument.py
sfc-gh-gtokernliang: haha - I checked the opentelemetry python package when implementing this (https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-api/src/opentelemetry/attributes/__init__.py#L31), and they don't verify the signed 64 bit int part of that too

src/core/trulens/experimental/otel_tracing/core/instrument.py
sfc-gh-gtokernliang: yeah this is my bad, thanks for catching it!

src/core/trulens/experimental/otel_tracing/core/span.py
```diff
@@ -138,3 +144,26 @@ def get_main_input(func: Callable, args: tuple, kwargs: dict) -> str:
     sig = signature(func)
     bindings = signature(func).bind(*args, **kwargs)
     return signature_utils.main_input(func, sig, bindings)
+
+
+def set_main_span_attributes(
+    span: Span,
+    /,
```
sfc-gh-dkurokawa, line 151: is this a typo?

src/core/trulens/experimental/otel_tracing/core/span.py
```diff
@@ -138,3 +144,26 @@ def get_main_input(func: Callable, args: tuple, kwargs: dict) -> str:
...
+    exception: Optional[Exception],
+) -> None:
+    span.set_attribute(
+        SpanAttributes.MAIN.MAIN_INPUT, get_main_input(func, args, kwargs)
+    )
+
+    if exception:
+        span.set_attribute(SpanAttributes.MAIN.MAIN_ERROR, str(exception))
+
+    elif ret is not None:
```
sfc-gh-dkurokawa, line 165: I realize that right now that `elif` is the same as `if`, but can you change this to `if` in case there's a future we can have both?

src/core/trulens/experimental/otel_tracing/core/span.py
```diff
@@ -138,3 +144,26 @@ def get_main_input(func: Callable, args: tuple, kwargs: dict) -> str:
...
     return signature_utils.main_input(func, sig, bindings)
+
+
+def set_main_span_attributes(
+    span: Span,
+    /,
+    func: Callable,
+    args: tuple,
+    kwargs: dict,
+    ret,
```
sfc-gh-dkurokawa, line 155: `Any`?

src/core/trulens/experimental/otel_tracing/core/span.py
sfc-gh-gtokernliang: I wanted the remaining args to be positional and kwargs for code readability, happy to change this

sfc-gh-dkurokawa (approved):

PR #1708, opened 2025-01-04
Enable exporting spans to snowflake stage if a `TruLensSnowflakeSpanExporter` is provided

sfc-gh-gtokernliang:
# Description

Enable exporting spans to a snowflake stage if `SnowflakeConnector` is provided. 

N/A

## Type of change

  not work as expected)
- [ ] New Tests

----

> [!IMPORTANT]
> Add `TruLensSnowflakeSpanExporter` to export spans to Snowflake stage and refactor OTEL tracing in `TruSession`.
> 
>   - **New Feature**:
>     - Introduced `TruLensSnowflakeSpanExporter` in `exporter/snowflake.py` to export spans to Snowflake stage.
>     - Added `_export_to_snowflake_stage` method for exporting spans as protobuf files.
>   - **Refactoring**:
>     - Removed `span_context` and `tokens` from `App` class in `app.py`.
>     - Replaced OTEL context management with `_prevent_invalid_otel_syntax` in `app.py`.
>     - Moved `convert_to_any_value` to `exporter/utils.py`.
>   - **Session Management**:
>     - Updated `TruSession` to handle `SpanExporter` and `TracerProvider` in `session.py`.
>     - Added `experimental_force_flush` method to `TruSession` for flushing spans.
>   - **Testing**:
>     - Updated tests in `test_exporter.py` and `test_otel_instrument.py` to reflect new exporter functionality.
>     - Regenerated golden test results in `api.trulens.3.11.yaml` and `api.trulens_eval.3.11.yaml`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=truera%2Ftrulens&utm_source=github&utm_medium=referral)<sup> for 4e27038e87a57349c0409fe4fe4e58650713fae9. It will automatically update as commits are pushed.</sup>


src/core/trulens/experimental/otel_tracing/core/exporter.py
```diff
@@ -59,12 +61,75 @@ def _construct_event(self, span: ReadableSpan) -> event_schema.Event:
...
+            try:
+                with tempfile.NamedTemporaryFile(
+                    delete=False, suffix=".csv", mode="w", newline=""
+                ) as tmp_file:
+                    tmp_file_path = tmp_file.name
+                    logger.debug(
+                        f"Writing spans to the csv file: {tmp_file_path}"
+                    )
+                    writer = csv.writer(tmp_file)
+                    writer.writerow(["span"])
```
sfc-gh-yuzhao, line 87: re: our discussion, let's take out the header

tests/unit/static/golden/test_otel_instrument__test_instrument_decorator.csv
```diff
@@ -1,11 +1,11 @@
...
+0,"{'name': 'root', 'kind': 'SPAN_KIND_TRULENS', 'parent_span_id': '', 'status': 'STATUS_CODE_UNSET'}",9121270002874598350,"{'name': 'root', 'kind': 'SPAN_KIND_TRULENS', 'trulens.span_type': 'record_root', 'trulens.app_id': 'app_hash_baf7b2cb6402e84fa3b0b3a028d4bf65', 'trulens.record_id': '45423cb3-114c-4dae-b759-ba1627fac008', 'trulens.record_root.app_name': 'default_app', 'trulens.record_root.app_version': 'base', 'trulens.record_root.app_id': 'app_hash_baf7b2cb6402e84fa3b0b3a028d4bf65', 'trulens.record_root.record_id': '45423cb3-114c-4dae-b759-ba1627fac008'}",EventRecordType.SPAN,"{'telemetry.sdk.language': 'python', 'telemetry.sdk.name': 'opentelemetry', 'telemetry.sdk.version': '1.29.0', 'service.name': 'trulens'}",2025-01-06 10:40:46.456557,2025-01-06 10:40:46.460513,"{'trace_id': '77915888720279316253686111587837941505', 'parent_id': '', 'span_id': '9121270002874598350'}"
```
sfc-gh-dkurokawa, line 2: haha I feel like I ask this every PR, but why did this change?

src/core/trulens/experimental/otel_tracing/core/exporter.py
```diff
@@ -59,12 +61,75 @@ def _construct_event(self, span: ReadableSpan) -> event_schema.Event:
         )
 
     def export(self, spans: Sequence[ReadableSpan]) -> SpanExportResult:
+        def check_if_trulens_span(span: ReadableSpan) -> bool:
+            if not span.attributes:
+                return False
+
+            return span.attributes.get("kind") == "SPAN_KIND_TRULENS"
+
+        trulens_spans = list(filter(check_if_trulens_span, spans))
```
sfc-gh-dkurokawa, line 70: why are we only writing the trulens spans?

src/core/trulens/experimental/otel_tracing/core/exporter.py
```diff
@@ -59,12 +61,75 @@ def _construct_event(self, span: ReadableSpan) -> event_schema.Event:
...
+
+        if not trulens_spans:
+            return SpanExportResult.SUCCESS
+
+        if isinstance(self.connector, SnowflakeConnector):
```
sfc-gh-dkurokawa, line 75: can you put all this code in a function called like "export_to_snowflake_stage" or something.

sfc-gh-dkurokawa: Also maybe as part of the initialization of the exporter we can choose to also allow you to do the sqlalchemy with Snowflake combination (though by defaultl this shouldn't be true).

src/core/trulens/experimental/otel_tracing/core/exporter.py
```diff
@@ -59,12 +61,75 @@ def _construct_event(self, span: ReadableSpan) -> event_schema.Event:
...
+                logger.error(f"Error writing spans to the csv file: {e}")
+                return SpanExportResult.FAILURE
+
+            try:
+                logger.debug("Uploading file to Snowflake stage")
+                snowpark_session = self.connector.snowpark_session
+
+                logger.debug("Creating Snowflake stage if it does not exist")
+                snowpark_session.sql(
+                    "CREATE STAGE IF NOT EXISTS trulens_spans"
```
sfc-gh-pdharmana, line 103: Should this include `TEMP` so its deleted at end of session?

sfc-gh-pdharmana: https://docs.snowflake.com/en/sql-reference/sql/create-stage#optional-parameters

src/core/trulens/experimental/otel_tracing/core/exporter.py
sfc-gh-gtokernliang: the process of uploading generates more spans from the snowflake `connect` function, resulting in an infinite loop if we don't filter out the generated spans 😅  happy to discuss what the exact logic should look like though

src/core/trulens/experimental/otel_tracing/core/exporter.py
sfc-gh-gtokernliang: @sfc-gh-yuzhao  can you confirm what's preferable? I saw a discussion where we were considering keeping it around for retries

src/core/trulens/experimental/otel_tracing/core/exporter.py
sfc-gh-gtokernliang: let's chat in the office - not quite sure I understand

sfc-gh-dkurokawa: Talked offline, but for posterity, I meant 1. Sqlalchemy with Snowflake is basically the non-event table sqlalchemy way.

src/core/trulens/experimental/otel_tracing/core/exporter.py
sfc-gh-dkurokawa: I think we should upload all the spans by default and we can have an option for this kinda targeted upload. I feel like when you set the exporter up you're implicitly agreeing that all spans are going to be exported anyway.

tests/unit/static/golden/test_otel_instrument__test_instrument_decorator.csv
sfc-gh-gtokernliang: I ended up updating the SPAN_KIND because it had a deeper meaning than what I thought it was haha

src/core/trulens/experimental/otel_tracing/core/exporter.py
sfc-gh-gtokernliang: setting it as temp for now

src/core/trulens/core/app.py
```diff
@@ -972,6 +970,22 @@ async def __aexit__(self, exc_type, exc_value, exc_tb):
 
         return
 
+    def __call__(self, *, run_name: str, input_id: str):
+        if not self.session.experimental_feature(
```
sfc-gh-dkurokawa, line 974: sorry can you explain to me what's going on in this function?

src/core/trulens/core/session.py
```diff
@@ -177,15 +181,38 @@ def experimental_otel_exporter(
...
+        Force flush the OpenTelemetry exporters.
+
+        Args:
+            timeout_millis: The maximum amount of time to wait for spans to be
+                processed.
+
+        Returns:
+            False if the timeout is exceeded, feature is not enabled, or the provider doesn't exist, True otherwise.
+        """
+        timeout_millis = timeout_millis or 300000
```
sfc-gh-dkurokawa, line 204: is there a reason you didn't just say `timeout_millis: int = 300000` in the signature? The only difference is that this will convert 0 to 300000 AFAIK.

src/core/trulens/experimental/otel_tracing/core/exporter.py
```diff
@@ -153,7 +155,96 @@ def construct_event(span: ReadableSpan) -> event_schema.Event:
...
+        except Exception as e:
+            logger.error(f"Error writing spans to the protobuf file: {e}")
+            return SpanExportResult.FAILURE
+
+        try:
+            logger.debug("Uploading file to Snowflake stage")
+
+            logger.debug("Creating Snowflake stage if it does not exist")
+            snowpark_session.sql(
+                "CREATE TEMP STAGE IF NOT EXISTS trulens_spans"
```
sfc-gh-dkurokawa, line 220: what happens if there's a non-temp stage with this name?

src/core/trulens/experimental/otel_tracing/core/session.py
```diff
@@ -1,37 +1,60 @@
...
-                "OpenTelemetry SDK not found. Please install OpenTelemetry SDK to use OpenTelemetry tracing."
-            )
+logger = logging.getLogger(__name__)
 
-        assert isinstance(
-            val, otel_export_sdk.SpanExporter
-        ), "otel_exporter must be an OpenTelemetry SpanExporter."
 
+class _TruSession(core_session.TruSession):
+    def _setup_otel_exporter(
```
sfc-gh-dkurokawa, line 23: now that I think about it, what happens if they're bringing their own tracer provider and/or exporter?

src/core/trulens/experimental/otel_tracing/core/exporter.py
sfc-gh-gtokernliang: it just uploads to that stage - this situation can happen if e.g. there are two exports happening at the same time, etc.

src/core/trulens/experimental/otel_tracing/core/session.py
sfc-gh-gtokernliang: so I've updated the API slightly - you can see `otel_exporter.ipynb`, where the user creates a `TruLensSnowflakeSpanExporter` and passes it in here to export spans to a trulens stage - the advantage here is that they'd be able to store their other tables in a different database, while being able to submit the span logs to the event table. 

For the tracer provider wise I think the only thing that it'll impact is the resource - see line 40 where we're initializing our tracer provider:
```python
        provider = TracerProvider(resource=resource)
        trace.set_tracer_provider(provider)
```
I'll update the code to use `trace.get_tracer_provider` for the remainder, so it won't matter that the user has set a different one globally.

If they bring their own exporter it shouldn't be a problem, because the trace processors attached to a tracer provider should all receive the same traces.

src/core/trulens/core/session.py
sfc-gh-gtokernliang: pylance was showing some weird error for me and recommended this. Updated it per your recommendation

src/core/trulens/core/app.py
sfc-gh-gtokernliang: haha - this might not have been the best way to achieve this. 

The goal is to enable syntax that looks like:
```python
    with custom_app({run_id: 'run_id', input_id='input_id'}) as recording:
        test_app.respond_to_query(input)
```

to achieve this - `custom_app({run_id: 'run_id', input_id='input_id'})` should return an object that is a recording context (in our case a class with the `__enter__` and `__exit__` methods), and that the setting of the `run_name` and `input_id` shouldn't taint the object itself.

hence - the approach here is, when `custom_app` is called, it:
1. Checks that OTEL tracing is enabled
2. creates a clone of itself using `model_construct` + `model_dump` that is classed to `OTELRecordingContext` (see https://github.com/truera/trulens/pull/1708/files/5a555840da38b280a0b3c3cbabb40cbd8a3e93c4#diff-25f7f4954d77bc71539bcd8d1e7b3133552596b468bf53a6962b636b176de682R96)
3. then performs the normal recording context duties

src/core/trulens/core/app.py
```diff
@@ -889,20 +876,18 @@ def model_dump(self, *args, redact_keys: bool = False, **kwargs):
             **kwargs,
         )
 
-    # For use as a context manager.
-    def __enter__(self):
-        if not core_instruments.Instrument._have_context():
-            raise RuntimeError(core_endpoint._NO_CONTEXT_WARNING)
-
+    def _prevent_invalid_otel_syntax(self):
```
sfc-gh-dkurokawa, line 879: why did you name this function this? Like what do you mean by otel tracing syntax?

src/core/trulens/experimental/otel_tracing/core/instrument.py
```diff
@@ -93,27 +93,53 @@ def wrapper(*args, **kwargs):
...
-        # To avoid issues with remembering to add/remove the baggage, we attach it to the runtime context.
-        self.tokens.append(
-            context_api.attach(
-                set_baggage(SpanAttributes.RECORD_ID, otel_record_id)
-            )
-        )
-        self.tokens.append(
-            context_api.attach(set_baggage(SpanAttributes.APP_ID, self.app_id))
-        )
+        self.attach_to_context(SpanAttributes.RECORD_ID, otel_record_id)
```
sfc-gh-dkurokawa, line 133: what happens if it's already in the baggage?

src/core/trulens/experimental/otel_tracing/core/instrument.py
```diff
@@ -130,7 +138,10 @@ def __enter__(self):
 
         tracer = trace.get_tracer_provider().get_tracer(TRULENS_SERVICE_NAME)
 
-        self.attach_to_context(SpanAttributes.RECORD_ID, otel_record_id)
+        # There might be nested records - in those scenarios use the outer one.
+        if not get_baggage(SpanAttributes.RECORD_ID):
+            self.attach_to_context(SpanAttributes.RECORD_ID, otel_record_id)
+
         self.attach_to_context(SpanAttributes.APP_NAME, self.app.app_name)
```
sfc-gh-dkurokawa, line 145: Can you do the same as RECORD_ID For the other attributes? Since I guess the APP_NAME and other stuff aren't thread safe since if you call the child LLM app then continue working the baggage will get changed on you.

sfc-gh-gtokernliang: changed `attach_to_context` to check for existence + ensure that value is non null

sfc-gh-dkurokawa (approved):

PR #1765, opened 2025-01-30
[DO NOT MERGE] Allow `@instrument` decorator to work without parentheses.

sfc-gh-dkurokawa:
# Description
Allow @instrument decorator to work without parentheses.

This can be a bit dangerous because the only way to do this, is to say if there's only a single positional argument to the decorator that's a callable we can assume the decorator was used without parentheses --- but this is a problem if we happen to use the decorator wrong and supply a single positional argument that's a function like a lambda for the `attributes`.

## Type of change
  not work as expected)
- [x] New Tests

----

> [!IMPORTANT]
> Allow `@instrument` decorator to be used without arguments, adding `instrument_arg_parser` for argument handling and updating tests.
> 
>   - **Behavior**:
>     - Allow `@instrument` decorator to be used without arguments, making `@instrument()` equivalent to `@instrument`.
>     - Introduces `instrument_arg_parser` in `instrument.py` to handle argument parsing for `instrument`.
>     - Raises `ValueError` if `@instrument` is called with positional arguments.
>   - **Tests**:
>     - Adds new test case `nested4` in `test_otel_tru_custom.py` to verify `@instrument` without arguments.
>     - Updates `test_otel_tru_custom__test_smoke.csv` with new expected results.
>   - **Misc**:
>     - Updates docstring for `instrument` to reflect new usage.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=truera%2Ftrulens&utm_source=github&utm_medium=referral)<sup> for be8dbe35ce33a93d66692b2fc0406841a1c4b0bf. It will automatically update as commits are pushed.</sup>


sfc-gh-gtokernliang (approved):

sfc-gh-dkurokawa:
After discussing with @sfc-gh-gtokernliang on whether this change is worth it given the potential caveats, we've decided to hold off for the time being but we'll have it here in waiting for the time being.

sfc-gh-jreini:
Closing: stale draft, superseded by current architecture.

PR #1771, opened 2025-01-31
make main_method required for app base class for OTEL

sfc-gh-dhuang:
# Description

and context.

## Type of change

  not work as expected)
- [ ] New Tests

----

> [!IMPORTANT]
> Make `main_method` required for `TruApp` when OTEL tracing is enabled, updating initialization and tests accordingly.
> 
>   - **Behavior**:
>     - `main_method` is now required for `TruApp` when OTEL tracing is enabled, ensuring proper method binding and serialization.
>     - Raises `ValueError` if `main_method` is not provided or improperly bound.
>   - **Code Changes**:
>     - Removed `main_method_loaded` and `main_method` attributes from `TruApp`.
>     - Added `main_method_name` attribute for serialization in `app.py`.
>     - Updated `main_call` in `app.py` to use `main_method_name`.
>   - **Tests**:
>     - Updated tests in `test_otel_guardrail.py`, `test_otel_main_method.py`, and `test_otel_multithreaded.py` to include `main_method`.
>     - Added `test_otel_main_method.py` to check for `ValueError` when `main_method` is missing.
>     - Modified other test files to ensure `main_method` is specified.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=truera%2Ftrulens&utm_source=github&utm_medium=referral)<sup> for 9341561b4098a1157fc8b57aaba355d9b81365c4. It will automatically update as commits are pushed.</sup>


src/core/trulens/core/app.py
```diff
@@ -433,6 +433,15 @@ def __init__(
         feedbacks: Optional[Iterable[core_feedback.Feedback]] = None,
         **kwargs,
     ):
+        otel_tracing_enabled = TruSession().experimental_feature(
```
sfc-gh-dhuang, line 436: @sfc-gh-dkurokawa is this how we we would check if OTEL feature is enabled? 

Also curious why we are initializing a new TruSession() instead of just use the self.session in `_wrap_main_function` below

src/core/trulens/apps/app.py
```diff
@@ -376,19 +375,24 @@ def __init__(self, app: Any, methods_to_instrument=None, **kwargs: Any):
...
+                            "and could not be found in the given `app` instance."
+                        )
+
                 main_name = main_method.__name__
                 main_method_loaded = main_method
                 main_method = pyschema_utils.Function.of_function(
                     main_method_loaded
                 )
 
-                if not python_utils.safe_hasattr(
```
sfc-gh-dhuang, line 385: I might not miss some previous context on this - but given not many users / none of them that we know of have been specifying `main_method` explicitly in the past, I am thinking moving forward we bind the method for users who passing in `TestApp.respond_to_query` to make their life easier. thoughts? @sfc-gh-dkurokawa

sfc-gh-dkurokawa: We need to have the main method logic in the actual `core_app.App` superclass because we want `TruLlama` and `TruChain` to have it too.

Also, we want to make sure it's only for OTEL.

Also, this legacy `main_method` stuff for `TruCustomApp` (now renamed `TruApp`) seems to be able to take in an actual callable. I'm not sure we should allow this since if it's not part of `app` then it's kinda weird.

I think we should do a few things:
1. Remove this `main_method` stuff from `TruCustomApp` because I doubt it's being used. My understanding is that it's legacy anyway.
2. We remove any reference to this old `main_method` in our docs (it appears it's in a few experimental notebooks).
3. We do the `main_method` stuff in the `core_app.App` but only for OTEL.

sfc-gh-dkurokawa (commented):
Also, can you add a unit test class like `test_otel_main_method` that just tries to create apps without the main method specified and asserts that there was an error?

src/core/trulens/apps/app.py
sfc-gh-dhuang: for the old references of `main_method` in some of the notebooks, they are all bound methods on the user's `app` so perhaps that's fine?

src/core/trulens/core/app.py
```diff
@@ -445,10 +445,76 @@ def __init__(
...
+                )
+            if app is None:
+                raise ValueError(
+                    "A valid app instance must be provided when specifying 'main_method'."
+                )
+
+            main_method = kwargs["main_method"]
+
+            if isinstance(main_method, dict):
+                main_method = pyschema_utils.Function.model_validate(
```
sfc-gh-dkurokawa, line 466: what does it mean for it to be a dictionary?

Also if this is supposed to be here, it doesn't define `main_name` anyway.

src/core/trulens/core/app.py
```diff
@@ -445,10 +445,76 @@ def __init__(
...
+                main_name = main_method.name
+
+                cls = main_method.cls.load()
+                mod = main_method.module.load().__name__
+
+            else:
+                # Handle both bound and unbound methods
+                if not hasattr(main_method, "__self__"):
+                    if hasattr(app, main_method.__name__):
+                        main_method = getattr(
```
sfc-gh-dkurokawa, line 481: I realize you didn't write this stuff, and just moved it here, but this doesn't make sense to me. Like if you pass in some function called `invoke` or something, it would try to bind it to my `app`? What if it wasn't the `app.invoke` that I was using and something else?

tests/unit/test_otel_main_method.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+class TestOtelMainMethod(OtelAppTestCase):
+    def test_missing_main_method_raises_error(self):
+        tru_session = TruSession()
+        tru_session.reset_database()
+        dummy_app = DummyApp()
+        # Attempt to create a TruApp without specifying main_method.
+        with self.assertRaises(ValueError) as context:
+            _ = TruApp(
```
sfc-gh-dkurokawa, line 22: nit: you don't need the `_ = `

tests/unit/test_otel_main_method.py
```diff
@@ -0,0 +1,32 @@
...
+# Dummy application to use with TruApp.
+class DummyApp:
+    def respond_to_query(self, query: str) -> str:
+        return "dummy response"
+
+
+class TestOtelMainMethod(OtelAppTestCase):
+    def test_missing_main_method_raises_error(self):
+        tru_session = TruSession()
+        tru_session.reset_database()
```
sfc-gh-dkurokawa, line 18: almost certainly you did this because of me and you saw code where I did this, but it's not necessary anymore as it's part of the `OtelAppTestCase::setUp` function.

src/core/trulens/core/app.py
```diff
@@ -461,6 +527,21 @@ def __init__(
...
+            # Set main_method to the unbound version. Will be passing in app for
+            # "self" manually when needed.
+            main_method_loaded = getattr(cls, main_name)
+
+            # This will be serialized as part of this TruApp. Importantly, it is unbound.
+            main_method = pyschema_utils.Function.of_function(
+                main_method_loaded, cls=cls
+            )
+
+            self.main_method = main_method
```
sfc-gh-dkurokawa, line 542: why do we even need to serialize this stuff? If we're serializing the `app` won't that be part of it already? We could just store the `main_name` (though I prefer we call it `main_method_name` or just `main_method`).

src/core/trulens/core/app.py
sfc-gh-dhuang: oh yeah good call out. should require the passed in main method to always be bound to the user's app

src/core/trulens/core/app.py
sfc-gh-dhuang: removed legacy code

src/core/trulens/core/app.py
sfc-gh-dhuang: re: 'why do we even need to serialize this stuff' my guess walking into this PR is that we previously try to set the unbound version of the main_method for easier serialization.
but yes agree to simplify the main_name / main_method_name should suffice given that now we require it to always be bound to `app`

src/core/trulens/core/app.py
```diff
@@ -445,10 +448,52 @@ def __init__(
...
+                or main_method.__self__ != app
+            ):
+                raise ValueError(
+                    f"main_method `{main_method.__name__}` must be bound to the provided `app` instance."
+                )
+
+            cls = app.__class__
+            mod = cls.__module__
+
+            kwargs["main_method"] = main_method
```
sfc-gh-dkurokawa, line 481: I don't think this line is doing anything given L468.

sfc-gh-dkurokawa (approved):

PR #1847, opened 2025-02-27
Check before trying to remove quotes if they enclose db and schema.

sfc-gh-dkurokawa:
# Description
Check before trying to remove quotes if they enclose db and schema.

## Type of change
  not work as expected)
- [ ] New Tests


sfc-gh-gtokernliang (approved):
Approved, but would prefer that we get an authority on snowpark to comment about the scenarios in which there'll be double quotes vs not. Been fixing some UI bugs around these so a little worried for you haha

Note that `"` is a valid snowflake identifier, but it's such an edge case

PR #1853, opened 2025-02-28
Fix double quotes issue in run operation system function / method

sfc-gh-dhuang:
# Description

we need to ensure the double-quoted `object_name` is passed into as-is to Run operation, given they are created with double quotes like below

`CREATE EXTERNAL AGENT IDENTIFIER('"{escaped_name}"') WITH VERSION "{version}`

This fix should allow users to pass in the `object_name` as-is without converting to upper cases explicitly on the client side.

## Type of change

  not work as expected)
- [ ] New Tests

----

> [!IMPORTANT]
> Fix handling of double-quoted identifiers in Snowflake operations by introducing `double_quote_identifier()` and updating relevant methods and tests.
> 
>   - **Behavior**:
>     - Use `double_quote_identifier()` in `create_new_run`, `get_run`, `list_all_runs`, `_update_run`, and `delete_run` in `run.py` to ensure `object_name` is passed as-is.
>     - Introduce `double_quote_identifier()` in `sql_utils.py` to double quote identifiers for SQL.
>   - **Tests**:
>     - Update `test_create_new_run` and `test_delete_run` in `test_snowflake_run_dao.py` to use `double_quote_identifier()`.
>     - Add `test_double_quote_identifier` in `test_snowflake_external_agent_dao.py` and `test_snowflake_run_dao.py` to verify quoting behavior.
>     - Modify `test_list_runs_after_adding` in `test_snowflake_agents_and_runs.py` to check for existing runs without raising exceptions.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=truera%2Ftrulens&utm_source=github&utm_medium=referral)<sup> for c06060b1613a6e49ec37313c80a06617fc066bf3. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/connectors/snowflake/trulens/connectors/snowflake/dao/run.py
+++ b/src/connectors/snowflake/trulens/connectors/snowflake/dao/run.py
@@ -6,6 +6,7 @@ import pandas as pd
 from snowflake.snowpark import Session
 from snowflake.snowpark.row import Row
 from trulens.connectors.snowflake.dao.enums import SourceType
+from trulens.connectors.snowflake.dao.sql_utils import double_quote_identifier
 from trulens.connectors.snowflake.dao.sql_utils import execute_query
 from trulens.core.run import SUPPORTED_ENTRY_TYPES
 from trulens.core.run import SupportedEntryType
@@ -79,7 +80,7 @@ class RunDao:
         """
         # Build the request payload dictionary.
         req_payload = {
-            "object_name": object_name,
+            "object_name": double_quote_identifier(object_name),
             "object_type": object_type,
             "run_name": run_name,
             "description": description,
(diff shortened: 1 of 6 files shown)


src/connectors/snowflake/trulens/connectors/snowflake/dao/run.py
```diff
@@ -79,7 +80,7 @@ def create_new_run(
         """
         # Build the request payload dictionary.
         req_payload = {
-            "object_name": object_name,
+            "object_name": double_quote_identifier(object_name),
```
sfc-gh-ssardana, line 83: we expect user to give the name in exact way they used to create?

sfc-gh-dhuang: yes

PR #1885, opened 2025-03-25
SDK: Fix logic for ingestion E2E timeout

sfc-gh-dhuang:
# Description

Fix and update the logic so that the timeout is accounted for the e2e latency from the time where app invocation is done to the records are ingested to the event table. 

relevant slack discussion: https://snowflake.slack.com/archives/C07A50XP2DQ/p1742507027097329 

## Type of change

  not work as expected)
- [x] New Tests

----

> [!IMPORTANT]
> Fixes ingestion timeout logic by using the latest record timestamp for telemetry latency in `Run` class and adds corresponding tests.
> 
>   - **Behavior**:
>     - Adjusts timeout logic in `Run` class to use `read_latest_record_root_timestamp_in_ms()` for determining telemetry latency.
>     - Changes timeout constant from `INVOCATION_TIMEOUT_IN_MS` to `EXPECTED_TELEMETRY_LATENCY_IN_MS` in `run.py`.
>   - **Functions**:
>     - Adds `read_latest_record_root_timestamp_in_ms()` in `run.py` to fetch the latest record root timestamp from the event table.
>     - Refactors `clean_up_snowflake_identifier()` to be used across `run.py`, `sql_utils.py`, and `otel_exporter.py`.
>   - **Tests**:
>     - Adds tests for `read_latest_record_root_timestamp_in_ms()` in `test_snowflake_run_dao.py` to handle valid, empty, and NaN cases.
>     - Updates golden test results in `api.trulens.3.11.yaml`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=truera%2Ftrulens&utm_source=github&utm_medium=referral)<sup> for 194950f498f8755d2a2dbe86040b15c8b39f0301. It will automatically update as commits are pushed.</sup>


src/core/trulens/core/run.py
```diff
@@ -748,17 +735,31 @@ def start(self, input_df: Optional[pd.DataFrame] = None):
...
             logger.exception(
                 f"Error encountered during invoking app main method: {e}."
             )
 
+            current_time = self._get_current_time_in_ms()
             self.run_dao.upsert_run_metadata_fields(
                 entry_type=SupportedEntryType.INVOCATIONS.value,
                 entry_id=invocation_metadata_id,
-                start_time_ms=start_time_ms,
+                start_time_ms=current_time,
```
sfc-gh-rroy, line 760: Why are we rewriting this?

sfc-gh-dhuang: they are required in every request by DPO

src/connectors/snowflake/trulens/connectors/snowflake/dao/run.py
```diff
@@ -565,6 +565,61 @@ def read_spans_count_from_event_table(
...
+            FROM
+                table(snowflake.local.GET_AI_OBSERVABILITY_EVENTS(
+                    ?,
+                    ?,
+                    ?,
+                    'EXTERNAL AGENT'
+                ))
+            WHERE
+                RECORD_ATTRIBUTES:"snow.ai.observability.run.name" = ? AND
+                RECORD_ATTRIBUTES:"ai.observability.span_type" = 'record_root' LIMIT 1;
```
sfc-gh-dkurokawa, line 593: can you use the variables from `trace.py` here? That is, like:
```
RECORD_ATTRIBUTES:"{SpanAttributes.SPAN_TYPE}" = '{SpanAttributes.SpanType.RECORD_ROOT}'
```
or w/e (I haven't checked this code I wrote here at all).

sfc-gh-dkurokawa: Also I don't think the `LIMIT 1` is necessary AFAICT.

src/connectors/snowflake/trulens/connectors/snowflake/dao/run.py
```diff
@@ -565,6 +565,61 @@ def read_spans_count_from_event_table(
...
+        """
+
+        try:
+            result_df = self.session.sql(
+                query,
+                params=[
+                    self.session.get_current_database()[1:-1],
```
sfc-gh-dkurokawa, line 600: can you use the `_clean_up_snowflake_identifier` function to do this? That way it's clearer why you're doing it to other ppl. Might want to make it not private in that case or even just move it to somewhere else though.

src/connectors/snowflake/trulens/connectors/snowflake/dao/run.py
sfc-gh-dhuang: right I had the LIMIT 1 there before I used the MAX aggregator. 

removed and also now using the ones from trace.py

src/connectors/snowflake/trulens/connectors/snowflake/dao/run.py
sfc-gh-dhuang: done

sfc-gh-dkurokawa (approved):

PR #1905, opened 2025-04-10
Set up algorithm for feedback selectors 2.0.

sfc-gh-dkurokawa:
# Description
Set up algorithm for feedback selectors 2.0. Basically this means:
1. Introduce the `ai.observability.span_groups` span attribute which can be a string or list of strings. By default, we consider a span to be part of only the `None` span group.
2. Introduce the `Selector` class which is a data class with the `span_attribute`, `span_name`, and `span_type`. The latter two of which are optional. 
3. Whenever we want to map an input of a feedback function, you have to specify a `Selector` object to that input.
4. Whenever we run a feedback function we only consider inputs that all have the same span group.
5. If all the inputs to the feedback function have selectors that describe the same spans (i.e. the `span_name` and `span_type` of the `Selector` is the same), then we run the feedback function on all spans that satisfy the those `span_name` and `span_type` individually.
6. If an attribute is missing, we silently don't run the feedback function on that span.
7. If the inputs to the feedback function have selectors that describe different spans, then we first group the inputs by if they describe the same spans. For example, if the feedback function takes in 3 arguments `x1`, `x2`, `x3` and the `Selector` for `x1` and `x2` describe the same spans (i.e. the `span_name` and `span_type` is the same), and `x3` is different, then we group `x1` and `x2` together, and `x3` is in its own group. Then we look at the number of spans that satisfy each group's `span_name`+`span_type`. If two or more groups have multiple spans that satisfy `span_name`+`span_type` we throw an error as it's unclear how to do combine them. Otherwise, we take the Cartesian product of the spans from each group and use that to determine the inputs.
8. Note that the case outlined in point 6 generalizes point 4.

## Type of change
  not work as expected)
- [x] New Tests

----

> [!IMPORTANT]
> Introduces feedback selectors 2.0 with span group handling and new `Selector` class, enhancing feedback computation logic and adding tests.
> 
>   - **Behavior**:
>     - Introduces `ai.observability.span_groups` span attribute in `trace.py`.
>     - Adds `Selector` class in `computer.py` with `span_attribute`, `span_name`, and `span_type`.
>     - Modifies feedback computation to consider inputs with the same span group in `computer.py`.
>     - Handles cases where inputs describe the same or different spans, throwing errors for ambiguous cases.
>   - **Functions**:
>     - Adds `compute_feedback_by_span_group` in `computer.py` to compute feedback based on span groups.
>     - Implements `_group_kwargs_by_selectors`, `_collect_inputs_from_events`, `_map_record_id_to_record_roots`, `_validate_unflattened_inputs`, `_flatten_inputs`, `_run_feedback_on_inputs`, and `_call_feedback_function` in `computer.py`.
>   - **Tests**:
>     - Adds tests in `test_otel_feedback_computation.py` for `compute_feedback_by_span_group` and `_group_kwargs_by_selectors`.
>     - Tests cover cases with multiple invocations, span groups, and error handling for ambiguous inputs.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=truera%2Ftrulens&utm_source=github&utm_medium=referral)<sup> for 77d04fd048069a70525f6c4614751b41f86facc5. It will automatically update as commits are pushed.</sup>


sfc-gh-dkurokawa:
> @sfc-gh-dkurokawa please break down this PR

@sfc-gh-gtokernliang testing

sfc-gh-gtokernliang:
@sfc-gh-dkurokawa when will the E2E tests be passing?

src/feedback/trulens/feedback/computer.py
```diff
@@ -73,6 +79,334 @@ def _compute_feedback(
...
+    ), kwarg_group_to_inputs in unflattened_inputs.items():
+        # Create a list of dictionaries to use in the product.
+        input_dicts = list(kwarg_group_to_inputs.values())
+        if not input_dicts:
+            continue
+        # For each combination, merge the dictionaries.
+        for combination in itertools.product(*input_dicts):
+            merged_input = {}
+            for input_dict in combination:
+                merged_input.update(input_dict)
```
sfc-gh-chu, line 358: do we need to check for conflicts or is that handled in L310?

src/feedback/trulens/feedback/computer.py
```diff
@@ -73,6 +79,334 @@ def _compute_feedback(
...
+    record_id_to_record_roots = _map_record_id_to_record_roots(events)
+    unflattened_inputs = _validate_unflattened_inputs(
+        unflattened_inputs,
+        kwarg_groups,
+        list(record_id_to_record_roots.keys()),
+        feedback_name,
+        error_messages,
+    )
+    flattened_inputs = _flatten_inputs(unflattened_inputs)
+    _run_feedback_on_inputs(
```
sfc-gh-chu, line 132: if we have at least one valid input, can we clear error_messages before calling _run_feedback_on_inputs? Might be confusing for users if errors are thrown related to selectors but the feedback computation is still successful

sfc-gh-dkurokawa: what about instead of throwing errors, I just give warnings in all cases. Are you good with that?

src/feedback/trulens/feedback/computer.py
sfc-gh-dkurokawa: we shouldn't have conflicts here because each kwarg_group has different kwargs and the input_dict should only have as keys the kwargs in that group.

src/feedback/trulens/feedback/computer.py
sfc-gh-chu: I feel an error should still be thrown if no successful eval spans are created (either bc no inputs or no successful feedback_function calls). But with partial failures, we can consider them warnings

sfc-gh-dkurokawa: okay done.

sfc-gh-chu (approved):

PR #1918, opened 2025-04-15
Fix `test_deprecation.py`.

sfc-gh-dkurokawa:
# Description
The `Makefile`'s `test-env-optional` overrides the `poetry.lock` file's `langchain-core` version and the `langchain-core` package was recently updated where `__getattr__` is overwritten incorrectly causing some failures during a test.

Until they fix this, I've pinned the `langchain-core` version to a lower one.

## Type of change
  not work as expected)
- [ ] New Tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,6 +51,7 @@ env-tests-optional: env env-tests
 		chromadb \
 	 	faiss-cpu \
 		langchain-openai \
+		'langchain-core<0.3.52' \
 		llama-index-embeddings-huggingface \
 		llama-index-embeddings-openai \
 		unstructured


sfc-gh-nvytla:
Is there a corresponding bug on the langchain repo? The most relevant change seems like this one: https://github.com/langchain-ai/langchain/pull/30769

sfc-gh-chu (approved):

sfc-gh-dkurokawa:
> Is there a corresponding bug on the langchain repo? The most relevant change seems like this one: [langchain-ai/langchain#30769](https://github.com/langchain-ai/langchain/pull/30769)

Yeah this is definitely the change that's causing this. I believe they need to throw an `AttributeError` if they want to defer to the default value but they don't appear to be which is causing the actual error to be thrown up.

sfc-gh-nvytla (approved):

Issue #1928, opened 2025-04-17
[BUG] Please remove poetry from list of dependencies


drooms-sandrus:
Hello,

I noticed that _poetry_ is listed as regular dependency in [pyproject.toml](https://github.com/truera/trulens/blob/main/pyproject.toml#L30).

This is causing downstream issues in dependency resolution on our side (as poetry<2 depends on outdated packages).

The restriction to poetry<2 was added in [this](https://github.com/truera/trulens/pull/1709) MR, with the stated reason being failing Python 3.8 tests. Python 3.8 support ended 6 months ago (see [here](https://endoflife.date/python)). Further, according to the pyproject.toml file linked above, Python 3.8 is not supported by Trulens anyway.

Therefore, please remove the restriction to poetry<2 from pyproject.toml.

If the restriction to poetry <2 absolutely needs to be in the file at the moment, I strongly recommend to move it to dev dependencies, as it's only required for development purposes (and according to the original MRs description, it's only added due to failing tests). This way, it would not affect dependency resolution in projects that import Trulens.

We temporarily solved this issue on our side by downgrading from Trulens 1.4.9 to 1.2.11, but this is not a long-term solution.

sfc-gh-jreini:
@sfc-gh-chu has an open PR to solve this - thanks for reporting @drooms-sandrus !

https://github.com/truera/trulens/pull/1930

PR #1930, opened 2025-04-17
Unlock poetry > 2.0

sfc-gh-chu:
# Description

No longer need this constraint since we deprecated python 3.8

Related issue: https://github.com/truera/trulens/issues/1928

## Type of change

  not work as expected)
- [ ] New Tests

----

> [!IMPORTANT]
> Remove `poetry <2.0.0` constraint from `pyproject.toml` due to Python 3.8 deprecation.
> 
>   - **Dependencies**:
>     - Removed `poetry = "<2.0.0"` constraint from `pyproject.toml` due to deprecation of Python 3.8.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=truera%2Ftrulens&utm_source=github&utm_medium=referral)<sup> for 05bb5e409cca36be2ed1dc9bf112145ec4987024. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,6 @@ classifiers = [
 ]
 
 [tool.poetry.dependencies]
-poetry = ">=1.8.0"
 python = "^3.9,!=3.9.7"
 trulens-core = { version = "^1.0.0", extras = [
   "otel",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ classifiers = [
 ]
 
 [tool.poetry.dependencies]
-poetry = "<2.0.0"
+poetry = ">=1.8.0"
 python = "^3.9,!=3.9.7"
 trulens-core = { version = "^1.0.0", extras = [
   "otel",
--- a/poetry.lock
+++ b/poetry.lock
@@ -63,92 +63,92 @@ reference = "pypi-public"
 
 [[package]]
 name = "aiohttp"
-version = "3.11.14"
+version = "3.11.16"
 description = "Async http client/server framework (asyncio)"
 optional = false
 python-versions = ">=3.9"
 files = [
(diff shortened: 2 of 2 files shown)


sfc-gh-jreini (approved):

pyproject.toml
```diff
@@ -27,7 +27,7 @@ classifiers = [
 ]
 
 [tool.poetry.dependencies]
-poetry = "<2.0.0"
+poetry = ">=1.8.0"
```
sfc-gh-nvytla, line 30: Just curious, why not `poetry = ">=2.0.0"`? Do we still have any dependencies on 1.8.x and below?

sfc-gh-dkurokawa: I'm actually confused why we need poetry at all to run anything.

sfc-gh-chu: @sfc-gh-dkurokawa fair point. We actually don't need this constraint here at all really

sfc-gh-chu: @sfc-gh-nvytla we don't, this just lets us be permissive as there's no reason to disallow 1.8

sfc-gh-dkurokawa: Presumably we need it as some sort of build dependency, but it shouldn't be like a run dependency which I think `tool.poetry.dependencies` is right? Though I defer to your `poetry` expertise @coreyhu

sfc-gh-chu: Yea build deps are defined that under `build-system.requires`

sfc-gh-dkurokawa (approved):
are you not going to put it in the build dependency place?

sfc-gh-chu:
> are you not going to put it in the build dependency place?

it's already there under poetry-core. We realistically don't need to add a version constraint now that we no longer need to support Python3.8

PR #1933, opened 2025-04-18
Fix `test_tru_llama.py` tests.

sfc-gh-dkurokawa:
# Description
Fix `test_tru_llama.py` tests.

This is quite a weird situation. There are a few things going on here:
1. The non-OTEL flow is broken in that it arbitrarily removed one decorator (if there was any) to any function it was instrumenting.
2. The non-OTEL flow is broken in that the instrument decoration is basically done via `functools.wraps`. This doesn't work well with wrapping a function that's already wrapped by `wrapt.decorator` like some `llama-index` functions and functions instrumented by the OTEL flow in `trulens`.
3. When an OTEL `llama-index` test like `tests/unit/test_otel_tru_llama.py::TestOtelTruLlama::test_smoke` would run before non-OTEL `llama-index` tests like `tests/e2e/test_tru_llama.py::TestLlamaIndex::test_query_engine_sync`, there would be a problem when the latter test ran in that the non-OTEL flow would try to wrap a function that's already wrapped by a `wrapt.decorator` twice (once from `llama-index` and once from OTEL-flow `trulens`). Although one of these two would be removed, the other would remain and due to point 2 (i.e. the inability for non-OTEL flow to handle `wrapt.decorator`), we'd run into an error. This didn't occur before OTEL flow tests were added because it just so happened that point 1 (i.e. arbitrarily removing one decorator) would often remove the problematic `wrapt.decorator` decorator and there'd only be one.

Fixing this properly requires a lot more work and understanding of the non-OTEL flow and using `wrapt.decorator` instead of `functools.wraps`. Frankly, this doesn't seem worth it to me given we're going to deprecate the non-OTEL flow so instead I've gone the route of having the non-OTEL flow remove all decorators (not just one). This is obviously not good either, but it makes somewhat more sense given it can't handle other decorators anyway.

This bug was ridiculous.

## Type of change
  not work as expected)
- [ ] New Tests

----

> [!IMPORTANT]
> Fixes non-OTEL flow in `instrument_object` by removing all decorators to resolve double-wrapping issues with `wrapt.decorator`.
> 
>   - **Behavior**:
>     - Fixes non-OTEL flow in `instrument_object` in `instruments.py` by replacing `if` with `while` to remove all decorators from functions.
>     - Resolves issues with double-wrapping functions already wrapped by `wrapt.decorator`.
>   - **Rationale**:
>     - Temporary fix due to planned deprecation of non-OTEL flow.
>     - Avoids complex changes by removing all decorators, acknowledging limitations in handling `wrapt.decorator`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=truera%2Ftrulens&utm_source=github&utm_medium=referral)<sup> for 842901c4a34268aee77d99cc0e87293bc5775d61. It will automatically update as commits are pushed.</sup>


sfc-gh-dhuang:
this is wild, so the behavior "This didn't occur before OTEL flow tests were added because it just so happened that point 1 (i.e. arbitrarily removing one decorator) would often remove the problematic wrapt.decorator decorator and there'd only be one." like you observed is mostly arbitrary?

sfc-gh-dhuang:
what I learned from GPT-o4-mini - seems reasonable given the test loading order is not fixed and can vary run by run:
The non‑OTEL flow’s single‑layer unwrap + functools.wraps approach inadvertently flattens the decorator stack and throws away one wrapper every time it sees a __wrapped__ attribute. When you mix wrapt‑decorator wrappers and OTEL’s own instrumentation, whichever wrapper happens to be on top gets chopped off—hence the “random” decorator removal. 

I'm assuming what happened was that If the function was itself a `wrapt`‑decorated function like you pointed in 2., we lose that deeper `__wrapped__` reference, because functools.wraps only preserves a single __wrapped__ pointer.  

anyhoo - thanks for looking into this! @sfc-gh-dkurokawa

sfc-gh-dhuang (approved):
LGTM modulo if the bot's suggestion is worth taking

PR #1934, opened 2025-04-18
allow blocking guardrails to take multiple feedbacks

sfc-gh-jreini:
# Description

Inspired by [OpenAI "A practical guide to building agents"](https://cdn.openai.com/business-guides-and-resources/a-practical-guide-to-building-agents.pdf), guardrails should allow the combination of multiple evaluations. This change allows you to pass a single feedback or list of feedbacks executed in parallel as guardrails.

Also update the example to show responding with a canned answer instead of None.

## Type of change

  not work as expected)
- [ ] New Tests

----

> [!IMPORTANT]
> Enhance `block_input` and `block_output` to support multiple feedbacks, updating documentation and tests accordingly.
> 
>   - **Behavior**:
>     - `block_input` and `block_output` in `base.py` now accept a single feedback or a list of feedbacks.
>     - If any feedback in the list fails, the input or output is blocked.
>     - Updated examples in `index.md` to demonstrate new functionality.
>   - **Tests**:
>     - Added tests in `test_guardrails.py` for multiple feedbacks in `block_input` and `block_output`.
>     - Tests cover both blocking and non-blocking scenarios with mixed feedback results.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=truera%2Ftrulens&utm_source=github&utm_medium=referral)<sup> for 5ac55fe4a40a9ac8a93074fa2fa0d18d72be3366. It will automatically update as commits are pushed.</sup>


src/core/trulens/core/guardrails/base.py
```diff
@@ -249,17 +269,34 @@ def __call__(self, func):
...
-                raise ValueError(
-                    "`block_output` can only be used with feedback functions that return a float."
-                )
-            if (self.feedback.higher_is_better and result < self.threshold) or (
-                not self.feedback.higher_is_better and result > self.threshold
-            ):
-                return self.return_value
-            else:
-                return output
+            # Support multiple feedback functions:
```
sfc-gh-dkurokawa, line 272: it seems like this and the `block_input` and `block_output` are practically the same. Can we just put these into a common function?

sfc-gh-dkurokawa: Also can you add some simple tests to `test_guardrails.py`? I imagine copilot can do that pretty easily haha.

docs/component_guides/guardrails/index.md
```diff
@@ -14,7 +14,7 @@ This mechanism for guardrails is supported via the `block_output` guardrail.
 
 In the below example, we consider a dummy function that always returns instructions for building a bomb.
 
-Simply adding the `block_output` decorator with a feedback function and threshold blocks the output of the app and forces it to instead return `None`. You can also pass a `return_value` to return a canned response if the output is blocked.
+Simply adding the `block_output` decorator with a feedback function (or list of feedback functions) and threshold blocks the output of the app and forces it to instead return `None`. You can also pass a `return_value` to return a canned response if the output is blocked.
```
sfc-gh-dkurokawa, line 17: to be clear, they all use the same threshold?

src/core/trulens/core/guardrails/base.py
sfc-gh-jreini: done

docs/component_guides/guardrails/index.md
sfc-gh-jreini: yes - I suppose this could be an issue if they're different directions. should we allow a list of thresholds for that case?

sfc-gh-jreini: If you want different thresholds, you can also stack the guardrails, e.g.

```python
class safe_input_chat_app:
    @instrument
    @block_input(
        feedback=f_criminality_input,
        threshold=0.9,
        keyword_for_prompt="question",
        return_value="I am not able to answer this question.",
    )
    @block_input(
        feedback=f_controversiality_input,
        threshold=0.7,
        keyword_for_prompt="question",
        return_value="I am not able to answer this question.",
    )
    def generate_completion(self, question: str) -> Optional[str]:
        """
```

sfc-gh-nvytla: From a user experience standpoint, would it be useful to allow an input of a list of thresholds? (e.g. maintaining a map of thresholds to feedbacks) so that someone could do this:

```python
    @block_input(
        feedback=[f_feedback1, f_feedback2],
        threshold=[0.7, 0.3],
        ...
    )
```

sfc-gh-jreini: Considered this, but I didn't want to create a new failure case, ie what do we do if len(feedbacks)<>len(thresholds) - error the app? Silently not use guardrails?

To me, UX seems better to allow the list of feedbacks but require the stacking decorators if you want different thresholds, but I'm not super firm on this opinion.

sfc-gh-nvytla: That's fair -- we could consider providing a warning to the user that the lengths of each list don't match, but that can be a decision for a later time.

One more UX question: do we expect users to use the same `keyword_for_prompt` and `return value` for multiple feedbacks? If those properties are not reusable / we expect them to heavily vary per feedback, then either we should allow users to also provide multiple keywords and multiple return values within a single decorator, or we should consider keeping the 1:1 mapping (decorator:feedback:threshold:keyword:return value).

tests/unit/test_guardrails.py
```diff
@@ -107,3 +107,51 @@ def chat(prompt: str) -> str:
...
+    def test_block_output_multiple_feedbacks(self):
+        threshold = 0.5
+
+        # Mix high and low feedback functions; the low one should trigger blocking.
+        @block_output([f_dummy_feedback_low, f_dummy_feedback_high], threshold)
+        def chat(prompt: str) -> str:
+            return "Response"
+
+        result = chat("example prompt")
+        self.assertEqual(result, None)
```
sfc-gh-nvytla, line 146: What happens if a user duplicates a feedback with stacking decorators? Do both persist? For example, is:

```python
class safe_input_chat_app:
    @instrument
    @block_input(
        feedback=f_criminality_input,
        threshold=0.7,
        ...
    )
    @block_input(
        feedback=[f_feedback1, f_criminality_input],
        threshold=0.7,
        ...
    )
```

equivalent to:
```python
    @block_input(
        feedback=[f_feedback1, f_criminality_input],
        threshold=0.7,
        ...
    )
```

?

examples/quickstart/blocking_guardrails.ipynb
```diff
@@ -28,7 +28,7 @@
    "source": [
```
sfc-gh-nvytla, line 28: <pre class="ql-syntax" spellcheck="false">Line #3.        chat.generate_completion("Is a hotdog a sandwich?")
</pre><p>Non-blocker (nit): pretty sure it's a taco according to the <a href="https://cuberule.com/" target="_blank">cube rule</a>.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/truera/trulens/pull/1934/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

sfc-gh-nvytla, line 28: <p>Nit: Does <code>block_output</code>  return <code>None</code> or does it return the canned response (based on the 2nd to last block of code)?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/truera/trulens/pull/1934/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='20'/>

sfc-gh-jreini: incredible

sfc-gh-jreini: <p>thanks - updated</p>

sfc-gh-nvytla (approved):

docs/component_guides/guardrails/index.md
sfc-gh-nvytla: @sfc-gh-jreini, to clarify:
- Simplicity-wise, I agree that encouraging users to stack decorators could be more readable, although we'll have to get user feedback and usage stats over time to determine if this is effective.
- Separately, I could imagine a world where we allow input of either single values for each OR lists of the  `feedback`, `threshold`, and `return_value` params (although in the latter we would have to enforce that all lists of params are the same length) -- this could be something we track in the backlog as a feature improvement and re-assess if relevant user feedback comes along.

Regardless, I agree with your proposal to add docs on stacking!

PR #1944, opened 2025-04-26
Increase Cortex complete timeout to 60s for tests.

sfc-gh-dkurokawa:
# Description
Increase Cortex complete timeout to 60s for tests.

It's unclear what the default is as the documentation is kinda lacking but I'm imagining it doesn't retry. I've made it 60s as of the last four failed e2e test runs on main, three of them were caused due to Cortex complete not working for w/e reason.

## Type of change
  not work as expected)
- [ ] New Tests

----

> [!IMPORTANT]
> Increase Cortex completion timeout to 60 seconds in tests and remove unused imports and parameters.
> 
>   - **Behavior**:
>     - Increase `retry_timeout` to 60s in `Cortex` class in `provider.py`.
>     - Set `timeout=60` in `respond_to_query` in `test_otel_costs.py`.
>     - Pass `retry_timeout=60` to `Cortex` in `test_snowflake_feedback_evaluation.py`.
>   - **Tests**:
>     - Remove unused `List` import and `cost_functions` parameter in `test_otel_costs.py`.
>     - Remove unused imports in `test_otel_costs.py`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=truera%2Ftrulens&utm_source=github&utm_medium=referral)<sup> for 1b01a421ab41d84c86613bd3095b48825c8065f2. You can [customize](https://app.ellipsis.dev/truera/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/tests/e2e/test_otel_costs.py
+++ b/tests/e2e/test_otel_costs.py
@@ -1,5 +1,5 @@
 import os
-from typing import Any, List
+from typing import Any
 import unittest
 
 from langchain_community.chat_models import ChatSnowflakeCortex
@@ -34,6 +34,7 @@ class _TestCortexApp:
             model="mistral-large2",
             prompt=query,
             session=self._snowpark_session,
+            timeout=60,
         )
 
 
@@ -84,7 +85,6 @@ class TestOtelCosts(OtelTestCase):
     def _test_tru_custom_app(
         self,
         app: Any,
-        cost_functions: List[str],
         model: str,
         currency: str,
         num_expected_spans: int = 1,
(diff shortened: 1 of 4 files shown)


tests/e2e/test_otel_costs.py
```diff
@@ -84,7 +85,6 @@ class TestOtelCosts(OtelTestCase):
     def _test_tru_custom_app(
         self,
         app: Any,
-        cost_functions: List[str],
```
sfc-gh-nvytla, line 87: If we're removing these here, are we testing cost_functions elsewhere?

sfc-gh-dkurokawa: oh it's not used anymore. I think I initially had this before because we used to basically write the costs in its own span, but now they're just written in whatever span is currently in context (hence why I told you before in your PR that the cost info could be in any span type) as Anubhav and Josh found the prior behavior confusing.

Anyway, at that time I could test what function emitted the cost info, because it would emit a span. Now that that's not true, this info isn't really used.

sfc-gh-nvytla (approved):

PR #1939, opened 2025-04-22
[SNOW-2005227] v0 OTEL implementation of get_records_and_feedback

sfc-gh-nvytla:
# Description

JIRA: [SNOW-2005227](https://snowflakecomputing.atlassian.net/browse/SNOW-2005277)

## Other details good to know for developers

Future TODOs (v1):
- add `higher_is_better` as a span attribute ([see comment](https://github.com/truera/trulens/pull/1939#discussion_r2054799605))
  - [SNOW-2061190](https://snowflakecomputing.atlassian.net/browse/SNOW-2061190)
- RecordViewer considerations ([see design doc](https://docs.google.com/document/d/1L_qs78Y522EgYdeysqW7IZsMuQx0V8jVrkgOQoudYGo/edit?tab=t.0#heading=h.ev4cohe70qle))
  - [SNOW-2061186](https://snowflakecomputing.atlassian.net/browse/SNOW-2061186)
- Return call stack -- where each entry maps to a span ([see comment](https://github.com/truera/trulens/pull/1939#discussion_r2056569392))
  - [SNOW-2061194](https://snowflakecomputing.atlassian.net/browse/SNOW-2061194)
- Handle mixed costs/currencies ([see comment](https://github.com/truera/trulens/pull/1939#discussion_r2054802093))
  - [SNOW-2061174](https://snowflakecomputing.atlassian.net/browse/SNOW-2061174)
- Handle pagination (offset/limit) based on unique record IDs in events table
  - [SNOW-2081987](https://snowflakecomputing.atlassian.net/browse/SNOW-2081987) 

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to
  not work as expected)
- [x] New Tests
- [ ] This change includes re-generated golden test results
- [ ] This change requires a documentation update

----

> [!IMPORTANT]
> Adds OTEL tracing support to `get_records_and_feedback` with new methods, flags, and tests for enhanced telemetry handling.
> 
>   - **Behavior**:
>     - Adds OTEL tracing support to `get_records_and_feedback` in `sqlalchemy.py` with `_get_records_and_feedback_otel` method.
>     - Introduces `use_otel` flag to switch between OTEL and pre-OTEL implementations.
>     - Updates `run_dashboard` and `run_dashboard_sis` in `run.py` to accept `otel_tracing` flag.
[...]

PR #1951, opened 2025-05-01
[SNOW-2081987] Enable OTEL record pagination based on unique, sorted record_ids from event table

sfc-gh-nvytla:
# Description

[SNOW-2081987](https://snowflakecomputing.atlassian.net/browse/SNOW-2081987)

This PR:
- Enables proper OTEL record pagination by unique `record_id`s
- Adds `app_id` as a new OTEL `SpanAttribute` ([comment](https://github.com/truera/trulens/pull/1951#pullrequestreview-2813044876))
- Integrates logic for computing OTEL `app_id`s
- Moves the `app_id` filter from post-query into the query itself
  - This resolves a pesky issue with potentially not being able to return all relevant records and improves query efficiency, especially on large datasets ([comment](https://github.com/truera/trulens/pull/1951#discussion_r2070403103) and [discussion](https://snowflake.slack.com/archives/C08PWJU4HJ6/p1746140717654759)))
- Adds tests for pagination (offset, limit) and filtering

Follow-up task from #1939 

## Type of change

  not work as expected)
- [x] New Tests

----

> [!IMPORTANT]
> Enhances OTEL record pagination and filtering by app_id, improves query efficiency, and adds tests for these functionalities.
> 
>   - **Behavior**:
>     - Enables OTEL record pagination by unique `record_id` in `sqlalchemy.py`.
>     - Moves `app_id` filtering into the query for efficiency.
>     - Integrates logic for computing OTEL `app_id` in `app.py`.
>   - **Functions**:
>     - Adds `_get_paginated_record_ids_otel()` in `sqlalchemy.py` for paginated record ID queries.
>     - Updates `_get_records_and_feedback_otel()` in `sqlalchemy.py` to use paginated queries.
>   - **Tests**:
>     - Adds tests for pagination and filtering in `test_otel_get_records_and_feedback.py`.
>     - Updates JSON test data files for OTEL spans.
>   - **Misc**:
>     - Removes `_compute_app_id_otel()` from `sqlalchemy.py` and uses `AppDefinition._compute_app_id()` instead.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=truera%2Ftrulens&utm_source=github&utm_medium=referral)<sup> for 8bd164270381505a289c3f91a14dd2c887eab179. You can [customize](https://app.ellipsis.dev/truera/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


src/core/trulens/core/database/sqlalchemy.py
```diff
@@ -979,6 +979,58 @@ def get_records_and_feedback(
...
+            app_name: App name to filter by. Defaults to None.
+            offset: Offset for pagination. Defaults to None.
+            limit: Limit for pagination. Defaults to None.
+
+        Returns:
+            A SQLAlchemy select statement for record IDs with pagination applied.
+        """
+        # First, get unique record IDs with pagination
+        record_id_subquery = (
+            sa.select(
```
sfc-gh-dkurokawa, line 1002: if the DB gets events without record ids this won't fail will it? Also it won't consider that as a unique record id will it?

src/core/trulens/core/database/sqlalchemy.py
```diff
@@ -979,6 +979,58 @@ def get_records_and_feedback(
...
+                sa.func.json_extract(
+                    self.orm.Event.record_attributes,
+                    f'$."{SpanAttributes.RECORD_ID}"',
+                ).label("record_id"),
+                self.orm.Event.start_timestamp,
+            )
+            .group_by("record_id")
+            .order_by(self.orm.Event.start_timestamp.desc())
```
sfc-gh-dkurokawa, line 1010: I'm somewhat confused what's going on here, the start timestamp isn't grouped. Does it somehow take the min of it?

Also now that I think about it, given we require a "record_root" span_type maybe we should just filter to those to begin with. That way we know that'll be there, and if you are taking the min start timestamp, that won't change if new events come in with weird timestamps or something.

src/core/trulens/core/database/sqlalchemy.py
```diff
@@ -996,36 +1048,23 @@ def _get_records_and_feedback_otel(
...
+                session, app_name, offset, limit
+            )
 
-            # TODO(SNOW-2081987): Apply pagination on unique record IDs, not events table (this is essentially a full-table scan)
-            if limit is not None:
-                stmt = stmt.limit(limit)
-            if offset is not None:
-                stmt = stmt.offset(offset)
+            # Now get all events for those record IDs
+            stmt = sa.select(self.orm.Event).where(
```
sfc-gh-dkurokawa, line 1062: Because we're filtering the app ids after the query is executed, we may technically return less than the `limit` number of records. In fact, we could theoretically return 0. Not really sure there's an easy fix for that, but something that could be a problem.

src/core/trulens/core/database/sqlalchemy.py
sfc-gh-nvytla: I'll change this to grab the min start timestamp and to filter to "record_root" spans. Then I'll order by later.

src/core/trulens/core/database/sqlalchemy.py
sfc-gh-nvytla: Can the DB get events without record ids? I thought Pydantic model validation would reject an insert into the Event table if a required attribute like record_id is missing...

sfc-gh-nvytla: Oops, never mind, have a missing attribute is one thing, but Pydantic does not do checks for empty or None value record_ids. I'll implement that here so that we only grab populated, unique (non-empty/null) record ids

src/core/trulens/core/database/sqlalchemy.py
sfc-gh-nvytla: I proposed a fix that should be able to let us filter by app_ids within the query itself ([discussion](https://snowflake.slack.com/archives/C08PWJU4HJ6/p1746213956083409?thread_ts=1746140717.654759&cid=C08PWJU4HJ6)).

I also added tests to confirm if the appropriate records are being returned.

sfc-gh-nvytla:
> I was talking to @sfc-gh-gtokernliang and he pointed out we should just emit the app id into the span attributes since we haven't released anything yet so we can just do that from now on and use that in the query.

@sfc-gh-dkurokawa, gotcha, just updated the PR to handle `app_id`s emitted into `SpanAttributes`.

src/core/trulens/core/database/sqlalchemy.py
```diff
@@ -1272,6 +1229,95 @@ def _get_records_and_feedback_otel(
 
             return df, feedback_col_names
 
+    def get_records_and_feedback(
```
sfc-gh-dkurokawa, line 1232: to be clear, I didn't check this function as I think your'e saying you didn't change it and just moved it.

sfc-gh-dkurokawa (approved):

PR #1989, opened 2025-05-20
Make old `@instrument` decorator work in OTel world.

sfc-gh-dkurokawa:
# Description
Make old `@instrument` decorator work in OTel world.

I had a previous PR to do something similar to this in https://github.com/truera/trulens/pull/1765, but that allows even the new OTel `@instrument` decorator to work without parentheses --- which I'm not sure we want to allow yet.

## Type of change
  not work as expected)
- [x] New Tests

----

> [!IMPORTANT]
> Introduces `legacy_instrument` for legacy `@instrument` compatibility with OTel, with conditional assignment and new tests.
> 
>   - **Behavior**:
>     - Introduces `legacy_instrument` class in `app.py` to support legacy `@instrument` decorator.
>     - Conditionally assigns `instrument` to `otel_instrument` or `legacy_instrument` based on `is_otel_tracing_enabled()`.
>   - **Testing**:
>     - Adds `test_otel_legacy_compatibility.py` to verify legacy `@instrument` compatibility with OTel.
>     - Tests ensure `instrument` is correctly assigned and legacy apps function with OTel tracing enabled.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=truera%2Ftrulens&utm_source=github&utm_medium=referral)<sup> for 79a585a280fc6719b843f9ef02e1a0a47b1c7b79. You can [customize](https://app.ellipsis.dev/truera/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/tests/e2e/test_snowflake_event_table_exporter.py
+++ b/tests/e2e/test_snowflake_event_table_exporter.py
@@ -146,7 +146,7 @@ class TestSnowflakeEventTableExporter(SnowflakeTestCase):
             TruApp,
             {"input": "custom_input"},
             pd.DataFrame({"custom_input": ["Kojikun", "Nolan"]}),
-            10,
+            8,
         )
 
     def test_tru_llama(self):
--- a/tests/unit/test_otel_legacy_compatibility.py
+++ b/tests/unit/test_otel_legacy_compatibility.py
@@ -2,11 +2,20 @@ import importlib
 import os
 
 import trulens.apps.app
+from trulens.apps.app import TruApp
 from trulens.apps.app import instrument as legacy_instrument
 from trulens.core.otel.instrument import instrument as otel_instrument
 from trulens.core.otel.utils import is_otel_tracing_enabled
+from trulens.core.session import TruSession
+from trulens.otel.semconv.trace import SpanAttributes
 
 from tests.util.otel_test_case import OtelTestCase
+
+
+class _LegacyTestApp:
+    @legacy_instrument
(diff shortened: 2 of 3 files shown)


sfc-gh-chu (approved):

sfc-gh-jreini:
Why rename to `legacy_instrument` rather than keeping the same? This will force users to change their code (even if they just alias back to instrument on import)

sfc-gh-dkurokawa:
> Why rename to `legacy_instrument` rather than keeping the same? This will force users to change their code (even if they just alias back to instrument on import)

So I'm just renaming it in that file, later on in the file, I declare what `instrument` is. If OTel tracing is enabled, I use the new decorator with default arguments. Otherwise, I use the old one. So basically old code in non-OTel mode should continue to work and old code in OTel mode should basically assume `@instrument` means `@instrument()`.

Does that make sense?

sfc-gh-jreini:
> > Why rename to `legacy_instrument` rather than keeping the same? This will force users to change their code (even if they just alias back to instrument on import)
> 
> So I'm just renaming it in that file, later on in the file, I declare what `instrument` is. If OTel tracing is enabled, I use the new decorator with default arguments. Otherwise, I use the old one. So basically old code in non-OTel mode should continue to work and old code in OTel mode should basically assume `@instrument` means `@instrument()`.
> 
> Does that make sense?

this makes sense - thanks!

PR #2000, opened 2025-05-22
Work on streamlit UI components, streaming for OTEL

sfc-gh-jreini:
# Description

- update snowflake-ai-stack to use otel
- fix streaming for openai endpoint
- add new helper wait_for_record
- make trulens_trace work with record_id (mostly from @sfc-gh-nvytla's PR + one fix )

https://github.com/user-attachments/assets/52d4011a-e2af-4167-a10f-f574589479e7

## Type of change

  not work as expected)
- [ ] New Tests

----

> [!IMPORTANT]
> Integrate OpenTelemetry with snowflake-ai-stack, fix streaming, and enhance observability and tracing mechanisms.
> 
>   - **Behavior**:
>     - Integrate OpenTelemetry (OTEL) with snowflake-ai-stack by setting `TRULENS_OTEL_TRACING` in `run_app1.sh` and `run_app2_filters.sh`.
>     - Fix streaming for OTEL in `app.py` by using `retrieve_and_generate_stream()` and handling `record_id` with `wait_for_record()`.
>     - Add `wait_for_record()` in `session.py` to wait for specific `record_id`.
>   - **Observability**:
>     - Update `create_evals()` in `observability.py` to use `Selector` for span attributes.
>     - Modify `instrument` decorators in `rag.py` to specify `span_type` and attributes.
>   - **Misc**:
>     - Handle legacy and streamed responses in `endpoint.py` with `prepend_first_chunk()`.
>     - Remove unused imports and fix minor issues in `generation.py` and `streamlit.py`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=truera%2Ftrulens&utm_source=github&utm_medium=referral)<sup> for c09164ad82f9e7fbe46cace7223a81be70387b6d. You can [customize](https://app.ellipsis.dev/truera/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/examples/expositional/use_cases/snowflake-ai-stack/app.py
+++ b/examples/expositional/use_cases/snowflake-ai-stack/app.py
@@ -95,7 +95,7 @@ if user_input:
 
         st.session_state.tru_session.force_flush()
         record_id = recording.get()
-        st.session_state.tru_session.wait_for_record(record_id) # This would wait for the events to show up in the table.
+        st.session_state.tru_session.wait_for_record(record_id)
         trulens_st.trulens_trace(record=record_id)
 
         # Add the assistant response to session state - only once!


examples/expositional/use_cases/snowflake-ai-stack/app.py
```diff
@@ -82,18 +83,28 @@ def setup_vector_store():
...
             with st.spinner("Thinking..."):
+                # full_response = st.session_state.rag.retrieve_and_generate(user_input, st.session_state.messages)
+                # message_area.markdown(full_response)
                 generator = st.session_state.rag.retrieve_and_generate_stream(user_input, st.session_state.messages)
                 for chunk in generator:
                     if chunk is not None:
                         full_response += chunk
                         message_area.markdown(full_response)
-            # Display TruLens feedback and metrics
-            record = recording.get()
```
sfc-gh-dkurokawa, line 94: this will work now, but you're not guaranteed to have the events be in the DB. You can do a `force_flush` to expediate it though.

examples/expositional/use_cases/snowflake-ai-stack/src/observability.py
```diff
@@ -47,8 +54,18 @@ def create_evals(provider: LLMProvider = None):
             provider.context_relevance, name="Context Relevance",
             criteria = context_relevance_custom_criteria,
         )
-        .on(Select.Record.app.retrieve.args.query)
-        .on(Select.RecordCalls.retrieve.rets[:])
+        .on({
```
sfc-gh-dkurokawa, line 57: any thoughts on this new selector approach? it definitely looks more verbose than the older one here (though obviously that doesn't mean it's more confusing necessarily).

examples/expositional/use_cases/snowflake-ai-stack/src/rag.py
```diff
@@ -49,30 +50,48 @@ def _retrieve_plain(self, query: str):
...
+        },
+    )
     def retrieve(self, query: str):
         if int(self.use_context_filter)==1:
             return self._retrieve_filtered(query)
         else:
             return self._retrieve_plain(query)
 
-    @instrument
+    instrument(span_type=SpanAttributes.SpanType.GENERATION)
```
sfc-gh-dkurokawa, line 66: you dropped this `@`

src/core/trulens/core/session.py
```diff
@@ -1171,6 +1172,39 @@ def stop_evaluator(self):
 
         self._evaluator_proc = None
 
+    def wait_for_new_record_id(tru_session, timeout=10, poll_interval=0.5):
```
sfc-gh-dkurokawa, line 1175: put the types in here instead of in the docstring.

sfc-gh-dkurokawa: Also I think you don't need this anymore --- might need something that waits for events to show up in the event table though.

src/dashboard/trulens/dashboard/streamlit.py
```diff
@@ -331,7 +331,8 @@ def trulens_trace(record: Union[record_schema.Record, str]):
     """
 
     session = core_session.TruSession()
-    app = session.get_app(app_id=record.app_id)
+    if isinstance(record, record_schema.Record):
+        app = session.get_app(app_id=record.app_id)
```
sfc-gh-dkurokawa, line 335: wait, shouldn't this `if` be coupled with the `elif` below?

src/providers/openai/trulens/providers/openai/endpoint.py
```diff
@@ -77,20 +77,35 @@
...
                 raise ValueError("OpenAI API returned non-200 status code!")
             response = response.parse()
+
         endpoint = OpenAIEndpoint()
         callback = OpenAICallback(endpoint=endpoint)
         model_name = ""
-        if getattr(response, "model"):
+
+        if hasattr(response, "__iter__") and not hasattr(response, "model"):
+            try:
```
sfc-gh-dkurokawa, line 91: sorry I'm confused, what's going on here? why can't we just use `response` directly?

sfc-gh-dkurokawa:
PR 2000!

examples/expositional/use_cases/snowflake-ai-stack/src/observability.py
sfc-gh-jreini: I like the matching between span attributes and the Selector. should be easier to figure out.

My biggest concern here is that we're now requiring the user to know the argument names for our evals

src/providers/openai/trulens/providers/openai/endpoint.py
sfc-gh-jreini: if response is a generator, it doesn't have model attribute. in this case, I'm getting model attribute from first chunk. (note: I just pushed the line where I actually set the model name here, was missing it before)

src/dashboard/trulens/dashboard/streamlit.py
sfc-gh-jreini: I don't think this is related to sis compatibility, more about otel v non-otel. am I missing something?

examples/expositional/use_cases/snowflake-ai-stack/app.py
sfc-gh-jreini: What does force_flush do?

Somewhat unrelated, will record.wait_for_feedback_result work? https://github.com/truera/trulens/blob/09c6ae63e031ea80fa793e2692b9ce0f193e540a/src/core/trulens/core/schema/record.py#L237

src/core/trulens/core/session.py
sfc-gh-jreini: updated the typing.

only difference for something that waits for new events would be just returning True instead of record_id?

examples/expositional/use_cases/snowflake-ai-stack/app.py
sfc-gh-dkurokawa: `TruSession::force_flush` will force the OTel exporter to flush it's buffer --- if you look at a lot of my tests, I do this before reading the DB since I don't want to wait around for it. When writing to like a sqlite DB locally this is pretty much instant. If writing to Snowflake there's overhead though.

sfc-gh-dkurokawa: yeah so basically if this is local sqlite DB you can probably get away with just:
```python
TruSession().force_flush()
record_id = recording.get()
trulens_st.trulens_feedback(record=record_id)
trulens_st.trulens_trace(record=record_id)
```
But you can also wait by doing something like:
```python
TruSession().force_flush()
record_id = recording.get()
wait_for_record(record_id) # This would wait for the events to show up in the table.
trulens_st.trulens_feedback(record=record_id)
trulens_st.trulens_trace(record=record_id)
```

sfc-gh-jreini: cool - I updated and implemented wait_for_record as you suggested

src/providers/openai/trulens/providers/openai/endpoint.py
```diff
@@ -481,3 +497,16 @@ def handle_chunk(chunk: T) -> T:
             )
 
         return response
+
+
+def prepend_first_chunk(stream, first_chunk):
+    # Save the original iterator
+    original_iter = stream._iterator
```
sfc-gh-dhuang, line 504: I'm a little confused why we have to modify the internal `_iterator` attribute here. 


It looks like all we need is to be able to "re-yield" the first chunk and then continue yielding all the remaining items from the stream right?


if so, could we just do:
```
def prepend_first_chunk(stream, first_chunk):
    def generator():
        yield first_chunk
        yield from stream
    return generator()
```

or even simpler
```
def prepend_first_chunk(stream, first_chunk):
    """Yields first_chunk, then all items from stream."""
    yield first_chunk
    yield from stream
```

sfc-gh-jreini: I think this doesn't work bc the downstream consumer of the stream relies on an intact stream._iterator

sfc-gh-dkurokawa (approved):

PR #2003, opened 2025-05-24
[Feat] Dedup feedback for same metric name and inputs on Streamlit UI

sfc-gh-nvytla:
# Description

- Sometimes, we may recompute feedbacks for the same metric name and given the same inputs. In this case, the UI should only consider the latest one (e.g. when showing a full table of feedback calls and in aggregations).

## Type of change

  not work as expected)
- [x] New Tests

### DIFF
--- a/src/core/trulens/core/database/sqlalchemy.py
+++ b/src/core/trulens/core/database/sqlalchemy.py
@@ -1124,13 +1124,20 @@ class SQLAlchemyDB(core_db.DB):
                             )
 
                         # Add call data
-                        # Extract kwargs by finding all attributes that start with the KWARGS prefix
-                        kwargs_prefix = SpanAttributes.CALL.KWARGS + "."
-                        kwargs = {
-                            key[len(kwargs_prefix) :]: value
-                            for key, value in record_attributes.items()
-                            if key.startswith(kwargs_prefix)
-                        }
+                        # Extract namespaced attributes using the helper method
+                        kwargs = self._extract_namespaced_attributes(
+                            record_attributes, SpanAttributes.CALL.KWARGS
+                        )
+                        args_span_id = self._extract_namespaced_attributes(
+                            record_attributes,
+                            SpanAttributes.EVAL_ROOT.ARGS_SPAN_ID,
+                        )
+                        args_span_attribute = (
+                            self._extract_namespaced_attributes(
+                                record_attributes,
+                                SpanAttributes.EVAL_ROOT.ARGS_SPAN_ATTRIBUTE,
+                            )
+                        )
 
                         call_data = {
                             # TODO(SNOW-2112879): Call data may not be populated in the OTEL spans yet
@@ -1153,6 +1160,12 @@ class SQLAlchemyDB(core_db.DB):
                                 "explanation": record_attributes.get(
                                     SpanAttributes.EVAL.EXPLANATION, None
                                 ),
+                                "eval_root_id": record_attributes.get(
+                                    SpanAttributes.EVAL.EVAL_ROOT_ID, None
+                                ),
+                                "timestamp": record_data["ts"],
+                                "args_span_id": args_span_id,
+                                "args_span_attribute": args_span_attribute,
                             },
                         }
                         feedback_result["calls"].append(call_data)
@@ -1595,6 +1608,25 @@ class SQLAlchemyDB(core_db.DB):
 
             # Execute query and return as DataFrame
             return pd.read_sql(q, session.bind)
+
+    def _extract_namespaced_attributes(
+        self, record_attributes: Dict[str, Any], namespace_prefix: str
+    ) -> Dict[str, Any]:
+        """Extract attributes that are namespaced under a given prefix.
+
+        Args:
+            record_attributes: Dictionary of all span attributes
+            namespace_prefix: The namespace prefix to look for (e.g., "ai.observability.call.kwargs")
+
+        Returns:
+            Dictionary with namespace-stripped keys and their values
+        """
+        prefix_with_dot = namespace_prefix + "."
+        return {
+            key[len(prefix_with_dot) :]: value
+            for key, value in record_attributes.items()
+            if key.startswith(prefix_with_dot)
+        }
 
 
 # Use this Perf for missing Perfs.
--- a/src/dashboard/trulens/dashboard/utils/records_utils.py
+++ b/src/dashboard/trulens/dashboard/utils/records_utils.py
@@ -77,6 +77,10 @@ def display_feedback_call(
         df["meta"] = pd.Series([call[i]["meta"] for i in range(len(call))])
         df = df.join(df.meta.apply(lambda m: pd.Series(m))).drop(columns="meta")
 
+        # Filter to only show the latest results for calls with the same args_span_id and args_span_attribute
+        # NOTE: this is an optimization to not show duplicate calls in the UI (e.g. when a feedback function is recomputed on the same metric name and inputs)
+        df = _filter_duplicate_span_calls(df)
+
         # note: improve conditional to not rely on the feedback name
         if "groundedness" in feedback_name.lower():
             df = expand_groundedness_df(df)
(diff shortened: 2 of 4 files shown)


sfc-gh-dkurokawa:
> I'm not sure deduping here makes sense - user would probably like to know if they are duplicating feedback computation (as its costly)?
> 
> Alternatively, more of a roadmap item - it would be useful to cache feedback computations with the same inputs.

So we shouldn't actually recompute feedback functions with the same provenance of inputs (i.e. that came from the same span and attribute) if they're already in the DB. The issue comes when we compute feedbacks and before they show up in the DB, we compute them again. So we can't really cache them as a result.

src/dashboard/trulens/dashboard/utils/records_utils.py
```diff
@@ -103,6 +107,63 @@ def display_feedback_call(
...
+    """Filter to only show rows from the most recent eval_root_id for calls with the same args_span_id and args_span_attribute.
+
+    Args:
+        df: DataFrame containing feedback call data with potential duplicates
+
+    Returns:
+        DataFrame with duplicate span calls filtered to show only rows from the most recent eval_root_id
+    """
+    if (
+        "eval_root_id" in df.columns
```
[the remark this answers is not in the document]
sfc-gh-nvytla: The conditional here relies on those four fields to be populated via span attributes. We need all four of them in order to dedup appropriately.

The structure of the df here is:
1. all columns from the records dataframe returned by `_get_records_and_feedback_otel()` and
2. the "meta" field is flattened and columnized -- all keys within the "meta" dict (first defined in sqlalchemy.py) are now separate columns

tests/unit/data/test_otel_spans/eval_root_span.json
```diff
@@ -22,7 +22,7 @@
         "ai.observability.app_name": "coffee_rag",
         "ai.observability.app_version": "openai",
         "ai.observability.app_id": "app_hash_65b218a1a09632779ea4c5f8ad9709e4",
-        "ai.observability.eval.eval_root_id": "98bbSpP2",
+        "ai.observability.eval.eval_root_id": "7wvRtVbg",
```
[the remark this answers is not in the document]
sfc-gh-nvytla: This was an artifact from a prior test that did not get updated -- the static eval_root and eval spans should all have the same eval_root_id for testing.

src/dashboard/trulens/dashboard/utils/records_utils.py
```diff
@@ -103,6 +107,63 @@ def display_feedback_call(
...
+        and "args_span_id" in df.columns
+        and "args_span_attribute" in df.columns
+        and "timestamp" in df.columns
+        and not df.empty
+    ):
+        # Convert dictionary columns to strings for grouping (since dicts are not hashable)
+        df_for_grouping = df.copy()
+        df_for_grouping["args_span_id_str"] = df_for_grouping[
+            "args_span_id"
+        ].apply(lambda x: str(x) if x is not None else "None")
```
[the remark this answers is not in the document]
sfc-gh-nvytla: Fixed!

src/core/trulens/core/database/sqlalchemy.py
```diff
@@ -1199,6 +1204,12 @@ def _get_records_and_feedback_otel(
                                     "metadata": record_attributes.get(
                                         SpanAttributes.EVAL.METADATA, {}
                                     ),
+                                    "eval_root_id": record_attributes.get(
+                                        SpanAttributes.EVAL.EVAL_ROOT_ID, None
```
sfc-gh-dkurokawa: nit: don't need the `None`.

src/core/trulens/core/database/sqlalchemy.py
```diff
@@ -1199,6 +1204,12 @@ def _get_records_and_feedback_otel(
...
+                                    ),
+                                    "timestamp": record_data["ts"],
```
sfc-gh-dkurokawa: is this the start or end timestamp?

src/dashboard/trulens/dashboard/utils/records_utils.py
```diff
@@ -110,6 +114,66 @@ def display_feedback_call(
...
+        # Only keep rows from the most recent eval_root_id for groups that have more than one row
+        indices_to_keep = []
+        for group_name, group_df in grouped:
+            if len(group_df) > 1:
+                # Multiple rows with same args_span_id and args_span_attribute
+                # Find the most recent eval_root_id based on timestamp
+                eval_root_max_timestamps = group_df.groupby("eval_root_id")[
+                    "timestamp"
+                ].max()
+                most_recent_eval_root_id = eval_root_max_timestamps.idxmax()
```
sfc-gh-dkurokawa: this is weird syntax. Like you do an idxmax on the max? I imagine your tests verified this is correct, but this seems odd to me.

src/dashboard/trulens/dashboard/utils/records_utils.py
```diff
@@ -110,6 +114,66 @@ def display_feedback_call(
...
+
+                # Keep all rows from the most recent eval_root_id
+                recent_rows = group_df[
+                    group_df["eval_root_id"] == most_recent_eval_root_id
```
sfc-gh-dkurokawa, line 166: you just grouped by this on L157, if you just store that you can probably get that group slightly faster.

src/dashboard/trulens/dashboard/utils/records_utils.py
```diff
@@ -77,6 +77,10 @@ def display_feedback_call(
         df["meta"] = pd.Series([call[i]["meta"] for i in range(len(call))])
         df = df.join(df.meta.apply(lambda m: pd.Series(m))).drop(columns="meta")
 
+        # Filter to only show the latest results for calls with the same args_span_id and args_span_attribute
+        # NOTE: this is an optimization to not show duplicate calls in the UI (e.g. when a feedback function is recomputed on the same metric name and inputs)
+        df = _filter_duplicate_span_calls(df)
```
sfc-gh-dkurokawa, line 84: aren't the columns you added showing up in the `df` now though? Obviously there's an argument that they should be there, but I feel like it may be confusing for the user. Like `eval_root_id` seems like something strange to show up for the user to me. I'll leave it to you to decide if you want to keep it all or w/e though.

tests/unit/test_records_utils.py
```diff
@@ -0,0 +1,247 @@
...
+
+        result = _filter_duplicate_span_calls(test_data)
+
+        # Should keep only the row from the most recent eval_root_id
+        self.assertEqual(len(result), 1)
+        self.assertEqual(result["eval_root_id"].iloc[0], "eval_2")
+        self.assertEqual(result["other_data"].iloc[0], "string_row_2")
+
+
+if __name__ == "__main__":
```
sfc-gh-dkurokawa: you don't need this.

src/core/trulens/core/database/sqlalchemy.py
```diff
@@ -1179,13 +1179,18 @@ def _get_records_and_feedback_otel(
...
-                            kwargs = {
-                                key[len(kwargs_prefix) :]: value
-                                for key, value in record_attributes.items()
-                                if key.startswith(kwargs_prefix)
-                            }
+                            # Extract namespaced attributes using the helper method
+                            kwargs = self._extract_namespaced_attributes(
+                                record_attributes, SpanAttributes.CALL.KWARGS
+                            )
+                            args_span_id = self._extract_namespaced_attributes(
```
sfc-gh-dkurokawa: this won't work because on L1179 we're checking if this is an eval span (not an eval root span). I think in general you want to be de-duping at the eval root span level (and not eval span level) so the changes you made on this line and below in this file should be changed to operate on the eval root level.

src/core/trulens/core/database/sqlalchemy.py
sfc-gh-nvytla: start timestamp

src/core/trulens/core/database/sqlalchemy.py
sfc-gh-nvytla: Removed

tests/unit/test_records_utils.py
sfc-gh-nvytla: Removed

src/dashboard/trulens/dashboard/utils/records_utils.py
sfc-gh-nvytla: Fixed

src/dashboard/trulens/dashboard/utils/records_utils.py
sfc-gh-nvytla: Updated

src/dashboard/trulens/dashboard/utils/records_utils.py
sfc-gh-nvytla: Will add logic to remove these columns from df before displaying to the user

src/core/trulens/core/database/sqlalchemy.py
sfc-gh-nvytla: Noted and fixed!

src/core/trulens/core/database/sqlalchemy.py
```diff
@@ -1173,25 +1173,49 @@ def _get_records_and_feedback_otel(
...
+
+                            call_data = {
+                                "eval_root_id": record_attributes.get(
+                                    SpanAttributes.EVAL.EVAL_ROOT_ID
+                                ),
+                                "timestamp": record_data["ts"],
+                                "args_span_id": args_span_id,
+                                "args_span_attribute": args_span_attribute,
+                            }
+                            feedback_result["calls"].append(call_data)
```
sfc-gh-dkurokawa, line 1194: by adding this here, won't this show up when you select the feedback pill?

src/dashboard/trulens/dashboard/utils/records_utils.py
```diff
@@ -112,6 +116,83 @@ def display_feedback_call(
         st.warning("No feedback details found.")
 
 
+def _drop_metadata_columns(
```
sfc-gh-dkurokawa: I think you can just say `df.drop(columns=columns_to_drop, errors="ignore")` or something like that.

src/core/trulens/core/database/sqlalchemy.py
sfc-gh-nvytla: Yes, there would be another row for the eval_root span that would show up if you select the feedback pill, would you prefer it not to? I don't think it poses an issue per se.

src/dashboard/trulens/dashboard/utils/records_utils.py
sfc-gh-nvytla: Fixed.

src/core/trulens/core/database/sqlalchemy.py
sfc-gh-nvytla: Specifically:
1. Currently, filtering occurs based on data extracted from the "calls" column.
2. `args_span_id` and `args_span_attribute` are only available in `EVAL_ROOT` spans, and given that the filtering uses these, we rely on the `EVAL_ROOT` calls to be populated in the "calls" column.

Issue #1987, opened 2025-05-19
[BUG]


nileshteji:
**Bug Description**
Getting this error , tried creating 3 different eval systems on llama index 
`Endpoint OpenAIEndpoint request failed 4 time(s): 
	'cached_property' object has no attribute 'completions'
	'cached_property' object has no attribute 'completions'
	'cached_property' object has no attribute 'completions'
	'cached_property' object has no attribute 'completions'
`

**To Reproduce**
`from trulens.core import TruSession

session = TruSession()
session.reset_database()

import urllib.request

url = "https://raw.githubusercontent.com/run-llama/llama_index/main/docs/docs/examples/data/paul_graham/paul_graham_essay.txt"
file_path = "sample_data/paul_graham_essay.txt"

if not os.path.exists("sample_data"):
    os.makedirs("sample_data")

if not os.path.exists(file_path):
    urllib.request.urlretrieve(url, file_path)

from llama_index.core import Settings
from llama_index.core import SimpleDirectoryReader
from llama_index.core import VectorStoreIndex
from llama_index.llms.openai import OpenAI

Settings.chunk_size = 128
Settings.chunk_overlap = 16
Settings.llm = OpenAI()

documents = SimpleDirectoryReader("sample_data").load_data()
index = VectorStoreIndex.from_documents(documents)

query_engine = index.as_query_engine(similarity_top_k=3)

response = query_engine.query("What did the author do growing up?")
print(response)

import numpy as np
from trulens.apps.llamaindex import TruLlama
from trulens.core import Feedback
from trulens.providers.openai import OpenAI

# Initialize provider class
provider = OpenAI(model_engine="gpt-4.1-mini")

# select context to be used in feedback. the location of context is app specific.

context = TruLlama.select_context(query_engine)

# Define a groundedness feedback function
f_groundedness = (
    Feedback(
        provider.groundedness_measure_with_cot_reasons, name="Groundedness"
    )
    .on(context.collect())  # collect context chunks into a list
    .on_output()
)

# Question/answer relevance between overall question and answer.
f_answer_relevance = Feedback(
    provider.relevance_with_cot_reasons, name="Answer Relevance"
).on_input_output()
# Question/statement relevance between question and each context chunk.
f_context_relevance = (
    Feedback(
        provider.context_relevance_with_cot_reasons, name="Context Relevance"
    )
    .on_input()
    .on(context)
    .aggregate(np.mean)
)

[...]

PR #2005, opened 2025-05-26
Fix OpenAI provider in python 3.12+ to work with the changed @cached_property behavior

sfc-gh-dhuang:
# Description
Under Python 3.12+, our OpenAIClient wrapper stopped honouring the `@cached_property` on `openai.OpenAI.chat`, so code like `self.endpoint.client.chat.completions.create(...)` in TruLens OpenAI provider was broken as it would be returning the raw `cached_property` descriptor instead of a Chat instance and hence throwing errors like:

```
RuntimeError: Endpoint OpenAIEndpoint request failed 4 time(s): 
	'cached_property' object has no attribute 'completions'
	'cached_property' object has no attribute 'completions'
	'cached_property' object has no attribute 'completions'
	'cached_property' object has no attribute 'completions'
```
like the issue raised by our users: 
https://github.com/truera/trulens/issues/1987

**Root cause:**

Python 3.12 change: the stdlib `functools.cached_property` no longer carries the undocumented per-property lock, altering how its descriptor is recognized by static attribute lookups. see https://github.com/celery/kombu/issues/1804

Our `safe_getattr` used `inspect.getattr_static` + `isinstance(..., property)` to detect and invoke only plain `@property` objects, so `cached_property` descriptors were treated as regular attributes and never invoked.

**Fix:** 
When any of those is found (and `get_prop=True`), call its `__get__(obj, type(obj))` so that the value is computed, cached on the instance, and returned.

Non-descriptor attributes remain untouched.

## Type of change

  not work as expected)
- [ ] New Tests

----

> [!IMPORTANT]
> Fix `safe_getattr` in `pyschema.py` to handle `@cached_property` in Python 3.12+, with tests added in `test_pyschema.py`.
> 
>   - **Behavior**:
>     - Fixes `safe_getattr` in `pyschema.py` to handle `@cached_property` descriptors by invoking `__get__` method.
>     - Returns computed and cached value for `@property` and `@cached_property`.
>     - Raises `ValueError` if `get_prop=False` for properties/descriptors.
>     - Returns error dict on exceptions during property access.
>   - **Tests**:
>     - Adds `SafeGetAttrTests` in `test_pyschema.py` to test `safe_getattr` with plain attributes, properties, and `cached_property`.
>     - Includes test for `cached_property` behavior in Python 3.12+.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=truera%2Ftrulens&utm_source=github&utm_medium=referral)<sup> for 7de92d9ff08b781f875ce36c10654823ad4e8d1d. You can [customize](https://app.ellipsis.dev/truera/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


sfc-gh-jreini:
Can you add a test for this?

sfc-gh-dhuang:
@sfc-gh-jreini  added ut as requested. this is also tested manually in different virtual envs (python 3.10, 3.11, and 3.12)

PR #2043, opened 2025-06-02
Have eval spans hold all direct function calls.

sfc-gh-dkurokawa:
# Description
Have eval spans hold all direct function calls.

I've changed the UI to show this better when clicking on the feedback pill as well.

## Type of change
  not work as expected)
- [ ] New Tests


src/core/trulens/core/database/sqlalchemy.py
```diff
@@ -1151,6 +1151,7 @@ def _get_records_and_feedback_otel(
                             }
 
                         # Update feedback result
+                        # TODO(otel): This isn't going to work if there are multiple with the same name.
```
sfc-gh-nvytla, line 1154: To my understanding, this would overwrite a feedback if it already exists on the frontend. The DB should be preserved. 

What would you prefer otherwise? I know we started exploring this in https://github.com/truera/trulens/pull/2003.

src/feedback/trulens/feedback/computer.py
```diff
@@ -636,92 +639,106 @@ def _call_feedback_function(
...
+    Args:
+        span: Span to set attributes on.
+        metadata: Metadata to extract attributes from.
+    """
+    for k, v in metadata.items():
+        set_span_attribute_safely(
+            span, f"{SpanAttributes.EVAL.METADATA}.{k}", v
+        )
+        if k in ["explanation", "explanations", "reason", "reasons"]:
+            set_span_attribute_safely(span, SpanAttributes.EVAL.EXPLANATION, v)
```
sfc-gh-nvytla: Can we define and use a dict or enum here?

tests/unit/test_otel_get_records_and_feedback.py
```diff
@@ -92,9 +92,9 @@ def setUp(self):
 
         # Constants for static eval test
         self.STATIC_FEEDBACK_NAME = "answer_relevance"
-        # NOTE: there are 3 total calls for the feedback column, see:
-        # test_otel_spans/{eval_root_span.json, eval_span_1.json, eval_span_2.json}
-        self.STATIC_NUM_CALLS = 3
+        # NOTE: there are 2 total calls for the feedback column, see:
+        # test_otel_spans/{eval_span_1.json, eval_span_2.json}
+        self.STATIC_NUM_CALLS = 2
```
sfc-gh-nvytla, line 97: To confirm, are we no longer returning the eval_root span? Does this mean we also lose access to all of the attributes from the eval_root span (score, higher_is_better, explanation, etc.)?

src/feedback/trulens/feedback/computer.py
```diff
@@ -636,92 +639,106 @@ def _call_feedback_function(
...
+    ) as eval_span:
+        set_general_span_attributes(eval_span, SpanAttributes.SpanType.EVAL)
+        res = None
+        exc = None
+        try:
+            res = feedback_function(**kwargs)
+            metadata = {}
+            if isinstance(res, tuple):
+                if (
+                    len(res) != 2
```
sfc-gh-nvytla, line 707: Nit: mind adding a comment here to clarify what 2 means?

tests/unit/test_otel_get_records_and_feedback.py
sfc-gh-nvytla: Nvm, this makes sense given that the "calls" column should only contain eval spans and not eval_root spans.

tests/unit/test_otel_get_records_and_feedback.py
```diff
@@ -976,10 +976,9 @@ def test_get_records_and_feedback_otel_kwargs_extraction(self):
 
         # Verify the call data structure
         self.assertIn("args", call_data)
-        self.assertIn("kwargs", call_data["args"])
 
         # This is the key test - verify that kwargs were properly extracted
-        extracted_kwargs = call_data["args"]["kwargs"]
+        extracted_kwargs = call_data["args"]
```
sfc-gh-nvytla, line 981: If we're no longer relying on the kwargs key here, we also remove the kwargs extraction logic from sqlalchemy.py?

tests/unit/test_otel_get_records_and_feedback.py
sfc-gh-dkurokawa: well we're no longer returning it in the calls. It's information is used outside the "feedback_calls" though. That said, I was wondering if you knew if we use it anywhere else?

tests/unit/test_otel_get_records_and_feedback.py
sfc-gh-dkurokawa: sorry where is that?

src/core/trulens/core/database/sqlalchemy.py
sfc-gh-dkurokawa: yeah so say we have three "answer relevance" metrics, then we want to show all three as separate pills I think with some way to distinguish between them. That's a question for another day though since that we'll have to come up with some good design for that.

src/feedback/trulens/feedback/computer.py
sfc-gh-dkurokawa: sorry what do you mean?

sfc-gh-nvytla: For `["explanation", "explanations", "reason", "reasons"]`

tests/unit/test_otel_get_records_and_feedback.py
sfc-gh-nvytla: Hmm, not off the top of my head.

tests/unit/test_otel_get_records_and_feedback.py
sfc-gh-nvytla: Lines 1182-1188 of this PR: https://github.com/truera/trulens/pull/2043/files#diff-8c95ff0b4a4c59089b8bb7dfea464d67d723bda1b0e50a2a2ad016b703598790R1182-R1188

sfc-gh-nvytla (approved):
LGTM overall! Left a few small comments, non-blocking.

tests/unit/test_otel_get_records_and_feedback.py
sfc-gh-dkurokawa: oh I did change that part of the code. I just put it in `args` directly now.

PR #2054, opened 2025-06-06
[Nit] Sans serif font fix

sfc-gh-nvytla:
# Description

- Make sans serif load properly

## Type of change

  not work as expected)
- [ ] New Tests

----

> [!IMPORTANT]
> Fixes font loading issue in `run_dashboard()` by correcting sans serif font specification in `run.py`.
> 
>   - **Behavior**:
>     - Fixes font loading issue by changing `--theme.font="sans-serif"` to `--theme.font=sans-serif` in `run_dashboard()` in `run.py`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=truera%2Ftrulens&utm_source=github&utm_medium=referral)<sup> for d7192e12844157ff413add98fd72d104088ee4b6. You can [customize](https://app.ellipsis.dev/truera/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/dashboard/trulens/dashboard/run.py
+++ b/src/dashboard/trulens/dashboard/run.py
@@ -106,7 +106,7 @@ def run_dashboard(
         "--server.headless=True",
         "--theme.base=dark",
         "--theme.primaryColor=#E0735C",
-        "--theme.font=sans serif",
+        "--theme.font=sans-serif",
     ]
     if _watch_changes:
         args.extend([
--- a/src/dashboard/trulens/dashboard/run.py
+++ b/src/dashboard/trulens/dashboard/run.py
@@ -106,7 +106,7 @@ def run_dashboard(
         "--server.headless=True",
         "--theme.base=dark",
         "--theme.primaryColor=#E0735C",
-        '--theme.font="sans-serif"',
+        "--theme.font=sans serif",
     ]
     if _watch_changes:
         args.extend([


sfc-gh-nvytla:
@sfc-gh-chu do you mind testing if Sans Serif loads properly with this PR?

src/dashboard/trulens/dashboard/run.py
```diff
@@ -106,7 +106,7 @@ def run_dashboard(
         "--server.headless=True",
         "--theme.base=dark",
         "--theme.primaryColor=#E0735C",
-        '--theme.font="sans-serif"',
```
sfc-gh-chu, line 109: Did some testing and this seems to be tied to a recent streamlit change version.
 - streamlit<1.44: expects "sans serif"
 - streamlit>=1.44: expects "sans-serif" but will accept "sans serif" for backwards compatibility

sfc-gh-chu (approved):

src/dashboard/trulens/dashboard/run.py
sfc-gh-nvytla: Hmm this is interesting, do you mind linking the Streamlit PR change that this is related to if it's not too much trouble? When @sfc-gh-gtokernliang and I tested this change locally on Streamlit version 1.45.0, the existing syntax (`"sans-serif"`) did not load properly.

sfc-gh-nvytla: That being said, I just tested the following, and both of them work on Streamlit version >=1.45.1 (so maybe the reason it wasn't working was because of the single/double quote syntax?). 

@sfc-gh-gtokernliang which would you prefer for compatibility?

```python
# Without dash
"--theme.font=sans serif"

# With dash
"--theme.font=sans-serif"
```

sfc-gh-gtokernliang: not too sure how python dependency management works here - if all users for this path will definitely be using streamlit >= 1.45.1 moving forward, I think we should use the latest syntax, which looks like it should be with dash?

also makes it clearer that it's one word and properly parsed as a cli option too :)

sfc-gh-gtokernliang (approved):

PR #2055, opened 2025-06-09
If we try to get `app` from a `TruApp` and it doesn't yet exist, don't go into an infinite loop.

sfc-gh-dkurokawa:
# Description
If we try to get `app` from a `TruApp` and it doesn't yet exist, don't go into an infinite loop.

## Type of change
  not work as expected)
- [ ] New Tests

### DIFF
--- a/src/core/trulens/core/app.py
+++ b/src/core/trulens/core/app.py
@@ -1604,9 +1604,10 @@ you use the `%s` wrapper to make sure `%s` does get instrumented. `%s` method
         except AttributeError:
             pass
 
-        app = self.app
+        if __name == "app":
+            return None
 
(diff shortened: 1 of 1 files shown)


sfc-gh-nvytla (approved):
Any tests to add here?

sfc-gh-dhuang (approved):

sfc-gh-dkurokawa:
> Any tests to add here?

Oh this is actually to fix some weird test behavior --- specifically during tests we often go into these infinite loops. Even saying that I would add a test that tests this exactly, but the issue kinda is hard to test since all that would happen before is that we'd go into some infinite loop during GC and it would just silently crash.

PR #2058, opened 2025-06-10
Add tooltips for diff, change variance to standard deviation

sfc-gh-nvytla:
# Description

- Add tooltip to explain diff (for 2 versions)
- Change variance to standard deviation (update calculation and add tooltip) (for >2 versions)

- [SNOW-2126681](https://snowflakecomputing.atlassian.net/browse/SNOW-2126681), [SNOW-2158287](https://snowflakecomputing.atlassian.net/browse/SNOW-2158287)

## Type of change

  not work as expected)
- [ ] New Tests

----

> [!IMPORTANT]
> Add tooltips for 'Diff' and 'Variance' in `Compare.py`, using 'Diff' for 2 versions and 'Std Dev' for 3+ versions.
> 
>   - **Tooltips**:
>     - Add tooltips for 'Diff' and 'Variance' in `_build_grid_options` and `_render_grid` based on `num_comparators`.
>     - For 2 versions, use 'Diff' (absolute difference); for 3+ versions, use 'Variance' (standard deviation).
>   - **Functions**:
>     - Update `_render_shared_records` to use 'Std Dev' instead of 'Variance' for 3+ versions.
>     - Modify `_render_all_app_feedback_plot` to calculate 'Diff' or 'Variance' based on the number of versions.
>   - **Misc**:
>     - Change 'Variance' to 'Std Dev' in column names and calculations in `Compare.py`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=truera%2Ftrulens&utm_source=github&utm_medium=referral)<sup> for 7e0ee2776af759f6d7f85e8bc3df06fd1f3d1244. You can [customize](https://app.ellipsis.dev/truera/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/dashboard/trulens/dashboard/pages/Compare.py
+++ b/src/dashboard/trulens/dashboard/pages/Compare.py
@@ -361,7 +361,6 @@ def _build_grid_options(
         agg_diff_col,
         cellClassRules=diff_cell_rules,
         filter="agNumberColumnFilter",
-        tooltipField=agg_diff_col,
         headerTooltip=agg_tooltip,
     )
     gb.configure_columns(
@@ -443,7 +442,7 @@ def _render_grid(
             help=diff_help, format="%.3f"
         )
 
-    column_order = ["input", *diff_cols, agg_diff_col]
+    column_order = ["input", agg_diff_col, *diff_cols]
     column_order = [col for col in column_order if col in df.columns]
     event = st.dataframe(
         df[column_order],


src/dashboard/trulens/dashboard/pages/Compare.py
```diff
@@ -347,15 +348,27 @@ def _build_grid_options(
         hide=True,
         filter="agSetColumnFilter",
     )
+
+    # Determine tooltip text based on number of comparators
+    if num_comparators == 2:
+        agg_tooltip = "Mean Diff: Average of absolute differences between the two app versions"
+        diff_tooltip = "Diff: Absolute difference between the two app versions (|version1 - version2|)"
+    else:
```
sfc-gh-jreini: Will this tooltip display if just one app version?

src/dashboard/trulens/dashboard/pages/Compare.py
```diff
@@ -347,15 +348,27 @@ def _build_grid_options(
...
+        agg_tooltip = f"Mean Variance: Average variance across all {num_comparators} app versions"
+        diff_tooltip = f"Variance: Statistical variance of values across all {num_comparators} app versions"
```
sfc-gh-jreini: Maybe would better to show how many standard deviations a given app version is away from mean

src/dashboard/trulens/dashboard/pages/Compare.py
sfc-gh-nvytla: There will be no diff or variance if there is only one app version. I believe we provide a warning to the user that there are not enough app versions to compare instead of rendering any tables.

That being said, I'll update the `else` to an `elif num_comparators > 2` to be more precise.

src/dashboard/trulens/dashboard/pages/Compare.py
sfc-gh-nvytla: This is a tooltip for the column name, so it's not being set dynamically per cell. I'm not sure if per-cell tooltips are officially supported for Streamlit Aggrid, many of the implementations I see online call for allowing unsafe JavaScript execution.

If you are looking for per-cell tooltips, I'm happy to track that in a separate ticket.

sfc-gh-dkurokawa: I agree with @sfc-gh-jreini , but I'm not sure I follow your response @sfc-gh-nvytla . What do you mean? I think Josh is basically suggesting instead of showing the variance we show the standard deviation since that has the same "units" as the actual value itself so is easier to intuit.

sfc-gh-nvytla: @sfc-gh-dkurokawa @sfc-gh-jreini ah gotcha. Just to be clear, z-score (# of standard deviations from the mean) is a per-value metric, so we would have different values per app version (whereas with variance, it's a metric across all versions). See below for what this would look like, and let me know if this is ideal.


Screenshot 2025-06-16 at 2 24 09 PM

sfc-gh-jreini: Thanks for the follow-up. Looking at it like this, z-score seems like too much. I think the middle-ground here is that we report the standard deviation across all app versions. Users can mentally infer z-score given the standard deviation - and as dave put the stdev is a bit nicer than variance bc it's in the same unit as the score itself

sfc-gh-nvytla: @sfc-gh-jreini, sounds good, I've updated the code to use standard deviation instead. I'll update the PR description as well!

Screenshot 2025-06-16 at 3 22 56 PM

sfc-gh-nvytla: Mind re-reviewing when you get a few minutes?

PR #2061, opened 2025-06-11
Add `on_trace` selector and show custom trajectory eval

sfc-gh-jreini:
# Description

Add `on_trace` selector and show use via custom trajectory eval

Screenshot 2025-06-11 at 6 32 10 PM

## Type of change

  not work as expected)
- [ ] New Tests

----

> [!IMPORTANT]
> Add `on_trace` selector to `Feedback` for custom trajectory evaluation using trace-based selection.
> 
>   - **New Feature**:
>     - Add `on_trace` method to `Feedback` in `feedback.py` for trace-based selection.
>     - Add `select_trace` static method to `Selector` in `selector.py` to match all spans and return summaries.
>   - **Behavior**:
>     - `on_trace` raises `RuntimeError` if OTel tracing is not enabled.
>     - `select_trace` collects span summaries into a list if `collect_list=True`.
>   - **Feedback Computation**:
>     - Update `_run_feedback_on_inputs` in `computer.py` to handle grouped span inputs for feedback computation.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=truera%2Ftrulens&utm_source=github&utm_medium=referral)<sup> for ec5829571e64f2f7daf7cb91cf447e971ef821c5. You can [customize](https://app.ellipsis.dev/truera/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/core/trulens/core/feedback/feedback.py
+++ b/src/core/trulens/core/feedback/feedback.py
@@ -609,17 +609,21 @@ class Feedback(feedback_schema.FeedbackDefinition):
         ret = self.model_copy()
         ret.selectors = new_selectors
         return ret
-    
-    def on_trace(self, arg: Optional[str] = "trace"):
+
+    def on_trace(
+        self, arg: Optional[str] = "trace", *, collect_list: bool = True
+    ):
         """
         The selector will return a list of dicts as expected by select_trace.
         """
         from trulens.core.otel.utils import is_otel_tracing_enabled
+
         if not is_otel_tracing_enabled():
             raise RuntimeError("on_trace is only supported in OTel mode.")
         from trulens.core.feedback.selector import Selector
+
         new_selectors = self.selectors.copy()
-        new_selectors[arg] = Selector.select_trace()
+        new_selectors[arg] = Selector.select_trace(collect_list=collect_list)
         ret = self.model_copy()
         ret.selectors = new_selectors
         return ret
@@ -1352,6 +1356,7 @@ Feedback function signature:
                 app=app, record=record, source_data=source_data
             )
         )
+
 
 class SnowflakeFeedback(Feedback):
     """[DEPRECATED] Similar to the parent class Feedback except this ensures the feedback is run only on the Snowflake server.
--- a/src/core/trulens/core/feedback/feedback.py
+++ b/src/core/trulens/core/feedback/feedback.py
@@ -609,6 +609,20 @@ class Feedback(feedback_schema.FeedbackDefinition):
         ret = self.model_copy()
         ret.selectors = new_selectors
         return ret
+    
+    def on_trace(self, arg: Optional[str] = "trace"):
+        """
+        The selector will return a list of dicts as expected by select_trace.
+        """
+        from trulens.core.otel.utils import is_otel_tracing_enabled
+        if not is_otel_tracing_enabled():
+            raise RuntimeError("on_trace is only supported in OTel mode.")
+        from trulens.core.feedback.selector import Selector
+        new_selectors = self.selectors.copy()
+        new_selectors[arg] = Selector.select_trace()
+        ret = self.model_copy()
+        ret.selectors = new_selectors
+        return ret
 
     def on(self, *args, **kwargs) -> Feedback:
         """
@@ -1338,7 +1352,6 @@ Feedback function signature:
                 app=app, record=record, source_data=source_data
             )
         )
-
 
 class SnowflakeFeedback(Feedback):
     """[DEPRECATED] Similar to the parent class Feedback except this ensures the feedback is run only on the Snowflake server.
--- a/src/core/trulens/core/feedback/feedback.py
+++ b/src/core/trulens/core/feedback/feedback.py
@@ -620,8 +620,6 @@ class Feedback(feedback_schema.FeedbackDefinition):
 
         if not is_otel_tracing_enabled():
             raise RuntimeError("on_trace is only supported in OTel mode.")
-        from trulens.core.feedback.selector import Selector
-
         new_selectors = self.selectors.copy()
         new_selectors[arg] = Selector.select_trace(collect_list=collect_list)
         ret = self.model_copy()
--- a/src/core/trulens/core/feedback/selector.py
+++ b/src/core/trulens/core/feedback/selector.py
@@ -48,7 +48,9 @@ class Selector:
     ):
         # Only require at least one selector if not using a processor that handles all spans
         if (
-            function_name is None and span_name is None and span_type is None
+            function_name is None
+            and span_name is None
+            and span_type is None
             and span_attributes_processor is None
         ):
             raise ValueError(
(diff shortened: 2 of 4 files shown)


src/feedback/trulens/feedback/computer.py
```diff
@@ -542,6 +542,38 @@ def _run_feedback_on_inputs(
     Returns:
         Number of feedbacks computed.
     """
+    # If there is only one kwarg and all inputs are for that kwarg, and collect_list is True,
+    # aggregate all values and call feedback function ONCE.
+    if flattened_inputs:
+        first_inputs = flattened_inputs[0][2]
```
sfc-gh-nvytla, line 546: How would this work when the `span_group` is an optional param within the Tuple? Wouldn't this point to null/None if `span_group` was not provided?

sfc-gh-jreini: I updated a bit to better handle span grouping - can you take a second look?

In general, I think span group None is also treated as a span group. New logic should handle this, and result will be that we collect the list and run the feedback function ONCE per span group

src/core/trulens/core/feedback/selector.py
```diff
@@ -203,3 +209,40 @@ def context_retrieval_processor(
...
+                return ""
+            # Extract function name, input, output if available
+            name = attrs.get(SpanAttributes.CALL.FUNCTION, "<unknown>")
+            output_val = attrs.get(
+                SpanAttributes.CALL.RETURN, attrs.get("output", "?")
+            )
+            return f"{name}: {output_val}"
+
+        return Selector(
+            span_type=None,
```
sfc-gh-nvytla, line 237: Is there a reason this is None? Do we not want to populate it with the appropriate span type?

src/core/trulens/core/feedback/selector.py
```diff
@@ -203,3 +209,40 @@ def context_retrieval_processor(
...
+            span_attributes_processor=trace_processor,
+            collect_list=collect_list,
+            match_only_if_no_ancestor_matched=False,
```
sfc-gh-nvytla, line 240: Just curious, when will this be True?

src/core/trulens/core/feedback/selector.py
sfc-gh-jreini: yeah - this is selecting a list of multiple span types. not sure it makes sense to set a single span type on that?

src/core/trulens/core/feedback/selector.py
sfc-gh-jreini: It's true for select_context bc retrieval is often in trees and we want to avoid selecting redundant context from both parent/child.

src/core/trulens/core/feedback/feedback.py
```diff
@@ -610,6 +610,24 @@ def on_context(
...
+        self, arg: Optional[str] = "trace", *, collect_list: bool = True
+    ):
+        """
+        The selector will return a list of dicts as expected by select_trace.
+        """
+        from trulens.core.otel.utils import is_otel_tracing_enabled
+
+        if not is_otel_tracing_enabled():
+            raise RuntimeError("on_trace is only supported in OTel mode.")
+        from trulens.core.feedback.selector import Selector
```
sfc-gh-dkurokawa: don't need to import anything.

src/core/trulens/core/feedback/selector.py
```diff
@@ -203,3 +209,40 @@ def context_retrieval_processor(
             collect_list=collect_list,
             match_only_if_no_ancestor_matched=True,
         )
+
+    @staticmethod
+    def select_trace(collect_list: bool = True) -> Selector:
```
sfc-gh-dkurokawa, line 214: So this is a bit confusing since `collect_list` has been used in other places to mean that it'll take a list and then basically run the feedback function on that. If set to `False` it means it'll run the feedback function once per entry in the list.

Here, this would mean that we're going to create a list by taking everything across all spans. So the meaning is different.

Also, would there ever be a case where `collect_list` would be `False`?

src/core/trulens/core/feedback/selector.py
```diff
@@ -203,3 +209,40 @@ def context_retrieval_processor(
...
+        """
+        Returns a Selector that matches all spans, returning a string summary for each non-root, non-eval span.
+        The system will collect these into a list (if collect_list=True), with each string describing a step: function name, input, and output.
+        """
+        from trulens.otel.semconv.trace import SpanAttributes
+
+        def trace_processor(attrs: dict) -> str:
+            if "record_attributes" in attrs:
```
sfc-gh-dkurokawa: shouldn't we handle the case where it's not in `attrs`?

src/core/trulens/core/feedback/selector.py
```diff
@@ -203,3 +209,40 @@ def context_retrieval_processor(
...
+                attrs = attrs["record_attributes"]
+            if not isinstance(attrs, dict):
+                return ""
+            span_type = attrs.get(SpanAttributes.SPAN_TYPE)
+            if span_type in (
+                SpanAttributes.SpanType.RECORD_ROOT,
+                SpanAttributes.SpanType.EVAL,
+                SpanAttributes.SpanType.EVAL_ROOT,
+                "record_root",
```
sfc-gh-dkurokawa: why are these here? These are the same as the ones above.

src/core/trulens/core/feedback/selector.py
```diff
@@ -203,3 +209,40 @@ def context_retrieval_processor(
...
+                "eval",
+                "eval_root",
+            ):
+                return ""
+            # Extract function name, input, output if available
+            name = attrs.get(SpanAttributes.CALL.FUNCTION, "<unknown>")
+            output_val = attrs.get(
+                SpanAttributes.CALL.RETURN, attrs.get("output", "?")
+            )
+            return f"{name}: {output_val}"
```
sfc-gh-dkurokawa: why don't you want the record root?

src/feedback/trulens/feedback/computer.py
```diff
@@ -542,6 +542,37 @@ def _run_feedback_on_inputs(
     Returns:
         Number of feedbacks computed.
     """
+    if flattened_inputs:
+        first_inputs = flattened_inputs[0][2]
+        if (
+            len(first_inputs) == 1
+            and hasattr(list(first_inputs.values())[0], "collect_list")
```
sfc-gh-dkurokawa, line 549: wait when would this be true?

sfc-gh-dkurokawa (commented):
I'm not sure you wanted a review on this since it was just for your demo purposes, but I reviewed it anyway haha.

src/core/trulens/core/feedback/selector.py
sfc-gh-jreini: can remove

src/core/trulens/core/feedback/selector.py
sfc-gh-jreini: Meaning seems analagous in my mind (with the exception that we're working across spans).

regardless of where data is coming from, collect list just tells us how to treat the data. if it's True, we lump in all of the data together and eval it once (whether that's data from different spans or same span). If it's false, we run evals on each piece of data (could be from different spans or same span).

It'd be useful to set collect_list to False for "progress" style evals where we want to assess something like goal completion on each step in the trace.

src/core/trulens/core/feedback/selector.py
sfc-gh-jreini: Record root is redundant to input of first intermediate step and output of last, and could confuse a trajectory eval

src/core/trulens/core/feedback/selector.py
sfc-gh-jreini: This logic was unneccessary - I removed it

src/feedback/trulens/feedback/computer.py
sfc-gh-jreini: This is true when the eval only takes one argument and sets collect_list

sfc-gh-jreini:
Tabling this PR for now as @sfc-gh-dkurokawa has an alternate format for selecting the trace.

PR #2062, opened 2025-06-11
agent evals (experimental)

sfc-gh-jreini:
# Description

and context.

## Type of change

  not work as expected)
- [ ] New Tests

### DIFF
--- a/src/feedback/trulens/feedback/prompts.py
+++ b/src/feedback/trulens/feedback/prompts.py
@@ -147,3 +147,19 @@ GENERATE_KEY_POINTS_USER_PROMPT = """
 COMPREHENSIVENESS_SYSTEM_PROMPT = v2.Comprehensiveness.system_prompt
 
 COMPREHENSIVENESS_USER_PROMPT = v2.Comprehensiveness.user_prompt
+
+TRAJECTORY_EVAL_SYSTEM_PROMPT = """
+Use the following rubric to evaluate the execution trace of the system:
+0: Agents made wrong or unnecessary calls. Critical steps were skipped or repeated without purpose. Generated outputs were off-topic, hallucinated, or contradictory. Confusion between agent roles. User goal was not meaningfully addressed.
+1: Several unnecessary or misordered agent/tool use. Some factual errors or under-specified steps. Redundant or partially irrelevant tool calls. Weak or ambiguous agent outputs at one or more steps
+2: Some minor inefficiencies or unclear transitions. Moments of stalled progress, but ultimately resolved. The agents mostly fulfilled their roles, and the conversation mostly fulfilled answering the query.
+3: Agent handoffs were well-timed and logical. Tool calls were necessary, sufficient, and accurate. No redundancies, missteps, or dead ends. Progress toward the user query was smooth and continuous. No hallucination or incorrect outputs
+"""
+
+GOAL_COMPLETENESS_SYSTEM_PROMPT = """
+Use the following rubric to evaluate if the final output fully meets the user's goal:
+0: The output does not address the user's goal at all, is off-topic, or is factually incorrect. Major requirements are missing or misunderstood.
+1: The output partially addresses the goal but misses key requirements, contains significant errors, or is incomplete in important ways.
+2: The output mostly fulfills the goal, with only minor omissions or inaccuracies. All major requirements are met, but some details could be improved.
+3: The output fully and accurately meets the user's goal. All requirements are satisfied, the answer is correct, complete, and well-presented.
+"""
--- a/src/feedback/trulens/feedback/llm_provider.py
+++ b/src/feedback/trulens/feedback/llm_provider.py
@@ -15,6 +15,9 @@ from trulens.feedback import generated as feedback_generated
 from trulens.feedback import prompts as feedback_prompts
 from trulens.feedback.v2 import feedback as feedback_v2
 
+from .prompts import GOAL_COMPLETENESS_SYSTEM_PROMPT
+from .prompts import TRAJECTORY_EVAL_SYSTEM_PROMPT
+
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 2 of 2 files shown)


src/feedback/trulens/feedback/prompts.py
```diff
@@ -147,3 +147,26 @@
...
+
+GOAL_COMPLETENESS_SYSTEM_PROMPT = """
+Use the following rubric to evaluate if the final output fully meets the user's goal:
+0 — Completely Inadequate: The output is off-topic, factually incorrect, or fails to address the user's goal in any meaningful way. Major requirements are missing or misunderstood.
+
+1 — Partially Complete: The output addresses the goal in a limited way but contains significant errors, omissions, or misunderstandings. Key requirements are unmet.
+
+2 — Mostly Complete: The output meets all major requirements with only minor issues. It may contain small inaccuracies or incomplete details, but the goal is largely fulfilled.
+
+3 — Fully Complete: The output is correct, comprehensive, and clearly presented. It satisfies all aspects of the user's goal without errors or omissions.
```
sfc-gh-nvytla, line 171: Out of curiosity: when we propose judging the "final output" here, does this take into consideration conversations where the user has multiple goals? Is it possible that a user may propose say, 3 goals, accomplish 2 in intermediate steps, and accomplish the 3rd in the final step, but that the prompt may score the goal completeness as 1/3 because the final output (e.g. last message/artifact generated by the LLM) only covers 1 of the 3 originally proposed goals?

I'm also curious about if the user's goal changes midway through the conversation -- can we pick up on that intent? If so, do we consider splitting the conversation based on some sort of semantic topic layer and running separate evals?

I understand there's no silver bullet here, but would love to hear your thoughts on how we might handle something like this in the future!

examples/experimental/otel/langgraph_quickstart_otel.ipynb
```diff
@@ -425,7 +431,12 @@
     "    multi_agent_graph,\n",
     "    app_name=\"Multi-Agent Chart Generation\",\n",
     "    app_version=\"Base\",\n",
-    "    feedbacks=[f_answer_relevance, f_context_relevance, f_groundedness],\n",
+    "    feedbacks=[\n",
+    "        f_context_relevance,\n",
+    "        f_groundedness,\n",
+    "        f_goal_completion,\n",
+    "        f_trajectory,\n",
+    "    ],\n",
```
sfc-gh-nvytla, line 439: Is there a reason we remove answer_relevance here? My guess is that goal_completion is intended to be a more holistic evaluation for agentic conversations compared to single-turn completions, is that fair?

src/feedback/trulens/feedback/prompts.py
sfc-gh-jreini: The intermediate steps in most cases would be hidden to the user (except through the trace!), so I think it's reasonable to only grade the final output.

In the multi-turn case, perhaps we should think about ways we can combine output across the conversation.

Example:

Human: What noise do dogs make and what's the price of a bag of Chewy dog food?
AI: Dogs bark. What size bag?
Human: 5 pounds.
AI: A 5 pound bag of chewy dog food is $20.

In this example, I think we could ideally handle this by:
1. Evaluate quality of follow-up questions, e.g. is the bag size the right question to ask and neccessary to answer the question? We *could* do this today by having some feedback function that assesses if a follow-up question is needed, etc.
2. Evaluate output quality by combining the conversation history. Right now we view each q/a pair independently. We do have access to conversation history, but our evals today don't have a meaningful way to use them, e.g. for answer relevance, you might want to combine "Dogs bark. What size bag?" and "A 5 pound bag of chewy dog food is $20." One way we could think about handling this would be to add optional fields into our feedback functions that take conversation history, that if provided is used to provide context to the eval?

examples/experimental/otel/langgraph_quickstart_otel.ipynb
sfc-gh-jreini: This is still single turn, but the goal completion is intended to capture a broader scope than answer relevance - tapping into its completeness and a broader sense of task completion as well.

src/feedback/trulens/feedback/prompts.py
sfc-gh-nvytla: That's fair, perhaps we need some sort of per-turn intent/goal tracking/lineage such that you can understand how a user's goals dynamically change through a conversation.

1. Fully agree on needing to account for some level of underspecification. So much of user queries in the wild are wholly exploratory, and upon reading the first response from an LLM they ask follow-up questions (or perhaps we suggest in the same way Perplexity et al. provide follow-up prompt suggestions).
2. Maybe there's some room here as well for retrieval-augmented intent prediction? At the same time, I wonder if tacking on conversation history for particularly complex queries could also lead to unnecessary redundancy or (separately) user overload, and asking a model to go too deep too fast in terms of trying to predict could increase the potential surface area for answer irrelevance.

Would the following be graded differently from the example you provided?

```
Human: What noise do dogs make? (**initial goal**)
AI: Dogs bark. 
Human: What's the price of a bag of Chewy dog food?
What size bag?
Human: 5 pounds.
AI: A 5 pound bag of chewy dog food is $20. (**final output**)
```

sfc-gh-jreini: Yeah this example is way cleaner:

Human: What noise do dogs make? (**initial goal**)
AI: Dogs bark. -> receives 3/3 grade for goal completion
Human: What's the price of a bag of Chewy dog food?
What size bag? -> receives low to middle grade for goal completion, perhaps a high score for follow-up questions eval
Human: 5 pounds. 
AI: A 5 pound bag of chewy dog food is $20. (**final output**) -> receives 3/3 grade for goal completion.

Right now we don't have natural way to group conversations, so the intent splitting needs to be done by the developer. This might be cleanest, bc they can note that the chewy dog food question is different and use that as a point to start the conversation history fresh.

sfc-gh-jreini:
Tabling this PR for now as @sfc-gh-dkurokawa has an alternate format for selecting the trace.

PR #2075, opened 2025-06-19
Allow for trace level metrics.

sfc-gh-dkurokawa:
# Description
Allow for trace level metrics.

## Type of change
  not work as expected)
- [ ] New Tests

### DIFF
--- a/src/core/trulens/core/feedback/selector.py
+++ b/src/core/trulens/core/feedback/selector.py
@@ -3,13 +3,95 @@ from __future__ import annotations
 from dataclasses import dataclass
 from typing import Any, Callable, Dict, List, Optional
 
+import pandas as pd
 from trulens.core.feedback.feedback_function_input import FeedbackFunctionInput
 from trulens.otel.semconv.trace import SpanAttributes
 
 
 @dataclass
+class ProcessedContentNode:
+    event: pd.Series
+    content: Any
+    parent: Optional[ProcessedContentNode]
+    children: List[ProcessedContentNode]
+
+
+@dataclass
+class Trace:
+    # Filtered events in the trace that match the selector. That is, Each row
+    # corresponds to a `trulens.core.schema.event.Event` along with an
+    # additional column `processed_content` that contains the processed content
+    # of the event described by the following fields of the selector:
+    # 1. span_attributes_processor
+    # 2. span_attribute
+    # 3. function_attribute
+    # Use this over `processed_content_roots` if the hierarchy of the events
+    # (i.e. parent-child relationship) is not important.
+    events: Optional[pd.DataFrame]
+
+    # This describes the DAG of the filtered events in the trace.
+    # Example:
+    # If we have three events pre-filtering A, B, and C where:
+    # 1. A is the parent of B and E.
+    # 2. B is the parent of C.
+    # 3. C is the parent of D.
+    # 4. B, E, and D are kept and A and C are filtered out.
+    # Then `processed_content_roots` will contain two events B and E where B has
+    # a single child D.
+    # Pre-filtering:
+    #       A
+    #      / \
+    #     B   E
+    #     |
+    #     C
+    #     |
+    #     D
+    # Post-filtering:
+    #     B    E
+    #     |
+    #     D
+    # Use this over `events` if the hierarchy of the events (i.e. parent-child
+    # relationship) is important.
+    processed_content_roots: List[ProcessedContentNode]
+
+    def __init__(self):
+        self.events = None
+        self.processed_content_roots = []
+
+    def add_event(
+        self,
+        processed_content: Any,
+        event: pd.Series,
+        parent_processed_content_node: Optional[ProcessedContentNode],
+    ) -> ProcessedContentNode:
+        if self.events is None:
+            self.events = pd.DataFrame(event).transpose()
+        else:
+            self.events = pd.concat(
+                [self.events, pd.DataFrame(event).transpose()],
+                ignore_index=True,
+            )
+        node = ProcessedContentNode(
+            event=event,
+            content=processed_content,
+            parent=parent_processed_content_node,
+            children=[],
+        )
+        if parent_processed_content_node is not None:
+            parent_processed_content_node.children.append(node)
+        else:
+            self.processed_content_roots.append(node)
+        return node
+
+
+@dataclass
 class Selector:
     # The following fields describe how to select the spans.
+    # If `trace_level` is True, then this `Selector` doesn't describe a single
+    # span but any number of spans in the trace. It will describe all spans
+    # unless further filtered by `function_name`, `span_name`, or `span_type`.
+    # If `trace_level` is False, then this `Selector` describes a single span.
+    trace_level: bool = False
     function_name: Optional[str] = None
     span_name: Optional[str] = None
     span_type: Optional[str] = None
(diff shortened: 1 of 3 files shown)


src/core/trulens/core/feedback/selector.py
```diff
@@ -74,7 +165,8 @@ def __init__(
 
     def describes_same_spans(self, other: Selector) -> bool:
         return (
-            self.function_name == other.function_name
+            self.trace_level == other.trace_level
+            and self.function_name == other.function_name
             and self.span_name == other.span_name
             and self.span_type == other.span_type
             and self.match_only_if_no_ancestor_matched
```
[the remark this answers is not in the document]
sfc-gh-dkurokawa: this is literally in the next line

src/feedback/trulens/feedback/computer.py
```diff
@@ -236,18 +237,105 @@ def _collect_inputs_from_events(
...
+    """DFS collect inputs from events based on a single trace level `Selector`.
+
+    Args:
+        sole_kwarg: Sole kwarg to the feedback function.
+        sole_selector: Selector for sole kwarg to the feedback function.
+        span_id_to_child_events: Mapping from span id to child events.
+        curr_event: Current event to process.
+        ret:
+            Mapping from (record_id, None) to kwarg group to inputs. This is
+            what will be updated throughout the DFS.
```
sfc-gh-nvytla: Nit: add missing description comment for `parent_processed_content_node`

src/feedback/trulens/feedback/computer.py
```diff
@@ -236,18 +237,105 @@ def _collect_inputs_from_events(
...
+    """
+    # Convenience variables.
```
sfc-gh-nvytla, line 313: Just curious, what do you mean by convenience here?

sfc-gh-nvytla: Nvm, I see this used as well for `_dfs_collect_inputs_from_events()`

src/feedback/trulens/feedback/computer.py
```diff
@@ -236,18 +237,105 @@ def _collect_inputs_from_events(
...
+    curr_processed_content_node = None
+    if sole_selector.matches_span(span_name, record_attributes):
+        processed_content = sole_selector.process_span(
+            span_id, record_attributes
+        ).value
+        curr_processed_content_node = ret[(record_id, None)][(sole_kwarg,)][0][
+            sole_kwarg
+        ].value.add_event(
+            processed_content, curr_event, parent_processed_content_node
+        )
```
sfc-gh-nvytla, line 328: Just to confirm I understand this code, this line is updating the `ret` data structure by adding event information to the trace-level `FeedbackFunctionInput` object associated with the current record and keyword argument. 

Specifically, if the current span matches the selector's criteria, it calls the `add_event` method on the `Trace` value held by the feedback input, passing the processed content, the current event, and a reference to the parent processed content node, and stores the result in `curr_processed_content_node` for further processing in the DFS traversal.

Is that correct?

src/feedback/trulens/feedback/computer.py
sfc-gh-nvytla, line 311: Also just to confirm, we are using `None` in `(record_id, None)` as a distinguishing factor for trace-level operations compared to span-level operations, is that correct? (because there is no meaningful span_group at the trace level compared to the more granular span level).

src/feedback/trulens/feedback/computer.py
```diff
@@ -236,18 +237,105 @@ def _collect_inputs_from_events(
...
+    if any(curr.trace_level for curr in kwarg_to_selector.values()):
+        if len(kwarg_to_selector) != 1:
+            raise ValueError(
+                "Cannot have multiple `Selectors` if any are trace level!"
+            )
+        return True
+    return False
+
+
+def _dfs_collect_trace_level_inputs_from_events(
```
sfc-gh-nvytla, line 289: Do we handle some sort of circular reference/cyclic graphs from the span event tree (e.g. if span incorrectly lists itself or one of its ancestors as a child)?

src/core/trulens/core/feedback/selector.py
```diff
@@ -3,13 +3,95 @@
...
+
 @dataclass
 class Selector:
     # The following fields describe how to select the spans.
+    # If `trace_level` is True, then this `Selector` doesn't describe a single
+    # span but any number of spans in the trace. It will describe all spans
+    # unless further filtered by `function_name`, `span_name`, or `span_type`.
+    # If `trace_level` is False, then this `Selector` describes a single span.
+    trace_level: bool = False
     function_name: Optional[str] = None
```
sfc-gh-jreini, line 100: Possible to make the trace filterable by lists of function names, span names or span types?

And are these AND filters or OR filters? Is this configurable?

src/core/trulens/core/feedback/selector.py
```diff
@@ -61,6 +151,7 @@ def __init__(
             raise ValueError(
                 "Must specify exactly one of `span_attributes_processor`, `span_attribute`, or `function_attribute`!"
```
sfc-gh-jreini, line 158: In many cases we might just want the full trace. In those cases, wouldn't we just want to do `.on({"trace": Selector(trace_level=True)})` and not have any processor or attribute specification?

src/feedback/trulens/feedback/computer.py
sfc-gh-dkurokawa: yes, except this part:
"and stores the result in curr_processed_content_node for further processing in the DFS traversal."

The DFS traverses all the spans, it's just when we call `add_event` we're adding it to the filtered events in the `Trace` object in `Trace::events` and also in the filtered DAG in `Trace::processed_content_roots`.

Incidentally I realize this is all very complicated looking haha, so if you have any good suggestions to make it simpler I'll gladly incorporate them.

src/feedback/trulens/feedback/computer.py
sfc-gh-dkurokawa: yeah I thought about it and I think we should ignore span groups since otherwise you can have a lot of weird stuff happening.

src/feedback/trulens/feedback/computer.py
sfc-gh-dkurokawa: no but I don't think we need to worry about that since we generally don't set the span and parent span ids ourselves (and it's done via the OTel library). That said if you feel strongly about it I can add it in since you're right, it's theoretically possible someone can do that by accident if they were writing it themselves for w/e reason.

src/core/trulens/core/feedback/selector.py
sfc-gh-dkurokawa: Yes to the first question. My comment is supposed to make that clear but I guess not haha (specifically the part that says "It will describe all spans unless further filtered by `function_name`, `span_name`, or `span_type`."

AND to the second question.

I clearly am not the best at conveying these haha. If you have suggestions on the comment I'll gladly take it.

sfc-gh-nvytla (approved):

src/core/trulens/core/feedback/selector.py
sfc-gh-jreini: In this case, type hints should allow lists, e.g. `Optional[List[str]]`? And I think it should be OR within a field. Here's a possibly better explanation:

When trace_level=True, this selector matches all spans, optionally filtered by function_name, span_name, and span_type.
Each filter field accepts a list of values and is treated as an OR within that field (e.g., function_name=["a", "b"] matches spans with function name "a" or "b").
Filters across fields are combined with AND logic (i.e., span must match all specified criteria).

sfc-gh-dkurokawa: oh wait sorry I misunderstood your question. I thought you meant like you can have both function name and span name. Not just one of them. You can't have like a list of function names. Does that change anything?

src/core/trulens/core/feedback/selector.py
sfc-gh-dkurokawa: Changed to allow this.

PR #2118, opened 2025-07-10
Run feedback function on everything in event table initially.

sfc-gh-dkurokawa:
# Description
Run feedback function on everything in event table initially.

I originally didn't do this since it may be expensive, but I think this is causing confusion and it's unlikely it's that expensive for our initial users anyway. Probably want to figure out something long term later though.

## Type of change
  not work as expected)
- [ ] New Tests

### DIFF
--- a/src/core/trulens/core/utils/evaluator.py
+++ b/src/core/trulens/core/utils/evaluator.py
@@ -18,6 +18,8 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger(__name__)
 
+_PROCESSED_TIME_DELTA = datetime.timedelta(hours=1)
+
 
 class Evaluator:
     def __init__(self, app: App):
@@ -28,7 +30,7 @@ class Evaluator:
         self._stop_event = threading.Event()
         self._compute_feedbacks_lock = threading.Lock()
         self._record_id_to_event_count = pd.Series(dtype=int)
-        self._start_time = datetime.datetime.now()
+        self._processed_time = datetime.datetime.now() - _PROCESSED_TIME_DELTA
 
     def _events_under_record_root(self, events: pd.DataFrame) -> pd.DataFrame:
         """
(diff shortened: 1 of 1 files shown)


src/core/trulens/core/utils/evaluator.py
```diff
@@ -138,6 +143,7 @@ def _compute_feedbacks(
                     TruSession().force_flush()
                 if in_evaluator_thread and self._stop_event.is_set():
                     break
+        self._processed_time = datetime.datetime.now() - _PROCESSED_TIME_DELTA
```
sfc-gh-nvytla: I'm confused, why are we subtracting an hour from the current time?

sfc-gh-nvytla: Non-blocking: Wondering if we should make this lower e.g. on the order of single-digit minutes instead to reduce potential for redundant processing. Then again, we could adjust easily in the future after getting user feedback. Thoughts?

sfc-gh-nvytla: One suggestion to make debugging easier: add logging for `_processed_time`, either:
- Log `_processed_time` before and after each feedback computation to make event selection traceable, or
- Log retrieved event time ranges for each cycle to clarify which events are being processed each time.

sfc-gh-dkurokawa: sorry I made more comments in the description now to explain the logic. Does that help?

sfc-gh-nvytla (approved):
lgtm after resolving some of the PR comments! 👍

src/core/trulens/core/utils/evaluator.py
```diff
@@ -138,6 +144,7 @@ def _compute_feedbacks(
                     TruSession().force_flush()
                 if in_evaluator_thread and self._stop_event.is_set():
                     break
+        self._processed_time = new_processed_time - _PROCESSED_TIME_DELTA
```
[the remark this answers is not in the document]
sfc-gh-dkurokawa: this is a bad idea since that would update the time for everything.

PR #2131, opened 2025-07-17
Handle when the spans don't have app ids in the UI.

sfc-gh-dkurokawa:
# Description
Handle when the spans don't have app ids in the UI.

If people use our semantic conventions to write their own spans, then we can't expect them to put in app ids that correspond exactly to our app ids that are functions of the app name and version. Thus, in such cases, we handle this by computing the app ids for them.

In the future we should move to not using app ids because of this or having some better way of getting the app ids.

## Type of change
  not work as expected)
- [ ] New Tests

----

> [!IMPORTANT]
> Handle spans without app IDs by computing them in `sqlalchemy.py`, ensuring proper event filtering and logging discrepancies.
> 
>   - **Behavior**:
>     - Handle spans without app IDs by computing app IDs in `_get_paginated_record_ids_otel` and `_get_records_and_feedback_otel` in `sqlalchemy.py`.
>     - Modify conditions to include spans with `None` or empty app IDs.
>     - Log a message when computed app ID is not in the provided app IDs list.
>   - **Misc**:
>     - Add comments about potential pagination issues and future improvements regarding app ID handling.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=truera%2Ftrulens&utm_source=github&utm_medium=referral)<sup> for 1cf78cca14c7fd902b20f23ac5b4a587c4a479c6. You can [customize](https://app.ellipsis.dev/truera/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


src/core/trulens/core/database/sqlalchemy.py
```diff
@@ -983,7 +983,13 @@ def _get_paginated_record_ids_otel(
             app_id_expr = self._json_extract_otel(
                 "resource_attributes", ResourceAttributes.APP_ID
             )
-            conditions.append(app_id_expr.in_(app_ids))
+            conditions.append(
+                sa.or_(
+                    app_id_expr.in_(app_ids),
+                    app_id_expr.is_(None),
```
sfc-gh-nvytla, line 989: The PR description mentions that we'll handle null app_ids by computing them ourselves -- mind adding a test to verify this is the case?

src/core/trulens/core/database/sqlalchemy.py
```diff
@@ -1082,6 +1088,10 @@ def _get_records_and_feedback_otel(
                         app_name, app_version
                     ),
                 )
+                if app_ids and app_id not in app_ids:
```
sfc-gh-nvytla, line 1091: Can we add some logging here?

src/core/trulens/core/database/sqlalchemy.py
sfc-gh-nvytla: Oops, please ignore, it's been a while since I wrote this logic. 😅

src/core/trulens/core/database/sqlalchemy.py
sfc-gh-dkurokawa: Also I need to add a comment about deprecation of this approach long term.

PR #2200, opened 2025-08-28
[Small fix] Properly set pypi token in make file

sfc-gh-dhuang:
# Description

Make variable vs environment variable: Make doesn't automatically import environment variables as Make variables. The `$(TOKEN)` syntax looks for a Make variable, not an environment variable

## Type of change

  not work as expected)
- [ ] New Tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,8 @@ CONDA_BUILD_DIRS := $(shell find . \
 	-name "*meta.yaml" \
 	-exec dirname {} \;)
 LAST_TRULENS_EVAL_COMMIT := 4cadb05 # commit that includes the last pre-namespace trulens_eval package
+TOKEN ?= $(shell echo $$TOKEN)
+
 
 # Global setting: execute all commands of a target in a single shell session.
 # Note for MAC OS, the default make is too old to support this. "brew install


Makefile
```diff
@@ -17,6 +17,8 @@ CONDA_BUILD_DIRS := $(shell find . \
 	-name "*meta.yaml" \
 	-exec dirname {} \;)
 LAST_TRULENS_EVAL_COMMIT := 4cadb05 # commit that includes the last pre-namespace trulens_eval package
+TOKEN ?= $(shell echo $$TOKEN)
```
sfc-gh-nvytla, line 20: Why not something like this?

`TOKEN ?= $(TOKEN)`

sfc-gh-dhuang: ah cuz then it'd be an no-op.


 I'm defining Make variable `TOKEN` by reading in the shell env variable, which later below can then be accessed by `$(TOKEN)`  (parenthesized Make variable) below for commands like `make upload-all`

sfc-gh-nvytla: Oops, I misread this, you're right! I was wondering if we could get around using the shell command, but it seems pretty straightforward for now.

sfc-gh-nvytla (approved):

Makefile
sfc-gh-dkurokawa, line 20: wait I'm confused. This whole time uploading to pypi has been working for me though.

sfc-gh-dhuang: ohh interesting. it's never been working for me for like forever, and I have always had to manually paste the token. 

perhaps it's the different shell we are using? I'm using `zsh`

sfc-gh-dkurokawa: Oh I use `gmake` instead of `make`. That might be it.

PR #2219, opened 2025-09-10
Use Feedback class instead of underlying function when running evals

sfc-gh-jreini:
# Description

Feedback class was not being directly run in otel recording, instead the underlying function was being called via feedback.imp. This change uses the Feedback class call so that we use all of the available parameters, such as criteria and custom instructions.

## Type of change

  not work as expected)
- [ ] New Tests

### DIFF
--- a/src/core/trulens/core/app.py
+++ b/src/core/trulens/core/app.py
@@ -2123,7 +2123,7 @@ you use the `%s` wrapper to make sure `%s` does get instrumented. `%s` method
             compute_feedback_by_span_group(
                 events,
                 feedback.name,
-                feedback,  # Pass the entire Feedback object instead of just feedback.imp
+                feedback,
                 feedback.higher_is_better,
                 feedback.selectors,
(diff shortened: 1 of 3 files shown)


src/core/trulens/core/app.py
```diff
@@ -2123,7 +2123,7 @@ def compute_feedbacks(
             compute_feedback_by_span_group(
                 events,
                 feedback.name,
-                feedback.imp,
+                feedback,
```
sfc-gh-nvytla, line 2125: Hmm, if we're passing in the full feedback anyways, why not replace L2127-2129 (`higher_is_better`, `selectors`, `aggregator`) and instead grab them from the feedback and propagate those values down as well)? Also valid for #2118.

sfc-gh-jreini: added this

sfc-gh-ajia (approved):
lgtm!

PR #2225, opened 2025-09-15
trace compression

sfc-gh-jreini:
# Description

Trace compression.

Removed from trace:
❌ Full page contents (truncated to ~200 chars)
❌ Debug/log spans (unless they contain errors)
❌ Duplicate information
❌ Verbose logging data

## Type of change

  not work as expected)
- [ ] New Tests

### DIFF
--- a/src/feedback/trulens/feedback/llm_provider.py
+++ b/src/feedback/trulens/feedback/llm_provider.py
@@ -2580,7 +2580,7 @@ class LLMProvider(core_provider.Provider):
         )
 
         if isinstance(trace, Trace):
-            trace = trace.events.to_json(default_handler=str)
+            trace = trace.to_compressed_json(verbose=True, default_handler=str)
         elif isinstance(trace, str):
             trace = trace
         else:
@@ -2654,7 +2654,7 @@ class LLMProvider(core_provider.Provider):
         )
 
         if isinstance(trace, Trace):
-            trace = trace.events.to_json(default_handler=str)
+            trace = trace.to_compressed_json(verbose=True, default_handler=str)
         elif isinstance(trace, str):
             trace = trace
         else:
@@ -2728,7 +2728,7 @@ class LLMProvider(core_provider.Provider):
         )
 
         if isinstance(trace, Trace):
-            trace = trace.events.to_json(default_handler=str)
+            trace = trace.to_compressed_json(verbose=True, default_handler=str)
         elif isinstance(trace, str):
             trace = trace
         else:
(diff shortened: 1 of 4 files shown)


src/feedback/trulens/feedback/llm_provider.py
```diff
@@ -2580,7 +2580,7 @@ def logical_consistency_with_cot_reasons(
         )
 
         if isinstance(trace, Trace):
-            trace = trace.events.to_json(default_handler=str)
+            trace = trace.to_compressed_json(default_handler=str)
```
sfc-gh-dhuang: [non-blocking] for future configurability and path to optimization, might be worth making it a config flag to use / or not use the compressed version of traces, given that we prob don't have yet a good idea on their impacts on eval performance

sfc-gh-dhuang: Actually, you mind making this a config in this PR and default to NOT using compressed format, if time allows? My concern after reading the existing heuristics  is that it might be too extreme in its removal process.

src/core/trulens/core/utils/trace_compression.py
```diff
@@ -0,0 +1,1162 @@
...
+                for skip in ["debug", "log", "trace", "monitor"]
+            ):
+                # Check if it has an error or important info
+                has_error = False
+                if "span_attributes" in span:
+                    attrs = span.get("span_attributes", {})
+                    if isinstance(attrs, dict):
+                        if any(
+                            key in attrs
+                            for key in ["error", "exception", "failure"]
```
sfc-gh-dhuang, line 283: (fully aware this might be vibe-coded) I'm a bit concerned this is too strict / rigid in the sense that it's not always considering case-insensitivity and pattern-matching seems to be limited to exact match.

src/core/trulens/core/utils/trace_compression.py
```diff
@@ -0,0 +1,1162 @@
+"""
+Trace compression utilities to reduce token usage in feedback functions.
+This module provides functionality to compress trace data while preserving
+essential information needed for evaluation.
+"""
```
sfc-gh-nvytla, line 5: Can we add a comment mentioning that this approach is experimental and perhaps should be used with caution?

sfc-gh-dhuang: +1

sfc-gh-jreini: added this

src/core/trulens/core/utils/trace_compression.py
sfc-gh-jreini: Agree with you - I'm going to merge as is to avoid further changing since it's working now and we're on a time crunch, but we should address this in a future update

PR #2263, opened 2025-10-14
Fix bugs with runs and metric computation

sfc-gh-jreini:
# Description

- fix type error from run.py when computing metrics
- fix issue where app creation failed when using same app_name due to incorrect snowpark query plan
- fix issue with computing metrics and filtering on correct record_attributes

## Type of change

  not work as expected)
- [ ] New Tests

----

> [!IMPORTANT]
> Fixes bugs in metric computation and app creation by addressing type errors, Snowpark query issues, and record attribute filtering.
> 
>   - **Behavior**:
>     - Fix type error in `run.py` during metric computation.
>     - Fix app creation failure in `sql_utils.py` due to incorrect Snowpark query plan for `SHOW` commands.
>     - Ensure correct filtering on `record_attributes` in `snowflake_event_table_db.py`.
>   - **Functions**:
>     - Add `_parse_json_field` in `run.py` for consistent JSON parsing.
>     - Modify `_compute_client_side_metrics_from_configs` in `run.py` to flush spans before querying.
>   - **Misc**:
>     - Update `api.trulens.3.11.yaml` to reflect version changes from 2.2.4 to 2.2.0.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=truera%2Ftrulens&utm_source=github&utm_medium=referral)<sup> for 0bee9209e980bf5d816fd786e65619d37e2df47f. You can [customize](https://app.ellipsis.dev/truera/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/apps/langchain/meta.yaml
+++ b/src/apps/langchain/meta.yaml
@@ -20,7 +20,7 @@ build:
 
 requirements:
   host:
-    - python >=3.9
+    - python >=3.9,<3.13
     - poetry-core <2.0.0
     - pip
   run:
--- a/src/apps/langgraph/meta.yaml
+++ b/src/apps/langgraph/meta.yaml
@@ -20,7 +20,7 @@ build:
 
 requirements:
   host:
-    - python >=3.9
+    - python >=3.9,<3.13
     - poetry-core <2.0.0
     - pip
   run:
--- a/src/connectors/snowflake/meta.yaml
+++ b/src/connectors/snowflake/meta.yaml
@@ -22,7 +22,7 @@ build:
 
 requirements:
   host:
-    - python >=3.9
+    - python >=3.9,<3.13
     - poetry-core <2.0.0
     - pip
   run:
--- a/src/core/meta.yaml
+++ b/src/core/meta.yaml
@@ -22,7 +22,7 @@ build:
 
 requirements:
   host:
-    - python >=3.9
+    - python >=3.9,<3.13
     - poetry-core <2.0.0
     - pip
   run:
--- a/src/dashboard/meta.yaml
+++ b/src/dashboard/meta.yaml
@@ -22,7 +22,7 @@ build:
 
 requirements:
   host:
-    - python >=3.9
+    - python >=3.9,<3.13
     - poetry-core <2.0.0
     - pip
   run:
--- a/src/feedback/meta.yaml
+++ b/src/feedback/meta.yaml
@@ -22,7 +22,7 @@ build:
 
 requirements:
   host:
-    - python >=3.9
+    - python >=3.9,<3.13
     - poetry-core <2.0.0
     - pip
   run:
--- a/src/otel/semconv/meta.yaml
+++ b/src/otel/semconv/meta.yaml
@@ -22,7 +22,7 @@ build:
 
 requirements:
   host:
-    - python >=3.9
+    - python >=3.9,<3.13
     - poetry-core <2.0.0
     - pip
   run:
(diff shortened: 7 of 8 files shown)


src/core/trulens/core/run.py
```diff
@@ -1125,19 +1125,79 @@ def _get_events_for_client_metrics(self) -> pd.DataFrame:
...
+                        trace = events_df.at[idx, "trace"]
+                        record = events_df.at[idx, "record"]
+                        record_attributes = (
+                            events_df.at[idx, "record_attributes"]
+                            if "record_attributes" in events_df.columns
+                            else {}
+                        )
+
+                        # Ensure trace is a dict, not a string
+                        if isinstance(trace, str):
```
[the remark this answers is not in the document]
sfc-gh-dhuang: perhaps take this suggestion

src/connectors/snowflake/trulens/connectors/snowflake/dao/sql_utils.py
```diff
@@ -52,7 +52,19 @@ def execute_query(
     Executes a query with optional parameters with qmark parameter binding (if applicable).
     """
     try:
-        df = session.sql(query, params=parameters).to_pandas()
+        # For SHOW commands, Snowpark has a bug where plan attributes don't match actual results
```
sfc-gh-dhuang, line 55: wait what does this comment mean here?


I discovered the issue just today after this change: https://github.com/truera/trulens/pull/2251/files, 
but the issue is not what's being documented here. 


What happens is `to_pandas` mostly works only for `SELECT` statement, and it doesn't work for some of the system functions, like `SHOW EXTERNAL AGENT`

src/connectors/snowflake/trulens/connectors/snowflake/dao/sql_utils.py
```diff
@@ -52,7 +52,19 @@ def execute_query(
...
+        # Use collect() and manually construct DataFrame to avoid schema mismatch
+        if query.strip().upper().startswith("SHOW"):
```
sfc-gh-dhuang, line 57: for the short-term fix,  we can prob instead of explicitly checking for `"SHOW"` (which is brittle), let's check for the present of `"SELECT"` and only call `to_pandas()` when it's a SELECT query, and manually construct DF otherwise

cc @sfc-gh-dkurokawa

sfc-gh-dhuang (changes requested):

src/connectors/snowflake/trulens/connectors/snowflake/dao/sql_utils.py
sfc-gh-jreini: AFAICT, the issue is:

1. SHOW returns the data without attached schema
2. To_pandas tries to (incorrectly) infer the schema because SHOW does not provide the needed schema info (in SELECT queries, the schema would come from the query plan)

We can either:
1. Use SELECT as you suggest
2. Rebuild the dataframe resulting from the SHOW command

sfc-gh-jreini: If you want to take this in a separate PR, I can remove it from this one @sfc-gh-dhuang - what do you prefer?

sfc-gh-dhuang: makes sense. 

Yeah I think it's good to leave it as is in your PR (with the `Rebuild the dataframe resulting from the SHOW command` ). 

I / Dave will take it in a separate PR for a longer-term fix

sfc-gh-dhuang (approved):

PR #2464, opened 2026-05-11
feat: GEPA (Genetic Prompt Optimization) Integration

### LINKED ISSUE #2421: GEPA (Genetic Prompt Optimization) Integration
## Summary
GEPA uses genetic/evolutionary algorithms to optimize prompts. Like DSPy, it needs a fitness function to score prompt variants — TruLens feedback functions are a natural fit for this. An integration would let users evolve prompts toward higher groundedness, lower toxicity, or any custom TruLens metric.

## What
Create a lightweight adapter that wraps TruLens feedback functions as GEPA fitness functions:

- `TruLensFitness(feedback_fn)` — adapts a TruLens feedback function to GEPA's expected callable interface
- A cookbook showing: define a base prompt, use TruLens context_relevance as the fitness function, run GEPA evolution, visualize the improvement trajectory in the TruLens dashboard
- Log each generation's best prompt + score as a TruLens record for audit trail

## Difficulty
Medium

bashward:
# Description

This should close #2421 . This PR adds `trulens-apps-gepa`, a new lightweight integration package that adapts TruLens feedback functions for use as fitness functions in GEPA (Genetic/Evolutionary Prompt Adaptation), a family of approaches that use evolutionary algorithms to iteratively optimize prompts.

TruLens feedback functions are a natural fit for GEPA fitness functions: they already score dimensions like context relevance, groundedness, and toxicity on a normalized `[0, 1]` scale, require no extra setup beyond what TruLens users already have, and produce auditable results. This integration closes the loop between prompt optimization and observability.

What’s included:

* `TruLensFitness` - a thin adapter that wraps any TruLens feedback callable into the `(prompt, **kwargs) -> float` callable interface expected by GEPA fitness functions.

  * Supports optional `context_key` / `context` parameters for metrics like `context_relevance` that require a reference document.
  * When a `TruVirtual` recorder is provided, each prompt evaluation is logged as a virtual TruLens record, creating a full generation-by-generation audit trail visible in the TruLens dashboard.

* `run_evolution` - a simple `(μ+λ)` evolutionary optimization loop:

  * scores candidates with the fitness function,
  * keeps the top-k candidates via elitist selection,
  * mutates survivors to fill the next population,
  * repeats for a configurable number of generations.

  The helper returns:

  * the best prompt,
  * its score,
  * and per-generation history.

  The function accepts any callable as a fitness function, not just `TruLensFitness`.

* Cookbook notebook:

  * `examples/expositional/frameworks/gepa/gepa_trulens_integration.ipynb`

  The notebook demonstrates:

  * defining a base prompt,
  * wrapping `context_relevance` as a fitness function,
  * running evolutionary optimization,
  * visualizing the improvement trajectory,
  * and viewing all evaluations in the TruLens dashboard.

* The package lives at `src/apps/gepa/` following the same structure as `trulens-apps-langchain`, `trulens-apps-llamaindex`, and related integration packages.

* The package depends only on `trulens-core`; no optional packages are required.

* All modules are importable without additional dependencies, so `test_import_base` in the static test suite passes without changes to `optional_mods`.

* `TruVirtual` / `VirtualRecord` imports are deferred inside `_log_record` (lazy import), so the recorder path is only activated when a recorder is actually passed.

* The `run_evolution` helper is intentionally minimal and designed to demonstrate the integration pattern without over-engineering the optimization layer.

* Users with more advanced mutation strategies (including LLM-based prompt rewriting) can pass any `Callable[[str], str]` as `mutate_fn`.

* `trulens-apps-gepa` has been added to the apps dependency group in the root `pyproject.toml` so it is installed and discoverable in CI alongside the other app integrations.

## Type of change

* [ ] Bug fix (non-breaking change which fixes an issue)
* [x] New feature (non-breaking change which adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to
  not work as expected)
* [x] New Tests
* [ ] This change includes re-generated golden test results
* [x] This change requires a documentation update


examples/expositional/frameworks/gepa/gepa_trulens_integration.ipynb
```diff
@@ -0,0 +1,303 @@
+{
```
sfc-gh-jreini, line 1: <pre class="ql-syntax" spellcheck="false">Line #10.        recorder=recorder,
</pre><p>what would you think about having <code>TruLensFitness</code> just record by default and add <code>app_name</code> &amp; <code>app_version</code> as parameters to the class?</p><p> </p><p>additional nit, a more trulens standard naming convention would be <code>TruGEPA</code></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/truera/trulens/pull/2464/discussion/#comment-3221639933'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

bashward: Hey, I like the direction having TruGEPA with app_name/app_version is the right convention and makes the API much cleaner for TruLens users who are already familiar with TruChain and TruLlama.

One thing: rather than creating the TruVirtual recorder lazily inside a try/except, I'd create it eagerly in __init__ the same way TruChain does. The lazy approach means if there's no active TruSession, the error gets swallowed silently and the user just loses all their logging with no clear indication of why. Failing loudly at construction time is much easier to debug.

So the final shape I'd go with:


fitness = TruGEPA(
    provider.context_relevance,
    input_key="question",
    context_key="context",
    context=REFERENCE_CONTEXT,
    app_name="gepa_optimizer",
    app_version="v1",
)
With TruVirtual created in __init__, recording is on by default, the naming follows the Tru* convention, and a missing session raises an error immediately rather than silently. What do you think?

sfc-gh-jreini: This makes sense to me - can you make the app_name/app_version optional and if they do not supply either, we do not log the evolutions? If they supply only app_name or only app_version (not both), also raise an error immediately that says something like "For logging to be enabled, you must supply both app_name and app_version; to disable logging, supply neither."

bashward: yeah i updated the logic as you described, updated the PR

src/apps/gepa/trulens/apps/gepa/fitness.py
```diff
@@ -0,0 +1,210 @@
...
+            supplied together with ``app_version``; omit both to disable logging.
+        app_version: Version string of the virtual app. Must be supplied
+            together with ``app_name``; omit both to disable logging.
+    """
+
+    def __init__(
+        self,
+        feedback_fn: Callable,
+        *,
+        input_key: str = "prompt",
```
joshreini1: Hey @bashward — this PR is looking great, nice work!

One last change I'd like before we merge: **replace `input_key`/`context_key`/`context` with `optimize_key` + `feedback_args`**.

The current API hardcodes support for at most two feedback function kwargs, which limits which feedback functions users can plug in. With two more general params this works with any signature:

- **`optimize_key: str`** — names which feedback arg the evolving candidate string maps to
- **`feedback_args: Optional[Dict[str, Any]]`** — all other args forwarded to the feedback function unchanged

```python
# Before
fitness = TruGEPA(
    provider.context_relevance,
    input_key="question",
    context_key="context",
    context=REFERENCE_CONTEXT,
)

# After
fitness = TruGEPA(
    provider.context_relevance,
    optimize_key="question",
    feedback_args={"context": REFERENCE_CONTEXT},
)
```

This way `TruGEPA` works naturally with `groundedness(source, statement)`, `coherence(text)`, custom fns with 3+ args — all without needing dedicated constructor params for each one.

Once this is in, I think we're good to go!

bashward: yeah, that makes sense, will update the PR

sfc-gh-jreini (approved):
Thanks for adding this @bashward !! This will be a huge add for anyone looking to get more visibility into their GEPA-based optimizations, or who wants to get more signal into GEPA to actually drive better optimizations 🚀

PR #2512, opened 2026-05-13
add py.typed markers to otel/semconv and hotspots; add SpanType.GUARDRAIL

### LINKED ISSUE #2468: Add SpanType.GUARDRAIL to Semantic Conventions
## Description

TruLens has first-class guardrail support (`test_otel_guardrail.py`, blocking guardrails docs), but there's no dedicated `SpanType.GUARDRAIL` in the semantic conventions. Guardrail spans currently show up as `UNKNOWN` in traces, making it harder to filter and visualize them.

## Proposed Change

Add a `GUARDRAIL` variant to the `SpanType` enum in `src/otel/semconv/trulens/otel/semconv/trace.py:60`:

```python
GUARDRAIL = "guardrail"
"""A guardrail check execution."""
```

Also add a corresponding `GUARDRAIL` attribute class (following the pattern of `TOOL`, `AGENT`, etc.) with attributes like:
- `GUARDRAIL.NAME` — name of the guardrail
- `GUARDRAIL.RESULT` — pass/fail result
- `GUARDRAIL.THRESHOLD` — configured threshold

## Files to Change

- `src/otel/semconv/trulens/otel/semconv/trace.py` — add enum value + attribute class
- Any guardrail instrumentation code that currently uses `UNKNOWN` should be updated to use `GUARDRAIL`

## Testing

- Add tests verifying the new span type is correctly set on guardrail spans.

### LINKED ISSUE #2479: Add Missing py.typed Marker to otel/semconv and hotspots
## Description

Every TruLens sub-package ships a `py.typed` marker file for [PEP 561](https://peps.python.org/pep-0561/) compliance (so type checkers like mypy/pyright recognize the package as typed) — except `src/otel/semconv/` and `src/hotspots/`.

## Proposed Change

Add empty `py.typed` files:

- `src/otel/semconv/trulens/otel/semconv/py.typed`
- `src/hotspots/trulens/hotspots/py.typed` (if not already present)

Also verify the `pyproject.toml` for each package includes the marker in its package data if needed.

## Files to Change

- Create `src/otel/semconv/trulens/otel/semconv/py.typed` (empty file)
- Create `src/hotspots/trulens/hotspots/py.typed` (empty file, if missing)
- Check `pyproject.toml` files for both packages

ykd007:
Closes #2479, closes #2468

## Changes

**py.typed markers (#2479)**
- Add `src/otel/semconv/trulens/otel/semconv/py.typed` (empty PEP 561 marker)
- Add `src/hotspots/trulens/hotspots/py.typed` (empty PEP 561 marker)
- No `pyproject.toml` changes needed — both packages already use `{ include = "trulens" }`

**SpanType.GUARDRAIL (#2468)**
- Add `GUARDRAIL = "guardrail"` to the `SpanType` enum in `trace.py`
- Add `SpanAttributes.GUARDRAIL` attribute class with `NAME`, `SCORE`, `PASSED`, `THRESHOLD`
- Wire OTEL spans into `context_filter`, `block_input`, `block_output` — spans wrap the feedback evaluation call so duration is meaningful
- Guard with `_OTEL_AVAILABLE` so guardrails still work when opentelemetry is not installed
- Update `AGENTS.md` span types list

### DIFF
--- a/src/core/trulens/core/guardrails/base.py
+++ b/src/core/trulens/core/guardrails/base.py
@@ -8,6 +8,21 @@ from trulens.core.utils import threading as threading_utils
 
 logger = logging.getLogger(__name__)
 
+_OTEL_AVAILABLE = False
+try:
+    from opentelemetry import trace as otel_trace
+    from trulens.experimental.otel_tracing.core.session import (
+        TRULENS_SERVICE_NAME,
+    )
+    from trulens.experimental.otel_tracing.core.span import (
+        set_general_span_attributes,
+    )
+    from trulens.otel.semconv.trace import SpanAttributes
+
+    _OTEL_AVAILABLE = True
+except ImportError:
+    pass
+
 
 class context_filter:
     """Provides a decorator to filter contexts based on a given feedback and threshold.
@@ -69,15 +84,47 @@ class context_filter:
             bindings = sig.bind(*args, **kwargs)
 
             contexts = func(*args, **kwargs)
+            guardrail_name = getattr(self.feedback, "name", func.__name__)
+
+            def _evaluate_context(context):
+                """Run feedback evaluation, wrapped in a guardrail span."""
+                if _OTEL_AVAILABLE:
+                    tracer = otel_trace.get_tracer_provider().get_tracer(
+                        TRULENS_SERVICE_NAME
+                    )
+                    with tracer.start_as_current_span(guardrail_name) as span:
+                        set_general_span_attributes(
+                            span, SpanAttributes.SpanType.GUARDRAIL
+                        )
+                        span.set_attribute(
+                            SpanAttributes.GUARDRAIL.NAME, guardrail_name
+                        )
+                        span.set_attribute(
+                            SpanAttributes.GUARDRAIL.THRESHOLD, self.threshold
+                        )
+                        result = self.feedback(
+                            bindings.arguments[self.keyword_for_prompt], context
+                        )
+                        passed = (
+                            self.feedback.higher_is_better
+                            and result > self.threshold
+                        ) or (
+                            not self.feedback.higher_is_better
+                            and result < self.threshold
+                        )
+                        span.set_attribute(SpanAttributes.GUARDRAIL.SCORE, result)
+                        span.set_attribute(SpanAttributes.GUARDRAIL.PASSED, passed)
+                        return result
+                else:
+                    return self.feedback(
+                        bindings.arguments[self.keyword_for_prompt], context
+                    )
+
             with threading_utils.ThreadPoolExecutor(
                 max_workers=max(1, len(contexts))
             ) as ex:
                 future_to_context = {
-                    ex.submit(
-                        lambda context=context: self.feedback(
-                            bindings.arguments[self.keyword_for_prompt], context
-                        )
-                    ): context
+                    ex.submit(_evaluate_context, context): context
                     for context in contexts
                 }
                 filtered = []
@@ -88,13 +135,14 @@ class context_filter:
                         raise ValueError(
                             "`context_filter` can only be used with feedback functions that return a float."
                         )
-                    if (
+                    passed = (
                         self.feedback.higher_is_better
                         and result > self.threshold
                     ) or (
                         not self.feedback.higher_is_better
                         and result < self.threshold
-                    ):
+                    )
+                    if passed:
                         filtered.append(context)
                 return filtered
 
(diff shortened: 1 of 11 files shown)


joshreini1 (commented):
Nice work adding guardrail observability! The semconv additions are clean and the py.typed markers are a welcome fix. A few suggestions below to tighten things up before merging.

Also worth adding tests for the new span instrumentation — verifying NAME, SCORE, PASSED, THRESHOLD attributes are emitted correctly for each guardrail type.

src/core/trulens/core/guardrails/base.py
```diff
@@ -8,6 +8,21 @@
...
+try:
+    from opentelemetry import trace as otel_trace
+    from trulens.experimental.otel_tracing.core.session import (
+        TRULENS_SERVICE_NAME,
+    )
+    from trulens.experimental.otel_tracing.core.span import (
+        set_general_span_attributes,
+    )
+    from trulens.otel.semconv.trace import SpanAttributes
+
```
joshreini1, line 18: OTEL is an expected dependency of `trulens-core` — every other file in the codebase imports these directly (e.g. `tru_graph.py`, `instrument.py`, `computer.py`). The `_OTEL_AVAILABLE` try/except guard + all the `if/else` branches roughly double the new code for a case that shouldn't happen in practice. I'd suggest dropping the guard and importing directly, which will also eliminate most of the duplication below.

src/core/trulens/core/guardrails/base.py
```diff
@@ -69,15 +84,47 @@ def wrapper(*args, **kwargs):
...
                 future_to_context = {
-                    ex.submit(
-                        lambda context=context: self.feedback(
-                            bindings.arguments[self.keyword_for_prompt], context
-                        )
-                    ): context
+                    ex.submit(_evaluate_context, context): context
                     for context in contexts
                 }
                 filtered = []
```
joshreini1, line 119: The span setup logic (get tracer → start span → set NAME, THRESHOLD, SCORE, PASSED) is repeated nearly identically across `context_filter`, `block_input`, and `block_output`. A small context manager or helper function would keep things DRY and make future attribute additions a one-line change instead of three.

src/core/trulens/core/guardrails/base.py
```diff
@@ -88,13 +135,14 @@ def wrapper(*args, **kwargs):
...
-                    if (
+                    passed = (
                         self.feedback.higher_is_better
                         and result > self.threshold
                     ) or (
                         not self.feedback.higher_is_better
                         and result < self.threshold
-                    ):
+                    )
+                    if passed:
```
joshreini1, line 132: This `passed` condition is the same logic that was already computed inside `_evaluate_context` above (line ~112). Since it's calculated twice and could drift, consider having `_evaluate_context` return a `(result, passed)` tuple so it's only computed once.

src/core/trulens/core/guardrails/base.py
```diff
@@ -248,15 +333,53 @@ def __call__(self, func):
...
                 )
-            if (self.feedback.higher_is_better and result < self.threshold) or (
-                not self.feedback.higher_is_better and result > self.threshold
-            ):
+                with tracer.start_as_current_span(guardrail_name) as span:
+                    set_general_span_attributes(
+                        span, SpanAttributes.SpanType.GUARDRAIL
+                    )
+                    span.set_attribute(
+                        SpanAttributes.GUARDRAIL.NAME, guardrail_name
```
joshreini1: Heads up — unlike `block_input`, this span wraps the decorated function's execution *and* the feedback evaluation. So the span duration will include the app's own execution time, not just the guardrail check. If that's intentional, worth a brief comment. If not, moving `output = func(*args, **kwargs)` above the span context would fix it.

ykd007:
Thanks for the detailed review! All four points addressed in the latest commit:

- **_OTEL_AVAILABLE guard removed** — switched to direct imports matching the rest of the codebase
- **DRY helper added** — extracted `_guardrail_span()` context manager; NAME/SCORE/PASSED/THRESHOLD are set in one place, used by all three decorators
- **Return tuple from _evaluate_context** — `(result, passed)` so pass/fail logic is computed once, not twice
- **block_output span placement** — `func()` now runs before the span opens so span duration reflects only the feedback evaluation
- **Tests added** — `tests/unit/test_otel_guardrail_spans.py` covers all three decorators (pass + blocked cases) and verifies NAME, SCORE, PASSED, THRESHOLD attributes

src/core/trulens/core/guardrails/base.py
```diff
@@ -1,14 +1,32 @@
...
 from trulens.core.utils import threading as threading_utils
+from trulens.experimental.otel_tracing.core.session import TRULENS_SERVICE_NAME
+from trulens.experimental.otel_tracing.core.span import set_general_span_attributes
+from trulens.otel.semconv.trace import SpanAttributes
 
 logger = logging.getLogger(__name__)
 
 
+@contextmanager
+def _guardrail_span(name: str, threshold: float, score: float, passed: bool):
```
joshreini1: The context manager takes `score` and `passed` upfront and sets them before `yield`, which means the callers do `with _guardrail_span(...): pass` — the span doesn't actually wrap any work. More conventional approach: have the context manager only take `name` + `threshold`, yield the span, and let callers set SCORE/PASSED inside the block after computing them:

```python
@contextmanager
def _guardrail_span(name: str, threshold: float):
    tracer = otel_trace.get_tracer_provider().get_tracer(TRULENS_SERVICE_NAME)
    with tracer.start_as_current_span(name) as span:
        set_general_span_attributes(span, SpanAttributes.SpanType.GUARDRAIL)
        span.set_attribute(SpanAttributes.GUARDRAIL.NAME, name)
        span.set_attribute(SpanAttributes.GUARDRAIL.THRESHOLD, threshold)
        yield span

# Usage:
with _guardrail_span(guardrail_name, self.threshold) as span:
    result = self.feedback(keyword_value)
    blocked = ...
    span.set_attribute(SpanAttributes.GUARDRAIL.SCORE, result)
    span.set_attribute(SpanAttributes.GUARDRAIL.PASSED, not blocked)
```

This way the span duration actually measures the feedback evaluation, the `yield` wraps real work, and there's no `with ... pass` anti-pattern.

joshreini1 (commented):
Nice cleanup — all the original comments are addressed. One more suggestion on the helper pattern, then this looks good to go.

joshreini1:
Thanks @ykd007 - added one more comment, otherwise it's good to merge once you've addressed it.

ykd007:
Good call — updated. `_guardrail_span` now only takes `name` + `threshold` and yields the span. Each caller sets SCORE and PASSED inside the `with` block after computing them, so the span duration actually covers the feedback evaluation. No more `with ... pass` anti-pattern.

joshreini1 (approved):
LGTM - thanks for contributing!

joshreini1:
@ykd007 there are some formatting issues breaking the CI, you can fix by running `make run-precommit` and committing the result.

ykd007:
Thanks for the thorough review @joshreini1! All checks are passing and the PR is approved — just pinging in case this is ready for merge on your end.

ykd007:
Gentle ping — CI is green and this has an approval. Happy to rebase if needed, otherwise ready to merge whenever convenient.

sfc-gh-jreini:
Merged! Thank you @ykd007 !!

seals187:
@sfc-gh-jreini @joshreini1 @andrewisplinghoff @piotrm0 just dont forget that this guy is AI slop , mentionned here: https://turso.tech/blog/the-wonders-of-ai and here https://github.com/tursodatabase/turso/pull/6874

Issue #2471, opened 2026-05-13
Add Unit Tests for Feedback Template Validation


joshreini1:
## Description

`test_feedback_templates_exports.py` only checks that templates are properly re-exported. There are no tests verifying that each template class has valid, well-formed prompts.

## Proposed Change

Create `tests/unit/test_feedback_templates_validation.py` that iterates over all exported template classes and verifies:

- `system_prompt` is a non-empty string
- `user_prompt` is a non-empty string (where applicable)
- `criteria` is a non-empty string (for classes using `CriteriaOutputSpaceMixin`)
- `output_space` is a valid `OutputSpace` variant
- Prompt templates have no unresolved `{placeholder}` formatting artifacts
- Template classes are valid Pydantic models (can be instantiated)

## Files to Reference

- `src/feedback/trulens/feedback/templates/__init__.py` — iterate `__all__`
- Each domain file: `base.py`, `rag.py`, `safety.py`, `quality.py`, `agent.py`

## Running Tests

```bash
TEST_OPTIONAL=true poetry run pytest tests/unit/test_feedback_templates_validation.py -v
```

devdattatalele:
Hey @joshreini1 — I'd like to take this on. Plan: add `tests/unit/test_feedback_templates_validation.py` that iterates `templates.__all__` and asserts each class satisfies the field/format checks in the issue — non-empty `system_prompt`/`user_prompt`/`criteria`, valid `OutputSpace` variant, no unresolved `{placeholder}` artifacts (via a probing `str.format_map` pass), and Pydantic-instantiable. Will mirror the fixture/skip pattern from `test_feedback_templates_exports.py` so the suite stays consistent.

devdattatalele:
Small correction on my prior comment — after cloning I checked the tree and this looks like it's already implemented. `tests/unit/test_feedback_templates_validation.py` was added by @1Utkarsh1 in #2516 (merged 2026-05-17) and covers all six checks in the description (Pydantic instantiation, non-empty `system_prompt`/`user_prompt`/`criteria`, valid `OutputSpace`, and prompt fields restricted to the allowed provider signature set).

Might be safe to close this one @joshreini1 — sorry for the noise, should've grepped the tree before claiming. Happy to grab a different unclaimed ticket instead.

sfc-gh-jreini:
Thanks @devdattatalele for checking! Since @1Utkarsh1 already landed this in #2516, closing this issue as resolved.

[...]

PR #2516, opened 2026-05-15
test: add feedback template validation coverage

1Utkarsh1:
## Summary

- Add unit tests for exported feedback prompt template definitions.
- Make `FeedbackTemplate` expose `examples` as an optional Pydantic field so exported prompt classes can instantiate cleanly.

## Why

The existing export test checks that template symbols are re-exported. This adds coverage for prompt-bearing templates themselves: non-empty prompts, valid output spaces, criteria for `CriteriaOutputSpaceMixin` classes, expected runtime format fields, and Pydantic instantiation.

## Testing

- `TEST_OPTIONAL=true uv run pytest tests/unit/test_feedback_templates_validation.py -q`
- `uv run pytest tests/unit/test_feedback_templates_exports.py tests/unit/test_feedback_templates_validation.py -q`
- `uvx ruff check --select I,E303 tests/unit/test_feedback_templates_validation.py`
- `uvx ruff format tests/unit/test_feedback_templates_validation.py --check`

## Notes

Related to #2471. I used `uv` locally because `poetry` was not installed in my environment.

### DIFF
--- a/tests/unit/test_feedback_templates_validation.py
+++ b/tests/unit/test_feedback_templates_validation.py
@@ -1,0 +1,134 @@
+"""Validation tests for exported feedback template definitions."""
+
+from collections.abc import Iterable
+from string import Formatter
+import unittest
+
+from trulens.feedback import templates
+from trulens.feedback.templates import base as templates_base
+
+_PROMPT_ATTRIBUTES = ("system_prompt", "user_prompt", "prompt")
+_ALLOWED_PROMPT_FIELDS = {
+    "context",
+    "hypothesis",
+    "key_point",
+    "max_score",
+    "min_score",
+    "premise",
+    "prompt",
+    "question",
+    "response",
+    "source",
+    "statement",
+    "statements",
+    "submission",
+    "summary",
+    "text",
+    "trace",
+}
+
+
+def _exported_prompt_template_classes() -> Iterable[
+    type[templates_base.FeedbackTemplate]
+]:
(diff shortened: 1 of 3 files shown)


tests/unit/test_feedback_templates_validation.py
```diff
@@ -0,0 +1,134 @@
...
+
+from collections.abc import Iterable
+from string import Formatter
+import unittest
+
+from trulens.feedback import templates
+from trulens.feedback.templates import base as templates_base
+
+_PROMPT_ATTRIBUTES = ("system_prompt", "user_prompt", "prompt")
+_ALLOWED_PROMPT_FIELDS = {
```
sfc-gh-jreini: can we dynamically create this from the feedback function signatures so we don't have to maintain this?

sfc-gh-jreini (commented):
one request, otherwise looks good - thanks for contributing!

Just the format/pre-commit is failing in CI, you can fix by running pre-commit and committing the result.

`make run-precommit`

tests/unit/test_feedback_templates_validation.py
1Utkarsh1: I’ve updated it so the allowed fields are derived from the provider code instead of keeping a hardcoded list.

1Utkarsh1:
@sfc-gh-jreini can you approve this , i think you approved my drafted PR instead of this

sfc-gh-jreini (approved):

PR #2462, opened 2026-05-09
feat: add conversation_id support for thread-based trace grouping

nuthalapativarun:
## Description

Adds `ai.observability.conversation_id` — a span attribute that groups multiple app invocations (records) into a single conversation/thread. Closes #2423.

### Usage

```python
# Pass conversation_id via __call__ (recommended)
with tru_app(conversation_id="conv-abc-123") as recording:
    response = app.query("First message")

# Later turn — same conversation
with tru_app(conversation_id="conv-abc-123") as recording:
    response = app.query("Follow-up message")
```

All spans recorded within the context carry `ai.observability.conversation_id = "conv-abc-123"`.

### Changes

**`src/otel/semconv/trulens/otel/semconv/trace.py`**
- Added `SpanAttributes.CONVERSATION_ID = "ai.observability.conversation_id"` with docstring

**`src/core/trulens/experimental/otel_tracing/core/span.py`**
- `set_general_span_attributes` now propagates `CONVERSATION_ID` from OTEL baggage to each span, alongside existing attributes (RECORD_ID, RUN_NAME, INPUT_ID, etc.)

**`src/core/trulens/core/otel/instrument.py`**
- `OtelRecordingContext.__init__` accepts optional `conversation_id: Optional[str] = None`
- `OtelRecordingContext.__enter__` calls `attach_to_context(SpanAttributes.CONVERSATION_ID, ...)` when set, placing it in OTEL baggage so all child spans inherit it

**`src/core/trulens/core/app.py`**
- `App.__call__` added — stores `conversation_id` on `self._pending_conversation_id` and returns `self`, enabling the `tru_app(conversation_id="...")` context manager syntax
- `App.__enter__` reads `_pending_conversation_id` and passes it to `OtelRecordingContext`

## Other details good to know for developers

The `conversation_id` is propagated through OTEL baggage, so it follows the same propagation path as `record_id`, `run_name`, and `input_id` — no special handling needed in child spans.

The `App.__call__` pattern is backward-compatible: `with tru_app as recording:` (no call) continues to work exactly as before with `conversation_id=None`.

## Type of change

- [x] New feature (non-breaking change which adds functionality)

PR #2526, opened 2026-05-30
feat: add MemoryRecallMetric for agent memory evaluation

### LINKED ISSUE #2524: Feature: Memory recall evaluation metrics for agent systems
Hi TruLens team,

Your RAG Triad metrics (Context Relevance, Groundedness, Answer Relevance) have become the standard for RAG evaluation. As agents become more prominent, I think there's a gap in evaluating **agent memory recall quality** that's quite different from static document retrieval.

Agent memory systems have unique evaluation needs:
- **Temporal relevance**: Does the right memory surface based on time? ("What did the user ask about last week?")
- **Forgetting directionality**: High-frequency memories should be retained over low-frequency ones
- **Cross-session continuity**: Can the agent recall user preferences across days/weeks?
- **Multi-hop reasoning**: Can the system chain across memories to answer complex queries?
- **Deep retrieval decay**: How does recall quality degrade for older memories?

I've been working on [MemTest](https://github.com/yubingz/memtest), which designs benchmark databases specifically for these agent memory scenarios. Key differentiator: it provides **test data generators** (both procedural and corpus-driven) rather than just metrics, so you can stress-test any memory system with controlled, reproducible inputs.

We tested with Chinese classical literature (Four Great Classical Novels, 21K+ memories) and found some surprising results — e.g., sentence-transformers (MiniLM) only achieves 9.1% precision on entity-level retrieval in these texts, while TF-IDF + LLM reranking hits 87%.

A `MemoryRecallMetric` that integrates with TruLens's existing framework could be valuable for the growing agent evaluation use case. Would this align with your roadmap?

Happy to discuss the design or contribute a draft implementation.

yubingz:
## Description

Adds `MemoryRecallMetric` — a feedback function for evaluating agent memory system recall quality, as discussed in #2524.

### What it does

Evaluates whether an agent memory store retrieves the right memories, with:
- `precision_at_k` / `recall_at_k` / `mrr` against ground truth memory IDs
- **`chain_completeness`**: fraction of a reasoning chain retrieved — captures whether the system can support multi-hop logical reasoning, not just single-fact retrieval
- Conversation-scoped evaluation via `conversation_id` (following the pattern from #2462)

### Example

```python
from trulens.feedback.memory_recall import MemoryRecallMetric

metric = MemoryRecallMetric(
    ground_truth={
        "conv_abc": {
            "mem_1": {"query": "user preferences", "relevant": ["mem_1", "mem_2"], "chain": ["mem_1", "mem_2"]},
        }
    },
    k=10
)

result = metric.evaluate(
    conversation_id="conv_abc",
    query="user preferences",
    retrieved_ids=["mem_1", "mem_5", "mem_3"],
)
# => {"precision_at_k": 0.33, "recall_at_k": 0.5, "mrr": 1.0, "chain_completeness": 0.5}
```

### Detecting issues

The included example shows how `chain_completeness` detects a specific failure mode: when a memory system retrieves only the endpoint of a reasoning chain (chain_completeness=0.33) but misses the prior context needed to answer "why" questions. Existing metrics like `recall_at_k` cannot distinguish this case from a full chain.

### Files

- `src/feedback/trulens/feedback/memory_recall.py` — MemoryRecallMetric implementation
- `src/feedback/trulens/feedback/__init__.py` — Export MemoryRecallMetric
- `examples/expositional/memory_recall/memory_recall_example.py` — Usage example

Closes #2524


joshreini1:
Shared more on direction in the issue: #2524

src/feedback/trulens/feedback/groundtruth.py
```diff
@@ -61,6 +61,9 @@ class GroundTruthAgreement(
 
     ground_truth_imp: Optional[Callable] = pydantic.Field(None, exclude=True)
 
+    conversation_id: Optional[str] = None
```
joshreini1: Since records are turns and a conversation is the set of turns sharing a `conversation_id`, the conversation_id naturally lives *on each record* — so the idiomatic fit is a per-call argument the feedback pipeline can populate via a `Selector`, e.g. `recall_at_k(query, retrieved, conversation_id=..., k=5)`. That way a single GTA instance holds the whole multi-conversation golden set and each turn gets routed to its own conversation's ground truth in one evaluation pass (rather than one frozen instance + metric registration per conversation).

The lookup needs scoping for correctness too: the same natural-language query recurs across conversations with different correct answers, and an unscoped lookup silently unions every conversation's `expected_chunks` for that query (inflating the recall denominator). Keeping an optional `conversation_id` on the constructor as a default is fine for the manual single-conversation drill-down case — just let the per-call arg override it.

src/feedback/trulens/feedback/groundtruth.py
```diff
@@ -768,6 +777,179 @@ def mae(self):
...
+        for qr in self.ground_truth:
+            if qr.get("query") == query:
+                if self.conversation_id is not None:
+                    if qr.get("conversation_id") != self.conversation_id:
+                        continue
+                if "expected_memories" in qr:
+                    return qr["expected_memories"]
+        return None
+
+    def memory_recall(
```
joshreini1, line 791: I'd suggest removing this and extending the existing `recall_at_k` (around L432) instead — it already does chunk-text set membership and adds `k`, `relevance_scores`, and tie-handling that this reimplementation drops. Same logic, one code path. While we're here, let's keep `precision_at_k` in play too: low precision is the "irrelevant memories were retrieved" signal we discussed in #2524.

src/feedback/trulens/feedback/groundtruth.py
```diff
@@ -768,6 +777,179 @@ def mae(self):
...
+        matched = 0
+        for exp_mem in unique_expected:
+            for ret_mem in retrieved_memories:
+                if self._is_similar(exp_mem, ret_mem, similarity_threshold):
+                    matched += 1
+                    break
+
+        return matched / len(unique_expected)
+
+    def memory_mrr(
```
joshreini1, line 877: Same thought here — `mrr` (around L493) already computes first-relevant reciprocal rank over chunk texts. Folding `conversation_id` into that method avoids a second, slightly-diverging implementation (this one iterates `expected_set` inside the retrieved loop and skips the relevance-score sorting the original supports).

src/feedback/trulens/feedback/groundtruth.py
```diff
@@ -768,6 +777,179 @@ def mae(self):
         raise NotImplementedError("`mae` has moved to `GroundTruthAggregator`")
 
 
+    def _find_expected_memories(self, query: str) -> Optional[List[str]]:
```
joshreini1: Rather than a new lookup + `expected_memories: [str]` schema, could we add an optional `conversation_id` filter to the existing `_find_golden_context_chunks_and_scores` (around L207) and reuse `expected_chunks: [{"text": ..., "expect_score": 1}]`? That keeps a single ground-truth format in the class (right now `conversation_id` only affects this new path, not `_find_response`/`_find_score`, which is a bit surprising). Memory texts go in as `expected_chunks` exactly like RAG chunks.

src/feedback/trulens/feedback/groundtruth.py
```diff
@@ -768,6 +777,179 @@ def mae(self):
...
+        expected_set = set(expected)
+        for i, ret_mem in enumerate(retrieved_memories):
+            for exp_mem in expected_set:
+                if self._is_similar(ret_mem, exp_mem, similarity_threshold):
+                    return 1.0 / (i + 1)
+
+        return 0.0
+
+    @staticmethod
+    def _is_similar(text1: str, text2: str, threshold: float = 1.0) -> bool:
```
joshreini1, line 924: Let's pull fuzzy matching out into a separate PR. For a curated ground-truth set, exact query matching is a reasonable expectation since the keys are controlled, and the existing metrics already match chunk texts exactly. When we do tackle fuzzy/semantic matching, it'd be worth doing it on the shared helper with a real semantic matcher rather than `difflib`.

examples/expositional/memory_recall/memory_recall_example.py
```diff
@@ -0,0 +1,177 @@
+"""
```
joshreini1, line 1: Could we turn this into a follow-along notebook rather than a script? The most compelling story here is end-to-end: stand up an agent with a memory store, show it *missing* required memories on a follow-up question (e.g. failing to recall a preference or commitment from earlier in the conversation), score that with `recall_at_k`/`precision_at_k`/`mrr` scoped by `conversation_id`, then adjust the agent (retrieval params, what's written to memory, conversation scoping) and show the scores improve. That turns the metric into an actionable debugging loop instead of a static demo, and a notebook lets readers run each step and see the before/after.

tests/memory_recall/test_memory_recall.py
```diff
@@ -0,0 +1,333 @@
+"""Tests for GroundTruthAgreement memory recall methods."""
```
joshreini1, line 1: Could we move these under the standard `tests/unit/...` path so CI discovery picks them up? Once the methods are folded into the existing ones, these can retarget `recall_at_k`/`precision_at_k`/`mrr` with `conversation_id` — they're already structured well, so it should mostly be renaming the calls.

joshreini1 (commented):
Thanks for the fast turnaround on this, @yubingz — the test suite is genuinely thorough (the dedup, conversation-isolation, and invalid-threshold cases are exactly the right things to cover), and dropping `chain_completeness` + returning plain `float` are spot on. We're close. The one remaining shift is to lean *all the way* into reusing the existing IR metrics instead of standing up a parallel set, which keeps memory recall flowing through the same pipeline/dashboard everything else uses. Three concrete changes below, all of which should let you *delete* code rather than add it:

1. **Surface `conversation_id` as a per-call argument on the existing metrics** (selector-populated from each turn-record), rather than constructor state.
2. **Reuse `recall_at_k` / `precision_at_k` / `mrr`** instead of new `memory_recall` / `memory_mrr` methods (this also brings back precision, which is the signal for "irrelevant memories surfaced").
3. **Reuse the existing `expected_chunks` ground-truth schema** instead of the new `expected_memories` field, by teaching `_find_golden_context_chunks_and_scores` about `conversation_id`.

On **fuzzy matching / `similarity_threshold`**: let's split that out into its own follow-up PR. For a curated ground-truth set, exact query matching is a reasonable expectation since the keys are controlled rather than free-form user input — that keeps this PR focused on the metric plumbing.

yubingz:
Hi Josh, thanks for the detailed review! I've pushed a v2 that addresses all 7 points:

1. **conversation_id as per-call arg** — Added to `recall_at_k`, `precision_at_k`, `mrr` as optional parameter. Instance-level default on constructor preserved, per-call overrides it. (`_find_golden_context_chunks_and_scores` uses `effective_conv_id` to resolve the override.)

2. **Removed `memory_recall`, extended `recall_at_k`** — Deleted. `recall_at_k` now accepts `conversation_id` and includes docstring noting memory recall use case.

3. **Removed `memory_mrr`, extended `mrr`** — Deleted. `mrr` now accepts `conversation_id` and `relevance_scores` (was missing before).

4. **No new `expected_memories` format** — Deleted. Memory texts go into `expected_chunks` exactly like RAG chunks, with `conversation_id` filtering added to `_find_golden_context_chunks_and_scores`.

5. **Removed fuzzy matching** — Deleted entirely. Exact match only, as suggested. Can revisit in a separate PR with a real semantic matcher.

6. **Follow-along notebook** — Replaced the .py script with a .ipynb that walks through a before/after debugging loop: broken agent (low recall/precision/MRR) → diagnose → fix → re-score (all 1.0), plus conversation scoping and per-call vs instance-level examples.

7. **Tests under `tests/unit/`** — Moved to `tests/unit/test_memory_recall.py`. 14 tests, all calling `recall_at_k`/`precision_at_k`/`mrr` with `conversation_id`.

All 14 tests pass. Happy to iterate further!

joshreini1:
**Before merge — CI is red; here's how to get it passing**

**pre-commit / ruff** (the `Bash exited with code 2` stage):
   - `tests/unit/test_memory_recall.py` — unsorted imports (`I001`) and an unused `import pytest` (`F401`).
   Running `pre-commit run --all-files` (or `ruff check --fix && ruff format`) and committing the result will fix it up.

---

**Enhancement idea (non-blocking): show the full TruLens recording loop, and consider promoting this to the quickstart folder.**

This is already a great deterministic walkthrough. The natural next step that would make it a flagship memory-eval example is to run it through an actual instrumented app + `TruSession` recording and read results from `get_leaderboard()`/the dashboard instead of `print`s — so memory recall shows up in the same place users already see the RAG triad, and the `conversation_id` from #2462 is exercised end-to-end.

The nice part: `conversation_id` is the `ai.observability.conversation_id` span attribute (#2462), so you can select it straight into `recall_at_k`'s argument and one metric definition scopes correctly for every conversation and app version. I validated this end-to-end against this branch — same query + same retrieval scored **0.5** under `conv_1` (expects dark mode + Python) and **0.0** under `conv_2` (expects light mode), with the only difference being the `conversation_id` on the recording context:

```python
from trulens.core import TruSession, Metric, Selector
from trulens.apps.app import TruApp
from trulens.core.otel.instrument import instrument
from trulens.otel.semconv.trace import SpanAttributes

session = TruSession()

class MemoryAgent:
    @instrument(
        span_type=SpanAttributes.SpanType.RETRIEVAL,
        attributes={
            SpanAttributes.RETRIEVAL.QUERY_TEXT: "query",
            SpanAttributes.RETRIEVAL.RETRIEVED_CONTEXTS: "return",
        },
    )
    def retrieve_memories(self, query: str) -> list[str]:
        ...

    @instrument(span_type=SpanAttributes.SpanType.RECORD_ROOT)
    def answer(self, query: str) -> str:
        return summarize(self.retrieve_memories(query))

f_recall = Metric(
    implementation=gta.recall_at_k,
    name="Memory Recall",
    selectors={
        "query": Selector(span_type=SpanAttributes.SpanType.RETRIEVAL,
                          span_attribute=SpanAttributes.RETRIEVAL.QUERY_TEXT),
        "retrieved_context_chunks": Selector.select_context(collect_list=True),
        "conversation_id": Selector(span_attribute=SpanAttributes.CONVERSATION_ID),
    },
)
# f_precision / f_mrr are identical, swapping implementation=gta.precision_at_k / gta.mrr

tru_agent = TruApp(agent, app_name="memory-agent", app_version="v1-broken",
                   main_method=agent.answer, feedbacks=[f_recall])

with tru_agent(conversation_id="conv_1") as recording:
    agent.answer("What are the user's preferences?")

session.get_leaderboard()   # Memory Recall per app version
```

The before/after story then becomes two app versions (`v1-broken` → `v2-fixed`) on the leaderboard rather than a printed table. The retrieval step just needs to be instrumented as a RETRIEVAL span (the standard RAG-triad pattern) so `select_context` picks up the retrieved memory texts.

If you'd rather keep a run scoped to a single conversation, the simpler alternative is to set the instance-level default — `GroundTruthAgreement(golden_set, provider=..., conversation_id="conv_1")` — and drop the `conversation_id` selector. The tradeoff is one GTA/metric per conversation, so the selector approach above is cleaner for a mixed-conversation dataset.

And with this recording + leaderboard upgrade, this would be a strong fit for `examples/quickstart/` (right next to `groundtruth_evals_for_retrieval_systems.ipynb`) rather than `expositional` — a complete, runnable memory-eval quickstart. Totally fine as a follow-up either way; the current deterministic version is a nice "metrics 101" intro before the full recording.

yubingz:
Thanks Josh! Fixed the ruff issues (removed unused `pytest` import, sorted imports) — CI should pass now on the latest push.

The recording loop + leaderboard upgrade is a great idea. I'll work on that as a follow-up PR — the current deterministic version serves as a solid baseline, and the instrumented `TruSession` version will show it in the dashboard where users already look. The `conversation_id` selector approach you outlined is clean; I'll use that pattern.

joshreini1 (approved):
looks good! will merge when CI is passing. Ping me if I'm slow :)

yubingz:
Hi @joshreini1, just a heads-up — all CI checks are now passing (lint, py310/py311/py313 tests, build, security). The PR is ready for merge whenever you get a chance. Thanks for the review!

PR #2530, opened 2026-06-05
Add HuggingFace Provider Capability Tests

### LINKED ISSUE #2475: Add HuggingFace Provider Capability Tests
## Description

Same gap — the HuggingFace provider (`src/providers/huggingface/`) has no unit tests. Note that `HuggingfaceBase` extends `core_provider.Provider` (not `LLMProvider`), so the test pattern will differ slightly.

## Proposed Change

Create `tests/unit/providers/test_huggingface_capabilities.py` covering:

- Local vs API-based initialization
- Model loading behavior (mocked)
- Feedback function execution path
- Error handling for missing models

## Reference Pattern

Use `tests/unit/providers/test_openai_capabilities.py` as a structural reference, but adapt for the `Provider` base class.

## Running Tests

```bash
TEST_OPTIONAL=true poetry run pytest tests/unit/providers/test_huggingface_capabilities.py -v
```

Animesh452:
Closes #2475

Creates `tests/unit/providers/test_huggingface_capabilities.py` covering 
all four areas from the issue:

- **Initialization**: `Huggingface` (via `Dummy`) and `HuggingfaceLocal` instantiate without errors
- **Error handling**: `_tci` decorator raises `ValueError` on empty string, `TypeError` on wrong type
- **Feedback execution paths**: `positive_sentiment`, `toxic`, and `context_relevance` return correct float values given mocked `_post_scores`
- **Model caching**: `HuggingfaceLocal._retrieve_tokenizer_and_model` loads on first call and returns cached objects on subsequent calls with same key

Used `Dummy` class for API-based tests (no network/API key needed) and `monkeypatch` for isolating external dependencies.


tests/unit/providers/test_huggingface_capabilities.py
```diff
@@ -0,0 +1,142 @@
...
+    assert result == 0.7
+
+
+@pytest.mark.optional
+def test_toxic_returns_float(monkeypatch):
+    from trulens.providers.huggingface.provider import Dummy
+
+    provider = Dummy()
+
+    fake_scores = [
```
joshreini1, line 65: The fake_scores use `"toxic"` as a label, but the real `toxic` method expects labels like `"LABEL_0"`/`"LABEL_1"` (HuggingFace's standard classifier output format — same pattern as `positive_sentiment` on line 45). Change the fake labels to match the actual model output format or the test will pass here but fail against the real implementation.

tests/unit/providers/test_huggingface_capabilities.py
```diff
@@ -0,0 +1,142 @@
...
+    assert result == 0.9
+
+
+@pytest.mark.optional
+def test_context_relevance_returns_float(monkeypatch):
+    from trulens.providers.huggingface.provider import Dummy
+
+    provider = Dummy()
+
+    fake_scores = [
```
joshreini1, line 84: Same issue: `"context_relevance"` isn't a real label format. HuggingFace classifiers return `"LABEL_N"` strings (where N is the class index). Update the fake_scores to use the correct label format so the test validates the actual parsing logic.

tests/unit/providers/test_huggingface_capabilities.py
```diff
@@ -0,0 +1,142 @@
...
+        {"label": "LABEL_1", "score": 0.2},
+        {"label": "LABEL_2", "score": 0.7},
+    ]
+    monkeypatch.setattr(
+        Dummy, "_post_scores", lambda self, **kwargs: fake_scores
+    )
+
+    result = provider.positive_sentiment(text="This is great!")
+    assert isinstance(result, float)
+    assert result == 0.7
```
joshreini1, line 56: Missing edge-case tests: what happens if `_post_scores` returns an empty list? The feedback methods use `max(...)` which raises `ValueError` on empty sequences. Add a test for that path or verify the methods handle it gracefully.

joshreini1 (changes requested):
Solid test coverage for the four capability areas from #2475. The tests correctly isolate external dependencies with monkeypatch and use the Dummy class to avoid network calls. One blocker: the toxic/context_relevance tests hardcode label strings that don't match the real model outputs — the actual methods expect "LABEL_0"/"LABEL_1" format (HuggingFace classifier convention). One should-fix: missing edge-case tests for empty scores lists. **VERDICT: Request changes — 1 blocker, 1 should-fix.**

tests/unit/providers/test_huggingface_capabilities.py
Animesh452: Thanks for the feedback! I checked the current source: _toxic_endpoint at line 595 explicitly checks if label["label"] == "toxic", so the fake label in the test matches what the method actually parses. Happy to update if the source changes, but wanted to flag this before changing it.

tests/unit/providers/test_huggingface_capabilities.py
Animesh452: Same: _context_relevance_endpoint at line 573 checks if label["label"] == "context_relevance" directly. The test label matches the actual parsing logic in the current source.

tests/unit/providers/test_huggingface_capabilities.py
Animesh452: Good catch. Adding this now.

tests/unit/providers/test_huggingface_capabilities.py
joshreini1: thanks for checking and sorry for the false alarm!

tests/unit/providers/test_huggingface_capabilities.py
joshreini1: thanks for checking and sorry for the false alarm!

joshreini1 (approved):
Thanks for the quick updates and fixes!

PR #2536, opened 2026-06-09
test: add Bedrock provider capability tests

OrionArchitekton:
## What

Adds `tests/unit/providers/test_bedrock_capabilities.py`, closing #2473. The Bedrock provider lacked capability tests (unlike openai/litellm/langchain/google). This mirrors the existing `test_openai_capabilities.py` pattern but exercises Bedrock's actual surface.

## Coverage

`_create_chat_completion` is fully mocked at the boto3 seam (`endpoint.client.invoke_model`), so no AWS account or credentials are needed:

- Per-model-family request-body routing **and** response parsing — amazon / cohere / ai21 / mistral / meta (parametrized), plus a dedicated amazon `textGenerationConfig` check and the anthropic system/messages + `anthropic_version` shape.
- The anthropic path's messages-required `ValueError`.
- The unsupported-family `NotImplementedError`.
- Message concatenation into `inputText`.
- The default model id (`amazon.nova-lite-v1:0`).

## Verification

- `TEST_OPTIONAL=1 pytest tests/unit/providers/test_bedrock_capabilities.py` -> 12 passed.
- `ruff check` + `ruff format --check` clean (ruff 0.5.5, matching the pre-commit pin).

Closes #2473.

PR #2537, opened 2026-06-09
test: add Cortex provider capability tests

### LINKED ISSUE #2474: Add Cortex Provider Capability Tests
## Description

Same gap as Bedrock — the Cortex provider (`src/providers/cortex/`) has no unit tests in `tests/unit/providers/`.

## Proposed Change

Create `tests/unit/providers/test_cortex_capabilities.py` covering:

- Model name handling
- `_create_chat_completion` behavior (mocked)
- Snowflake connection parameter forwarding
- Error handling

## Reference Pattern

Follow the structure of `tests/unit/providers/test_openai_capabilities.py`.

## Running Tests

```bash
TEST_OPTIONAL=true poetry run pytest tests/unit/providers/test_cortex_capabilities.py -v
```

OrionArchitekton:
## What

Adds `tests/unit/providers/test_cortex_capabilities.py`, closing #2474. The Cortex provider lacked capability tests (OpenAI/Bedrock and others have them). This mirrors the existing `test_openai_capabilities.py` pattern but exercises Cortex's actual surface.

## Coverage

The Snowflake `complete()` call is monkeypatched at the provider module, and the required `snowpark_session` is a no-op `Session` subclass — so no Snowflake account, connection, or credentials are needed:

- Default model engine (`llama3.3-70b`).
- `messages` forwarded to `complete()` with the resolved model, the session, and `stream=False`.
- An explicit `model_engine` override.
- A bare `prompt` wrapped into a single system message.
- The prompt-or-messages-required `ValueError`.
- `response_format` parse-success (valid JSON → pydantic model) and parse-failure (raw string returned rather than raising).

## Verification

- `TEST_OPTIONAL=1 pytest tests/unit/providers/test_cortex_capabilities.py` → 7 passed (python 3.11, snowflake-ml installed).
- `ruff check` + `ruff format --check` clean (ruff 0.5.5, matching the pre-commit pin).

Closes #2474.


tests/unit/providers/test_bedrock_capabilities.py
```diff
@@ -0,0 +1,204 @@
+# pyright: reportMissingImports=false, reportMissingModuleSource=false
```
joshreini1: The bedrock test changes are out of scope for this PR. Move them to a separate PR.

tests/unit/providers/test_cortex_capabilities.py
```diff
@@ -0,0 +1,136 @@
...
+    from trulens.providers.cortex import (
+        provider as cortex_provider,  # type: ignore[import-not-found]
+    )
+
+    calls: List[Dict[str, Any]] = []
+
+    def fake_complete(*, model, prompt, options, session, stream, timeout):
+        calls.append({
+            "model": model,
+            "prompt": prompt,
```
joshreini1, line 37: `timeout` is not a parameter of `snowflake.cortex.complete()` — the public signature is `complete(model, prompt, options=None, session=None, stream=False)`. Including `timeout` here means `fake_complete` will never match the real call signature, so monkeypatching will silently fail or the real `complete` will be invoked. Remove `timeout` from the `fake_complete` signature and from the `calls.append` dict.

tests/unit/providers/test_cortex_capabilities.py
```diff
@@ -0,0 +1,136 @@
...
+
+@pytest.mark.optional
+def test_default_model_engine():
+    """No model_engine -> the documented llama3.3-70b default."""
+    from trulens.providers.cortex import (
+        Cortex,  # type: ignore[import-not-found]
+    )
+
+    assert Cortex.DEFAULT_MODEL_ENGINE == "llama3.3-70b"
+
```
joshreini1: Delete this entire `test_default_model_engine` test. It just encodes the current value of `DEFAULT_MODEL_ENGINE` as a constant — every time the default changes, someone has to remember to update this test too, which is pure churn with no real safety benefit. The behavioral tests below already exercise whatever default is active.

joshreini1 (changes requested):
Cortex tests are clean and follow the established pattern. The bedrock file is out of scope and must be moved to its own PR. One issue in the Cortex tests needs fixing: the spurious `timeout` parameter in `fake_complete`. Also drop the `test_default_model_engine` test — it just encodes a constant that will need updating every time the default changes, adding churn with no real safety benefit.

tests/unit/providers/test_bedrock_capabilities.py
OrionArchitekton: Agreed, removed from this PR. The identical file lands on its own in #2536, so nothing is
lost by dropping it here. This PR is now Cortex-only.

tests/unit/providers/test_cortex_capabilities.py
OrionArchitekton: Agreed, dropped. `test_messages_forwarded_to_complete` already asserts that whatever
default is active flows through to `complete()`, so the constant assertion was redundant
churn on top of it.

tests/unit/providers/test_cortex_capabilities.py
OrionArchitekton: I dug into this one and I do not think the premise holds, so I have left `timeout` in
place. Three checks:

**1. `snowflake.cortex.complete()` does accept `timeout`.** From `snowflake-ml-python`,
`snowflake/cortex/_complete.py`:

```python
def complete(
    model: Union[str, snowpark.Column],
    prompt: Union[str, list[ConversationMessage], snowpark.Column],
    *,
    options: Optional[CompleteOptions] = None,
    session: Optional[snowpark.Session] = None,
    stream: bool = False,
    timeout: Optional[float] = None,
    deadline: Optional[float] = None,
) -> Union[str, Iterator[str], snowpark.Column]:
```

The docstring documents it: `timeout (float): Timeout in seconds to retry failed REST
requests.` I checked this at tag `1.7.2` (the floor of this package's
`snowflake-ml-python = "^1.7.2"` constraint) and at `1.48.0` (what resolves today), and
the kwarg is present in both, so it holds across the whole supported range.

**2. TruLens itself passes it.** `src/providers/cortex/trulens/providers/cortex/provider.py`
L165-172:

```python
completion_res: str = complete(
    model=model,
    prompt=messages,
    options=options,
    session=self.snowpark_session,
    stream=False,
    timeout=self.retry_timeout,
)
```

**3. The change fails the suite.** The test monkeypatches the module-level `complete` that
the provider imported, so `fake_complete` has to match that call site rather than the
library's documented signature. I applied the requested edit on a scratch copy and ran it:

```
E  TypeError: _make_provider.<locals>.fake_complete() got an unexpected keyword argument 'timeout'
   .../trulens/providers/cortex/provider.py:165: TypeError
5 failed, 1 passed
```

With `timeout` kept, all 6 pass.

Happy to change it if you are seeing a `snowflake-ml-python` version where the kwarg is
gone. That would also mean the provider call site above needs fixing, which would be a
separate PR rather than a test change.

joshreini1 (approved):
Thanks for digging into this. Agree with your assessment on the timeout kwarg.

Issue #2496, opened 2026-05-13
Integration Test: Score Parsing Through Full Pipeline


joshreini1:
## Description

`test_feedback_score_generation.py` tests the score parser (`re_0_10_rating`) with hardcoded strings, but it doesn't test parsing through the full feedback pipeline. We should verify that the full pipeline handles various LLM response formats — including edge cases that real LLMs produce.

## Proposed Change

Create `tests/integration/test_score_parsing_pipeline.py` that:

1. Creates multiple `MockLLMProvider` variants, each returning a different realistic response format:
   - Plain number: `"2"`
   - With explanation: `"The relevance is moderate. Score: 2"`
   - JSON format: `'{"criteria": "...", "supporting_evidence": "...", "score": 2}'`
   - With markdown: `` "```json\n{\"score\": 2}\n```" ``
   - Whitespace variations: `"\n  Score: 2  \n"`
   - Float format: `"2.5"`
2. For each mock variant, runs a feedback function through the full pipeline
3. Asserts:
   - The score parser extracts a valid float without raising `ParseError`
   - The extracted score is in the valid range [0.0, 1.0] after normalization
   - The full `Feedback` pipeline completes without error

## Files to Reference

- `src/feedback/trulens/feedback/generated.py` — `re_0_10_rating` and score extraction
- `src/feedback/trulens/feedback/llm_provider.py` — `generate_score()`, `generate_score_and_reasons()`
- `tests/unit/test_feedback_score_generation.py` — existing hardcoded tests

## Running Tests

```bash
poetry run pytest tests/integration/test_score_parsing_pipeline.py -v
```

RyanFlowerYes:
Hi @joshreini1 ,  I'd like to take this one.
I've got a draft of tests/integration/test_score_parsing_pipeline.py covering the formats in the issue via a small MockLLMProvider, plus a couple of negative cases (no-number / out-of-range) to confirm ParseError still fires.
One question before I open the PR: is there an existing mock provider or test fixture I should build on, or is a new MockLLMProvider stubbing _create_chat_completion the right approach here?
Starting local repro this week,  thanks!

RyanFlowerYes:
Opened a PR for this:
https://github.com/truera/trulens/pull/2554

Adds integration coverage for the score parsing pipeline by exercising the real provider parsing/normalization path through generate_score, generate_score_and_reasons, and a full feedback-method execution path via context_relevance.

[...]

PR #2554, opened 2026-06-14
Add integration tests for feedback score parsing pipeline

RyanFlowerYes:
## Summary

Adds integration coverage for score parsing through the full feedback provider pipeline.

The existing score parser tests cover raw rating extraction directly with hardcoded strings. This PR adds integration tests that exercise the real `generate_score` and `generate_score_and_reasons` parsing and normalization paths by overriding only `_create_chat_completion` in a mock provider.

## Coverage

* Plain numeric responses
* Responses with explanatory text
* Markdown-fenced JSON-like responses
* Whitespace variations
* Decimal-style responses
* Prose responses containing a score
* Custom score ranges
* Scale boundaries
* Parse errors for missing or out-of-range scores
* Structured-output schema objects
* `generate_score_and_reasons`
* Full feedback-method execution via `context_relevance`

## Issue surfaced

While adding integration coverage, this test suite exposes an inconsistency in `generate_score` for structured JSON responses.

The same structured JSON payload is normalized correctly by `generate_score_and_reasons`, but `generate_score` returns a `(score, reason)` tuple instead of a plain normalized float despite its `-> float` return annotation.

This behavior is captured as a strict `xfail`, documenting the current behavior without breaking CI.

## Validation

```bash
poetry run pytest tests/integration/test_score_parsing_pipeline.py -v
```

Result:

* 26 passed
* 1 xfailed

Addresses #2496.


joshreini1 (approved):
This is a fantastic contribution to harden our metric reliability. Thanks for adding @RyanFlowerYes

sfc-gh-jreini:
Resolved the merge conflict by merging latest `main`.

Heads up: `main` merged an integration test at the same path (`tests/integration/test_score_parsing_pipeline.py`) via #2568, which caused an add/add conflict. That suite focuses on the three response branches (JSON dict/list averaging, plain-string regex, malformed fall-through).

Your suite is complementary rather than redundant — it covers normalization across text shapes, decimal/custom-range/boundary handling, structured `BaseFeedbackResponse`/`ChainOfThoughtResponse` objects, `generate_score_and_reasons`, `context_relevance`, and a strict-xfail documenting the `generate_score` tuple-vs-float inconsistency. So rather than drop either, I kept #2568's file in place and moved your broader suite to `tests/integration/test_score_parsing_normalization.py`.

Verified locally on latest main: `26 passed, 1 xfailed`. Thanks for the thorough coverage!

PR #2564, opened 2026-06-19
feat: add CriteriaABTest for comparing judge configurations

### LINKED ISSUE #2504: Criteria A/B Testing Utility
## Description

TruLens feedback functions accept `criteria` and `additional_instructions` parameters to customize judge behavior. But there's no easy way to compare two versions of criteria against the same golden set to see which one aligns better with human judgment.

## Proposed Change

Add a `CriteriaABTest` utility that:

1. Takes two versions of a feedback function config (e.g., different `criteria` strings, different `additional_instructions`, or different temperatures)
2. Runs both against the same golden set
3. Computes alignment metrics for each (MAE, Kendall's tau, Spearman, Brier score) using `GroundTruthAggregator`
4. Produces a comparison report:
   - Side-by-side metrics table
   - Per-example score comparison (where do they disagree most?)
   - Statistical significance test (paired t-test or Wilcoxon signed-rank) on the score differences
   - Winner recommendation

## Usage Example

```python
from trulens.benchmark import CriteriaABTest

provider = OpenAI()

test = CriteriaABTest(
    golden_set=my_golden_set,
    variant_a={"fn": provider.relevance, "name": "Default criteria"},
    variant_b={"fn": provider.relevance, "kwargs": {"criteria": "Score strictly..."}, "name": "Strict criteria"},
)
report = test.run()
report.print_comparison()
```

## Files to Reference

- `src/feedback/trulens/feedback/groundtruth.py` — `GroundTruthAggregator` metrics
- `src/benchmark/trulens/benchmark/benchmark_frameworks/tru_benchmark_experiment.py` — benchmark runner
- `src/feedback/trulens/feedback/llm_provider.py` — `criteria` and `additional_instructions` params

thunderstornX:
## Summary

Closes #2504.

TruLens feedback functions accept `criteria` and `additional_instructions` to tune judge behaviour, but there is no easy way to tell whether a new criteria string actually agrees with human judgement better than the old one. `CriteriaABTest` runs two configurations of a feedback function over the same golden set and reports which one aligns better.

## Implementation

`src/benchmark/trulens/benchmark/criteria_ab_test.py`:

- `CriteriaABTest(golden_set, variant_a, variant_b, ...)` runs each variant (a `{"fn", "name", "kwargs"}` dict, where `kwargs` is forwarded to `fn`, e.g. `criteria=...`) over each golden row, skipping any row where either variant errors so they stay aligned, and returns a `CriteriaABTestReport`.
- `CriteriaABTestReport` computes:
  - **side-by-side metrics** vs the ground truth: MAE, Spearman, Kendall's tau, Brier
  - **top disagreements** — the examples where the two variants differ most
  - **significance** — mean score difference and a paired **sign-flip permutation** p-value
  - **winner** — the variant with lower MAE vs ground truth
  - `print_comparison()`

Metrics and the significance test are computed with numpy (no scipy dependency) to keep the benchmark utility dependency-light.

## Testing

- `tests/unit/test_criteria_ab_test.py`: 8 unit tests covering the metrics, winner, the permutation significance test (identical → p=1.0, consistent shift → significant), top disagreements, the run loop with kwargs forwarding, and the guards. No model or API key required; guarded with `pytest.importorskip` since `trulens-benchmark` is optional.
- `ruff format` + `ruff check` clean against the repo config.
- Verified end-to-end: ran default vs strict-`criteria` `relevance` over `answer_relevance_golden_set` against a live OpenAI-compatible model; the report produced the side-by-side metrics and correctly reported the difference as not statistically significant (only one example diverged).

## Docs

Adds a usage guide at `docs/component_guides/evaluation/criteria_ab_test.md` (registered in the nav). The API reference picks the module up automatically from its docstrings.

### DIFF
--- a/docs/component_guides/evaluation/criteria_ab_test.md
+++ b/docs/component_guides/evaluation/criteria_ab_test.md
@@ -1,0 +1,66 @@
+# Criteria A/B Testing
+
+TruLens feedback functions accept `criteria` and `additional_instructions` to
+tune judge behaviour, but there is no easy way to tell whether a new criteria
+string actually agrees with human judgement better than the old one.
+
+`CriteriaABTest` runs two configurations of a feedback function over the same
+golden set and reports which one aligns better: side-by-side MAE / Spearman /
+Kendall / Brier against the ground truth, the examples where the two disagree
+most, a paired significance test (a sign-flip permutation test, no scipy
+dependency) on the score differences, and a winner.
+
+!!! Initial Setup
+
+    ```bash
+    pip install trulens-benchmark
+    pip install trulens-providers-openai
+    ```
+
+!!! example
+
+    ```python
+    import os
+    from trulens.benchmark.criteria_ab_test import CriteriaABTest
+    from trulens.providers.openai import OpenAI
+
+    os.environ["OPENAI_API_KEY"] = "<replace-with-your-api-key>"
+    provider = OpenAI()
+
+    test = CriteriaABTest(
+        golden_set=my_golden_set,
+        variant_a={"fn": provider.relevance, "name": "default"},
+        variant_b={
+            "fn": provider.relevance,
+            "kwargs": {"criteria": "Be strict; a high score only if fully relevant."},
+            "name": "strict",
+        },
+    )
+    report = test.run()
+    report.print_comparison()
+    ```
+
+`print_comparison()` produces:
+
+```
+Criteria A/B Test
+========================================
+A = default    B = strict    n = 8
+
+  metric           default        strict
+  mae                0.383         0.425
+  spearman           0.438         0.319
+  kendall            0.500         0.455
+  brier              0.200         0.244
+
+Mean difference (A - B): +0.042  (permutation p = 1.000)
+
+Largest disagreements:
+  +0.33  A=0.33 B=0.00  How does the social structure of a lion pride im...
+
+Winner: default (lower MAE vs ground truth; difference not significant at alpha=0.05).
+```
+
+Use this when iterating on a judge's `criteria` or `additional_instructions`, to
+confirm a change actually improves alignment with human labels rather than just
+shifting scores around.
--- a/src/benchmark/trulens/benchmark/criteria_ab_test.py
+++ b/src/benchmark/trulens/benchmark/criteria_ab_test.py
@@ -9,8 +9,11 @@ one aligns better: side-by-side MAE / Spearman / Kendall / Brier against the
 ground truth, the examples where the two disagree most, a paired significance
 test on the score differences, and a winner.
 
-Metrics are computed with numpy (no scipy dependency); significance uses a
-sign-flip permutation test so the utility stays dependency-light.
+Spearman and Kendall reuse
+[GroundTruthAggregator][trulens.feedback.groundtruth.GroundTruthAggregator] so
+there is a single source of truth for those metrics; the paired sign-flip
+permutation significance test lives alongside it in
+[groundtruth][trulens.feedback.groundtruth].
 
 Example:
     ```python
@@ -36,11 +39,10 @@ import logging
 from typing import Any, Callable, Dict, List, Optional, Tuple
 
 import numpy as np
+from trulens.feedback import groundtruth as feedback_groundtruth
 
 log = logging.getLogger(__name__)
 
-# Number of sign-flip permutations for the significance test.
-PERMUTATIONS = 10000
 # Significance level below which the two variants are called different.
 SIGNIFICANCE_ALPHA = 0.05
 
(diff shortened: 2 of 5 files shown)


src/benchmark/trulens/benchmark/criteria_ab_test.py
```diff
@@ -0,0 +1,332 @@
...
+
+
+def _spearman(a: np.ndarray, b: np.ndarray) -> float:
+    ranks_a, ranks_b = _rankdata(a), _rankdata(b)
+    if np.std(ranks_a) == 0 or np.std(ranks_b) == 0:
+        return float("nan")
+    return float(np.corrcoef(ranks_a, ranks_b)[0, 1])
+
+
+def _kendall_tau(a: np.ndarray, b: np.ndarray) -> float:
```
joshreini1: don't reimplement, just call: https://github.com/truera/trulens/blob/b741377470a7f33e432e277bab558b0cb8710ec5/src/feedback/trulens/feedback/groundtruth.py#L823

src/benchmark/trulens/benchmark/criteria_ab_test.py
```diff
@@ -0,0 +1,332 @@
...
+    ranks[order] = np.arange(1, len(values) + 1)
+    _, inverse, counts = np.unique(
+        values, return_inverse=True, return_counts=True
+    )
+    rank_sums = np.zeros(len(counts))
+    np.add.at(rank_sums, inverse, ranks)
+    return (rank_sums / counts)[inverse]
+
+
+def _spearman(a: np.ndarray, b: np.ndarray) -> float:
```
joshreini1: don't reimplement, just call: https://github.com/truera/trulens/blob/b741377470a7f33e432e277bab558b0cb8710ec5/src/feedback/trulens/feedback/groundtruth.py#L841

src/benchmark/trulens/benchmark/criteria_ab_test.py
```diff
@@ -0,0 +1,332 @@
...
+            sign = np.sign(a[i] - a[j]) * np.sign(b[i] - b[j])
+            if sign > 0:
+                concordant += 1
+            elif sign < 0:
+                discordant += 1
+    total = concordant + discordant
+    return float((concordant - discordant) / total) if total else float("nan")
+
+
+def _permutation_pvalue(diffs: np.ndarray, seed: int = 0) -> float:
```
joshreini1: add implementation to ground truth file and just call it where needed  https://github.com/truera/trulens/blob/b741377470a7f33e432e277bab558b0cb8710ec5/src/feedback/trulens/feedback/groundtruth.py

docs/component_guides/evaluation/criteria_ab_test.md
```diff
@@ -0,0 +1,66 @@
+# Criteria A/B Testing
```
joshreini1, line 1: this fills a really important gap to aligning trulens metrics. Thanks for congtributing!

joshreini1 (changes requested):
Thanks for contributing! form factor looks great. Requested changes to use existing metric implementations and move one to the ground truth file. Please re-request my review when you've put these changes in

src/benchmark/trulens/benchmark/criteria_ab_test.py
thunderstornX: Done. Removed the local `_spearman`; this now calls `GroundTruthAggregator.spearman_correlation` directly. Thanks for the pointer.

src/benchmark/trulens/benchmark/criteria_ab_test.py
thunderstornX: Done, same here. Now using `GroundTruthAggregator.kendall_tau` instead of the local implementation.

src/benchmark/trulens/benchmark/criteria_ab_test.py
thunderstornX: Moved this into `groundtruth.py` as `paired_permutation_pvalue` and call it from here now, so it's reusable across the benchmark. Good call.

docs/component_guides/evaluation/criteria_ab_test.md
thunderstornX: Thank you, that genuinely means a lot. Consistent judge-alignment metrics were exactly the goal here.

thunderstornX:
@joshreini1 thanks for the detailed review, this is exactly the kind of guidance I was hoping for. Pushed the changes:

- Dropped the local `_spearman` and `_kendall_tau`; `CriteriaABTest` now calls `GroundTruthAggregator.spearman_correlation` and `.kendall_tau`, so those metrics have a single source of truth.
- Moved the paired sign-flip permutation significance test into `groundtruth.py` as `paired_permutation_pvalue` and call it from here.
- Unit tests green (importorskip-guarded) and ruff clean on the 0.5.5 pin.

Note: this adds a `trulens.feedback` import to the benchmark module (the same use `dataset_preprocessing` already has). Happy to also add `trulens-feedback` to the `trulens-benchmark` dependencies if you'd like that made explicit.

Re-requesting your review whenever you have a moment. Thanks again.

joshreini1 (approved):
LGTM - thanks for the quick updates.

PR #2568, opened 2026-06-19
Add integration test for score parsing through full pipeline (#2496)

### LINKED ISSUE #2496: Integration Test: Score Parsing Through Full Pipeline

maitraymukeshkumarmodi-aiml:
# Description

Adds an integration test verifying that `LLMProvider.generate_score` (in
`trulens.feedback.llm_provider`) correctly parses LLM scoring responses
through its full real pipeline, rather than testing the regex parser
(`re_0_10_rating` / `re_configured_rating`) in isolation as the existing
unit test (`tests/unit/test_feedback_score_generation.py`) does.

Closes #2496.

While building this, I found that `generate_score` actually has three
distinct response-handling branches, not one:

1. **JSON-direct**: if the raw response is valid JSON containing a `score`
   key (dict) or a list of such dicts, the score is extracted directly —
   no regex involved. This branch returns a `(score, reason_dict)` tuple,
   unlike the other branches.
2. **Structured object**: if the response is already a
   `BaseFeedbackResponse` instance, `.score` is read directly.
3. **Regex fallback**: if the response is a plain string that isn't valid
   "scored" JSON, it falls through to `re_configured_rating`.

The test covers all three branches, the transitions between them, and two
edge cases I didn't expect:

- JSON-shaped strings that fail to parse as valid JSON (e.g. truncated
  JSON) correctly fall through to the regex branch rather than crashing.
- JSON-shaped strings that parse successfully but lack a `score` key
  (e.g. `{"rating": 7}`) *also* fall through to the regex branch — and
  the regex fallback is permissive enough to still extract a bare number
  from the string. I added a case with no extractable number at all to
  confirm `ParseError` is still raised when nothing parseable is found.

- `MockLLMProvider` in this test constructs a real
  `trulens.core.feedback.endpoint.Endpoint(name="mock-endpoint")` rather
  than a duck-typed stub, since `endpoint` is a strictly-typed Pydantic
  field on `Provider`/`LLMProvider` and rejects non-`Endpoint` objects at
  construction time.
- Only `_create_chat_completion` is overridden to return canned
  responses per test case; `generate_score` itself runs unmodified, so
  the test exercises the real parsing logic end-to-end.
- Worth a maintainer's eyes on the JSON-branch vs. regex-branch return
  type asymmetry (`Tuple[float, Dict]` vs. bare `float`) — not fixed
  here since it's out of scope for a test-only PR, but it's a real
  inconsistency callers should be aware of.

## Type of change

  not work as expected)
- [x] New Tests

### DIFF
--- a/tests/integration/test_score_parsing_pipeline.py
+++ b/tests/integration/test_score_parsing_pipeline.py
@@ -1,0 +1,140 @@
+"""Integration test verifying trulens.feedback.llm_provider.LLMProvider.generate_score
+parses scores correctly end-to-end through all three response branches:
+JSON, structured BaseFeedbackResponse, and plain-string regex fallback
+(trulens.feedback.generated.re_configured_rating).
+
+Complements tests/unit/test_feedback_score_generation.py, which tests
+re_0_10_rating in isolation. This test exercises the real generate_score
+control flow, including JSON-parse-first behavior and fallthrough on
+malformed/non-scored JSON.
+"""
+
+from typing import ClassVar, Optional
+
+import pytest
+from trulens.feedback import generated as feedback_generated
+from trulens.feedback import llm_provider
+
+
+class _StubEndpoint:
+    """Minimal real endpoint stub. Only implements what run_in_pace needs:
+    pace_me() and retries. No network, no rate limiting.
+    """
+
+    def __init__(self):
+        self.retries = 0
+
+    def pace_me(self):
+        return None
+
+    def run_in_pace(self, func, *args, **kwargs):
+        return func(*args, **kwargs)
+
+
+class MockLLMProvider(llm_provider.LLMProvider):
+    """LLMProvider with a real (stub) endpoint, so generate_score's actual
+    parsing branches execute rather than being bypassed. Only
+    _create_chat_completion is overridden, to return a canned response per
+    test case.
+    """
+
+    model_config: ClassVar[dict] = {"extra": "allow"}
+
+    canned_response: Optional[object] = None
+
+    def __init__(self, **kwargs):
+        super().__init__(
+            endpoint=_StubEndpoint(),
+            model_engine="mock-model",
+            **kwargs,
+        )
+
+    def _create_chat_completion(
+        self,
+        prompt: Optional[str] = None,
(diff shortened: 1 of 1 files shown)


joshreini1 (approved):
Approving this as the canonical implementation for #2496.

Why this one:
- Only adds the requested `tests/integration/test_score_parsing_pipeline.py` file.
- Exercises the real `generate_score` path across JSON dict/list responses, plain-string regex fallback, malformed JSON fallback, and unparseable responses.
- CI is green.

I also validated it locally after setting up a Python 3.11 virtualenv with local editable packages:

```bash
.venv/bin/python -m pytest tests/integration/test_score_parsing_pipeline.py -q
# 11 passed, 1 warning
```

PR #2575, opened 2026-06-23
feat(feedback): add CitationAccuracy evaluator for RAG

### LINKED ISSUE #2408: Add Citation / Attribution Evaluator
## Summary
A common RAG failure mode is generating answers without proper citations. No existing template evaluates citation quality.

## What
Add a `CitationAccuracy` template to [`src/feedback/trulens/feedback/templates/rag.py`](https://github.com/truera/trulens/blob/main/src/feedback/trulens/feedback/templates/rag.py) that checks whether citations in a response map to actual retrieved contexts. Wire up a corresponding method in [`src/feedback/trulens/feedback/llm_provider.py`](https://github.com/truera/trulens/blob/main/src/feedback/trulens/feedback/llm_provider.py).

## Difficulty
Easy-Medium

rhiannalitchfield:
## Summary

Adds a `CitationAccuracy` feedback evaluator that scores whether the citations in a RAG response are actually supported by the retrieved context. Closes #2408.

- **`src/feedback/trulens/feedback/templates/rag.py`** — new `CitationAccuracy` template (`Semantics, WithPrompt, CriteriaOutputSpaceMixin`, 0–3 Likert), following the existing `ContextRelevance` pattern. Exported via `__all__`.
- **`src/feedback/trulens/feedback/llm_provider.py`** — `citation_accuracy` (returns a float) and `citation_accuracy_with_cot_reasons` (returns score + reasons), mirroring the signature conventions of the other RAG metrics (criteria / additional_instructions / examples overrides, configurable score range, deprecated-kwarg handling).

## Why

A common RAG failure mode is generating answers whose citations are fabricated or point to passages that do not support the cited claim. No existing template evaluated citation quality. The criteria are explicitly scoped to judge citation accuracy only, not overall response quality or relevance, so it composes cleanly alongside the existing RAG triad metrics.

## Testing

- `TEST_OPTIONAL=true pytest tests/unit/test_feedback_templates_validation.py tests/unit/test_feedback_templates_exports.py -q` — passes. The validation suite auto-discovers and validates the new template (non-empty prompts, valid output space, criteria, runtime format fields, Pydantic instantiation).
- `ruff` (v0.5.5, repo-pinned) format `--check` and lint are clean on both changed files.

## Notes

I included both the base and CoT variants to match the convention of the other RAG metrics (`context_relevance` / `groundedness`). Happy to drop the CoT variant or adjust the prompt wording if you'd prefer a smaller surface.

### DIFF
--- a/src/feedback/trulens/feedback/llm_provider.py
+++ b/src/feedback/trulens/feedback/llm_provider.py
@@ -739,6 +739,176 @@ class LLMProvider(core_provider.Provider):
             temperature=temperature,
         )
 
+    def citation_accuracy(
+        self,
+        response: str,
+        context: str,
+        criteria: Optional[str] = None,
+        additional_instructions: Optional[str] = None,
+        examples: Optional[List[str]] = None,
+        min_score_val: int = 0,
+        max_score_val: int = 3,
+        temperature: float = 0.0,
+        **kwargs,
+    ) -> float:
+        """
+        Uses chat completion model. A function that completes a template to
+        check whether the citations in a response are supported by the
+        retrieved context.
+
+        Example:
+            ```python
+            from trulens.core import Metric, Selector
+            feedback = Metric(
+                implementation=provider.citation_accuracy,
+                name="Citation Accuracy",
+                criteria=criteria,
+                additional_instructions=additional_instructions,
+                examples=examples,
+                selectors={
+                    "response": Selector.select_record_output(),
+                    "context": Selector.select_context(
+                        collect_list=True
+                    ),
+                },
+                agg=np.mean,
+            )
+            ```
+
+        Args:
+            response (str): The response containing citations to evaluate.
+            context (str): The retrieved context the citations should map to.
+            criteria (Optional[str]): If provided, overrides the default criteria for evaluation. Defaults to None.
+            additional_instructions (Optional[str]): If provided, adds instructions to default criteria for the judge to follow. Defaults to None.
+            examples (Optional[List[str]]): Optional few-shot examples to guide the evaluation. Defaults to None.
+            min_score_val (int): The minimum score value. Defaults to 0.
+            max_score_val (int): The maximum score value. Defaults to 3.
+            temperature (float): The temperature for the LLM response, which might have impact on the confidence level of the evaluation. Defaults to 0.0.
+
+        Returns:
+            float: A value between 0.0 (citations inaccurate) and 1.0 (citations accurate).
+        """
+        # Handle deprecated parameter names
+        additional_instructions = deprecation_utils.handle_deprecated_kwarg(
+            kwargs,
+            "custom_instructions",
+            "additional_instructions",
+            additional_instructions,
+        )
+
+        output_space = self._determine_output_space(
+            min_score_val, max_score_val
+        )
+
+        system_prompt = templates_rag.CitationAccuracy.generate_system_prompt(
+            min_score=min_score_val,
+            max_score=max_score_val,
+            criteria=criteria,
+            additional_instructions=additional_instructions,
+            examples=examples,
+            output_space=output_space,
+        )
+
+        return self.generate_score(
+            system_prompt=system_prompt,
+            user_prompt=str.format(
+                templates_rag.CitationAccuracy.user_prompt,
+                response=response,
+                context=context,
+            ),
+            min_score_val=min_score_val,
+            max_score_val=max_score_val,
+            temperature=temperature,
+        )
+
+    def citation_accuracy_with_cot_reasons(
+        self,
+        response: str,
+        context: str,
+        criteria: Optional[str] = None,
+        additional_instructions: Optional[str] = None,
+        examples: Optional[List[str]] = None,
+        min_score_val: int = 0,
+        max_score_val: int = 3,
+        temperature: float = 0.0,
+        **kwargs,
+    ) -> Tuple[float, Dict]:
+        """
+        Uses chat completion model. A function that completes a template to
+        check whether the citations in a response are supported by the
+        retrieved context. Also uses chain of thought methodology and emits
+        the reasons.
+
+        Example:
+            ```python
+            from trulens.core import Metric, Selector
+            feedback = Metric(
+                implementation=provider.citation_accuracy_with_cot_reasons,
+                name="Citation Accuracy",
+                criteria=criteria,
+                additional_instructions=additional_instructions,
+                examples=examples,
+                selectors={
+                    "response": Selector.select_record_output(),
+                    "context": Selector.select_context(
+                        collect_list=True
+                    ),
+                },
+                agg=np.mean,
+            )
+            ```
+
+        Args:
+            response (str): The response containing citations to evaluate.
+            context (str): The retrieved context the citations should map to.
+            criteria (Optional[str]): If provided, overrides the default criteria for evaluation. Defaults to None.
+            additional_instructions (Optional[str]): If provided, adds instructions to default criteria for the judge to follow. Defaults to None.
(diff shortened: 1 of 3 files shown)


joshreini1 (approved):
Good new evaluator -- CitationAccuracy fills a real gap and the implementation follows the established pattern well (Semantics + WithPrompt + CriteriaOutputSpaceMixin, criteria_template with min/max interpolation, both plain and CoT variants). A few notes for consideration:

1. The criteria_template has leading indentation on each line that will appear in the rendered prompt -- consider using cleandoc() on it or stripping like the other templates do.

2. The context param assumes a single string, but RAG retrievers typically return a list of chunks. The docstring example shows collect_list=True on the selector but the function signature takes `context: str`. Would be worth either accepting List[str] directly or documenting the expected join behavior.

These are polish items rather than blockers. Overall this is a clean, well-structured addition. Thanks!

joshreini1 (approved):
Good new evaluator -- CitationAccuracy fills a real gap and the implementation follows the established pattern well (Semantics + WithPrompt + CriteriaOutputSpaceMixin, criteria_template with min/max interpolation, both plain and CoT variants). Two notes:

1. The criteria_template has leading indentation on each line that will appear in the rendered prompt -- consider using cleandoc() on it or stripping like other templates do.

2. The context param assumes a single string, but RAG retrievers typically return a list of chunks. The docstring example shows collect_list=True on the selector but the signature takes context: str. Worth either accepting List[str] directly or documenting the expected join behavior.

These are polish items rather than blockers. Overall a clean, well-structured addition. Thanks!

joshreini1 (changes requested):
Updating my earlier approval -- after checking main I wanted to clear up the differece between the recently added `CitationAttribution` that covers closely related ground. Before merging I'd like you to address the overlap.

**What already exists in main: `citation_attribution` / `CitationAttribution`**

The existing evaluator also checks whether citations in a RAG response are supported by retrieved passages. Key differences in its design:
- Expects explicit `[N]` numbered citation markers in the response (e.g. "Paris is the capital [1]")
- Uses `_number_citation_sources()` to number a list of passages so markers resolve correctly
- Includes `question` as an input (gives the judge context for what's being asked)
- Binary 0-1 output (misattribution = hard fail)
- Explicitly does NOT penalize claims that carry no citation marker

**How `CitationAccuracy` differs**
- Format-agnostic -- no `[N]` markers required
- Graded Likert 0-3 output
- Penalizes missing citations ("missing where support exists should score {min_score}")
- Takes raw `context: str` instead of a numbered source list
- No `question` parameter

**What I need clarified before this merges**

1. **Is there a concrete use case where `CitationAccuracy` applies but `CitationAttribution` does not?** If yes, please add a note to the docstring explaining when to use each (e.g. "Use `citation_attribution` when your RAG pipeline produces `[N]`-style markers; use `citation_accuracy` when citations are inline or implicit and you want graded feedback including missing-citation penalties").

2. **The `context: str` signature vs `collect_list=True`.** The docstring example shows `Selector.select_context(collect_list=True)` which returns a list, but the method signature takes `context: str`. How is the list joined before it reaches the evaluator? This should either be handled inside the method (like `citation_attribution` does with `_number_citation_sources`) or documented explicitly.

3. **Missing citation penalization is a significant behavioral difference** from `CitationAttribution`. If that's intentional, it's worth calling out explicitly in the docstring -- users choosing between the two should understand this tradeoff.

Happy to re-approve once these are addressed. The implementation itself is solid; this is just about positioning it correctly relative to the existing API.

rhiannalitchfield:
Thanks for the careful look — rebased onto current main so `CitationAttribution` and this now sit side by side.

## 1. Concrete cases where `CitationAccuracy` applies but `CitationAttribution` does not

Three, and they're what motivated keeping it separate:

- **Non-`[N]` citation formats.** Inline prose attributions ("according to the 2023 annual report..."), footnotes, or URLs. `CitationAttribution`'s prompt instructs the judge to check each `[N]` marker against numbered passage N. With no markers present it doesn't just degrade, it *vacuously passes*: the criteria say a claim carrying no marker isn't a misattribution, so a wholly uncited answer scores 1.
- **Graded signal for regression tracking.** Binary gives no gradient — one bad citation out of ten scores identically to ten out of ten. The 0–3 Likert with `agg=np.mean` gives something you can watch move across prompt revisions.
- **Detecting under-citation.** For pipelines whose requirement is "context-derived claims must be cited," `CitationAttribution` structurally cannot catch a model that simply stops citing.

I've added a class docstring laying out all three axes, with explicit "use X when / use Y when" guidance in both provider docstrings.

## 2. `context: str` vs `collect_list=True`

Fair catch — that was a genuine inconsistency between the example and the signature. The signature is now `Union[str, List[str]]`, and a list is joined inside the method via a new `_join_context_passages` helper.

I deliberately did **not** reuse `_number_citation_sources`: adding `[1]`/`[2]` markers would present the judge a numbered citation format the response isn't actually using, which is the opposite of what a format-agnostic metric wants. That reasoning is documented on the helper and on the `context` arg in both methods.

## 3. Missing-citation penalization

Intentional, and you're right that it deserves to be explicit. It's now called out in the class docstring and both method docstrings as the key tradeoff, with a pointer that `citation_attribution` is the correct choice when under-citation is acceptable and only misattribution matters.

## On the criteria indentation (from your earlier review)

I looked into this and left it as-is. Every template in the repo renders with indented criteria lines — I checked by rendering `system_prompt` for each:

| Template | Indented criteria lines |
|---|---|
| `Groundedness` | 7 (4-space) |
| `ContextRelevance` | 6 (8-space) |
| `Comprehensiveness` | 4 (4-space) |
| `CitationAttribution` | 5 (8-space) |
| `CitationAccuracy` | 6 (8-space) |

`cleandoc()` can't fix it at the call site: it's applied to `system_prompt_template` *before* `criteria` is interpolated, and the zero-indent surrounding lines make the common prefix 0, so nothing gets stripped. Stripping only this one template would make it the odd one out. Happy to do it repo-wide in a separate PR — the clean fix is dedenting `final_criteria` inside `CriteriaOutputSpaceMixin.generate_system_prompt`, so all templates change together and stay consistent.

## Testing

Added `tests/unit/test_citation_accuracy.py` (15 tests) mirroring `test_citation_attribution.py`: the joining helper, prompt construction and delegation for both variants, score-range handling, criteria/instructions/examples overrides, and the deprecated `custom_instructions` kwarg.

- 66 tests pass locally (citation suites plus `test_feedback_templates_validation`, `test_feedback_templates_exports`, `test_feedback_criteria_and_additional_instructions`, `test_feedback_score_generation`).
- `ruff` 0.5.5 format `--check` and lint clean on all three files.

The two failing checks on the previous run were `test_bedrock_capabilities.py::test_model_family_request_and_response` and `::test_messages_are_concatenated_into_input_text` — unrelated to this PR, and fixed on main by #2645. The rebase picks that up.

joshreini1 (approved):
Thanks for the thorough response and rebasing. The three concrete differentiators you laid out are convincing:

- **Non-[N] citation formats**: prose/footnote/URL citations where `citation_attribution` vacuously passes (no markers = no misattribution by its own criteria)
- **Graded signal**: 0-3 scale enabling regression tracking, vs binary pass/fail
- **Missing-citation penalty**: explicitly penalizes uncited claims, which `citation_attribution` deliberately ignores

The docstring already captures this clearly under "Choosing between this and `citation_attribution`", so users have the guidance they need. The template class, provider method, and test coverage all look solid.

One small nit: the `_join_context_passages` helper is a `@staticmethod` on `LLMProvider` but it's really a module-level utility — no compelling reason it needs to be a method. Not a blocker, just a cleanup opportunity.

Approving — the implementation is well-structured and the distinction from the existing evaluator is clear enough to justify both.

PR #2576, opened 2026-06-27
feat(feedback): add citation_attribution feedback function

rsrijith:
## What

Adds two feedback functions on `LLMProvider`: `citation_attribution` and `citation_attribution_with_cot_reasons`. They check whether each `[N]` citation marker in an answer points to the SOURCE passage that supports the **specific claim it is attached to** — distinct from `groundedness_measure_with_cot_reasons`, which checks whether a statement is supported by the source *somewhere*.

This targets the case in #2408, and specifically the **supported-elsewhere misattribution** failure: a claim cited to passage `[A]` that does not support it, even though some other passage `[B]` in the source would. Groundedness passes that; an attribution-aware check fails it.

## How

- `citation_attribution(question, source, statement)` -> score in [0, 1] via `generate_score`.
- `citation_attribution_with_cot_reasons(...)` -> `(score, reasons)` via `generate_score_and_reasons`.
- `source` accepts a list (numbered `[1] ...`, `[2] ...` so the `[N]` markers resolve) or a pre-numbered string.
- New `CitationAttribution` template in `templates/rag.py` (system + cot_system + user prompts), exported in `__all__`.
- API golden snapshot updated for the two new `LLMProvider` methods.
- Unit tests in `tests/unit/test_citation_attribution.py` (no API key required).

## Verification

Ran on Python 3.11: `test_feedback_templates_exports`, `test_feedback_templates_validation`, and the new `test_citation_attribution` all pass. The `LLMProvider` golden entry was confirmed against introspection.

## Usage

```python
from trulens.core import Metric, Selector

f_citation = Metric(
    implementation=provider.citation_attribution,
    name="Citation Attribution",
    selectors={
        "question": Selector.select_record_input(),
        "source": Selector.select_context(collect_list=True),
        "statement": Selector.select_record_output(),
    },
)
```

Relates to #2408.


rsrijith:
This implements the citation/attribution evaluator requested in #2408: it adds `citation_attribution` (and `citation_attribution_with_cot_reasons`) to `LLMProvider`, plus a `CitationAttribution` template in `templates/rag.py`, matching the issue's spec.

Beyond the base case, it also covers the **supported-elsewhere misattribution** failure — a claim cited to `[A]` that `[A]` does not support, even though some other retrieved passage `[B]` does. Plain `groundedness_measure_with_cot_reasons` marks that as grounded; this feedback function flags it as a citation error.

Tests are green (9 unit tests, Python 3.11). @joshreini1 since you opened #2408 — happy to adjust naming or the template wording to your preference.

src/feedback/trulens/feedback/llm_provider.py
```diff
@@ -2864,6 +2864,111 @@ def _remove_trivial_statements(self, statements: List[str]) -> List[str]:
...
+            question=question,
+            source=self._number_citation_sources(source),
+            statement=statement,
+        )
+        return self.generate_score_and_reasons(
+            templates_rag.CitationAttribution.cot_system_prompt.format(
+                min_score=0, max_score=1
+            ),
+            user_prompt,
+            min_score_val=0,
```
joshreini1: don't hard code the min/max scores as binary, instead add support for them as kwargs to the metric, similar to groundedness. Also include criteria, additional_instructions and examples support

joshreini1 (changes requested):
Please add the mentioned kwargs to the metric to be consistent with other metrics. These arguments allow the user to tune the performance of the feedback function to their use case/domain. otherwise looks good, thanks for the contribution!

src/feedback/trulens/feedback/llm_provider.py
rsrijith: Addressed in 49e2e0f.

`CitationAttribution` now extends `CriteriaOutputSpaceMixin`, so both `citation_attribution` and `citation_attribution_with_cot_reasons` take `criteria`, `additional_instructions`, `examples`, `min_score_val` and `max_score_val` — the same surface as `groundedness_measure_with_cot_reasons` and `context_relevance`.

- The 0/1 range is no longer hard-coded. It is now the *default* (`min_score_val=0`, `max_score_val=1`), preserving the previous behavior, and the range resolves through `_determine_output_space`, so `LIKERT_0_3` / `LIKERT_0_10` work as well.
- The scoring rules moved into a `criteria_template` parameterized on `{min_score}`/`{max_score}`, with the system prompt built by `generate_system_prompt(...)`.
- The separate `cot_system_prompt` is gone. Both variants build the system prompt the same way, so criteria/instructions/examples apply to each, and the CoT variant swaps the trailing label for `COT_REASONS_TEMPLATE` the way `context_relevance_with_cot_reasons` does.

Unit tests cover the new kwargs: score-range passthrough, rejection of an unsupported range, `criteria` + `additional_instructions` injection, and few-shot `examples`. 10/10 pass locally, and the rest of `tests/unit` is unchanged (293 passed).

joshreini1: thanks for the quick update

joshreini1 (approved):

Issue #2440, opened 2026-05-04
Nested App / Sub-Agent Tracing (NESTED_RECORD_ROOT)


sfc-gh-jreini:
## Summary

When an instrumented app calls another instrumented app (e.g., an orchestrator agent calling sub-agents), TruLens should properly nest their trace trees rather than creating disconnected root spans.

## Motivation

Multi-agent architectures (LangGraph supervisors, CrewAI crews, AutoGen groups) commonly have an outer "orchestrator" that delegates to inner "specialist" agents. Each inner agent may be its own `TruApp` with independent eval. Currently there is no way to express this parent-child relationship between apps in the trace.

## Proposed Approach

1. Add `NESTED_RECORD_ROOT` span type (see PR #1825 for prior art)
2. When a `Recording` context detects it is already inside another `Recording`, mark the inner root span as `NESTED_RECORD_ROOT` and link it to the outer span via `parent_span_id`
3. Add `parent_app_id` attribute to nested record roots for cross-app lineage
4. Dashboard should render nested apps as collapsible sub-trees

## Acceptance Criteria

- [ ] `NESTED_RECORD_ROOT` span type added to `SpanAttributes.SpanType`
- [ ] Inner `TruApp` recording within outer `TruApp` produces linked spans
- [ ] Each app still independently evaluates with its own metrics
- [ ] Unit test demonstrating nested recording with 2+ apps
- [ ] Dashboard renders nested spans (can be follow-up)

## Related

- PR #1825 (prior implementation attempt)
- Issue #2423 (conversation_id grouping)

YingzuoLiu:
Hi TruLens team, I’m interested in helping with this.

I took a look at #1825 (`davidkurokawa/otel/nested_apps`). It looks like that draft PR added `NESTED_RECORD_ROOT` to `SpanType` and a spec-style test file (`test_otel_nested_apps.py`), but the implementation itself was not completed or merged.

Reading the test cases, the desired behavior seems a bit more subtle than just adding a new span type: a `RECORD_ROOT`-decorated call needs to know whether the relevant `TruApp` / recording context is active, and nested app calls should be linked into the outer recording rather than becoming disconnected roots. From the current OTEL path, it looks like `CreateSpanFunctionCallContextManager` mainly decides whether to start a record based on whether `RECORD_ID` already exists in baggage, without an explicit notion of which app instance owns the active recording context.

A focused first pass I could work on would be:

[...]

PR #2583, opened 2026-06-30
Add nested record root span linkage for nested TruApp calls

YingzuoLiu:
# Description

Part of #2440.

This PR adds initial support for representing nested `TruApp` calls as explicit nested record roots in OTEL traces.

When an inner `TruApp` starts while an outer recording context is active, the inner root span is marked as `NESTED_RECORD_ROOT` instead of being treated as a disconnected root or being silently folded into the outer record.

The nested root keeps its own `record_id` / app identity and carries explicit parent linkage fields:

- `ai.observability.nested_record_root.parent_span_id`
- `ai.observability.nested_record_root.parent_app_id`

This is intended as a focused first pass on trace-shape semantics. Dashboard rendering is left as a follow-up.

The implementation updates OTEL recording context handling so that nested app calls can establish their own app identity while preserving explicit parent linkage to the active outer span.

Added tests cover:

- outer app starts a normal `RECORD_ROOT`
- inner app called under an active outer recording becomes `NESTED_RECORD_ROOT`
- inner nested root keeps a distinct `record_id` / app identity
- nested root links back to the outer active span and app
- a standalone app call without an outer context remains a normal `RECORD_ROOT`
- active recording without a resolvable parent span falls back to a normal `RECORD_ROOT`
- baggage-only parent linkage is not treated as authoritative

Local test results:

```bash
pytest tests/unit/test_otel_nested_record_root.py tests/unit/test_otel_record_root.py -q
# 7 passed

pytest tests/unit/test_otel_nested_record_root.py --collect-only -q
# 4 tests collected

git diff --check
# no output
```


sfc-gh-jreini:
Did a careful review + built a runnable demo. Summary: **solid, well-tested first pass** — the nested-root semantics are correct and the design is sound. A few non-blocking suggestions below.

### Verified locally
- `pytest tests/unit/test_otel_nested_record_root.py` → 4 passed.
- Built an orchestrator→knowledge-base demo (outer `TruApp` delegating to an inner `TruApp`). Confirmed end-to-end:
  - inner root is `NESTED_RECORD_ROOT`, outer is `RECORD_ROOT`
  - distinct `record_id` and `app_id`
  - `nested.parent_span_id == outer.span_id` and `nested.parent_app_id == outer.app_id`

### What's good
- Nested detection is appropriately strict: requires all four outer baggage markers (`record_id`, `app_id`, `__trulens_otel_ctx__`, `__trulens_recording__`) **and** a valid current span before linking.
- The `parent_record_id == record_id` guard neatly prevents deeper child spans from being re-marked as nested roots.
- Graceful `unjoinable` fallback to a normal `RECORD_ROOT` when the parent span can't be resolved — nice, and tested.
- Token-based cleanup in `__exit__` restores the outer context correctly, so the `override=True` attaches don't leak.
- Backward-compatible for single-app usage (override on a fresh context is a no-op).

### Suggestions (non-blocking)
1. **Extract the baggage keys to constants.** The `__trulens_nested_record_parent_*__` / `__trulens_nested_record_unjoinable_parent_record_id__` string literals are hand-written in both `function_call_context_manager.py` and `instrument.py`. A shared module-level constant (or a small enum) avoids silent drift from a typo.
2. **`parent_span_id` format.** It's stored as `str(current_span_context.span_id)` (decimal). Since the follow-up dashboard work will need to join nested roots to their parent, it's worth confirming this matches how span IDs are represented elsewhere in the event/trace records (hex vs decimal) so the join key lines up later.
3. **Consider a 3-level nesting test.** The chaining logic looks correct (each level re-derives its parent from the then-current span), but a depth-3 case would lock in that behavior.
4. Optional: a one-line comment on the reworked `attach_to_context` (the `None`/`override` handling) since the early-return condition changed.

Happy to share the demo notebook if useful for the dashboard follow-up.

src/core/trulens/core/otel/function_call_context_manager.py
```diff
@@ -30,8 +30,31 @@ def __init__(self, span_name: str) -> None:
     def __enter__(self) -> Span:
         # Set record_id into context.
         started_record = False
+        nested_record_root = False
+
         record_id = get_baggage(SpanAttributes.RECORD_ID)
-        if not record_id:
+        parent_record_id = get_baggage(
+            "__trulens_nested_record_parent_record_id__"
```
joshreini1: These nested-record baggage keys are hardcoded string literals here and duplicated verbatim in `instrument.py`. Suggest extracting them to shared module-level constants (or a small enum) so the producer and consumer can't silently drift from a typo in one place.

src/core/trulens/core/otel/instrument.py
```diff
@@ -622,12 +629,57 @@ def __init__(
...
+            and parent_recording is not None
+        ):
+            current_span = trace.get_current_span()
+            current_span_context = current_span.get_span_context()
+            if (
+                current_span_context is not None
+                and current_span_context.is_valid
+            ):
+                self.attach_to_context(
+                    "__trulens_nested_record_parent_record_id__",
```
joshreini1: Same baggage-key literals as in `function_call_context_manager.py`. A single shared constant for each key would keep the producer (here) and consumer (the context manager) in sync.

src/core/trulens/core/otel/instrument.py
```diff
@@ -553,8 +556,12 @@ def __init__(
 
     # Calling set_baggage does not actually add the baggage to the current context, but returns a new one
     # To avoid issues with remembering to add/remove the baggage, we attach it to the runtime context.
-    def attach_to_context(self, key: str, value: object):
-        if get_baggage(key) or value is None:
+    def attach_to_context(
```
joshreini1, line 597: The early-return condition changed here (None-handling split out, new `override` flag). A one-line comment on the intended semantics — skip when value is unset; only overwrite existing baggage when `override=True` — would help future readers.

joshreini1 (commented):
Re-reviewed after a local rebase onto `main` (clean, no conflicts). Re-ran the unit tests (4 passed) and my orchestrator→knowledge-base demo — all nested-root linkage checks still pass.

On my earlier suggestion about the `parent_span_id` format: **checked, no change needed.** The event store's `trace.span_id` is serialized as `str(context.span_id)` (decimal) in `exporter/utils.py`, and this PR stores `parent_span_id` the same way (`str(...span_id)`), matching the existing `root_span_id` convention in `instrument.py`. So the join key will line up for the dashboard follow-up.

Two small things worth addressing before merge (inline). The extra-nesting-depth test I mentioned earlier isn't necessary — the chaining logic reads correctly.

src/core/trulens/core/otel/function_call_context_manager.py
YingzuoLiu: Addressed in 3a699b4ac: extracted the nested-record baggage keys into shared module-level constants in function_call_context_manager.py and imported them from instrument.py so the producer and consumer stay in sync.

src/core/trulens/core/otel/instrument.py
YingzuoLiu: Addressed in 3a699b4ac: instrument.py now imports and uses the shared nested-record baggage-key constants from function_call_context_manager.py instead of duplicating the string literals.

src/core/trulens/core/otel/instrument.py
YingzuoLiu: Addressed in 3a699b4ac: added a comment documenting that unset values are skipped and existing baggage is only overwritten when override=True, including why nested TruApp calls need that behavior.

YingzuoLiu:
Addressed the review feedback in 3a699b4ac:

- extracted the nested-record baggage keys into shared constants used by both the producer and consumer
- documented the None/override semantics in attach_to_context

Validation:
- poetry run pytest tests/unit/test_otel_nested_record_root.py tests/unit/test_otel_record_root.py -q → 7 passed
- pre-commit checks for both modified files → passed
- git diff --check → clean

@joshreini1 could you please take another look when you have a chance?

joshreini1 (approved):
Thanks for contributing @YingzuoLiu !

PR #2596, opened 2026-07-11
Google multimodal support

### LINKED ISSUE #2403: Add Multi-Modal Support to Google Provider
## Summary
There's an explicit TODO in [`src/providers/google/trulens/providers/google/provider.py`](https://github.com/truera/trulens/blob/main/src/providers/google/trulens/providers/google/provider.py) (line ~119):

```python
# TODO: Add multi-modal (text + image) handling here
```

## What
Extend the Google/Gemini provider to accept image inputs alongside text, enabling multi-modal evaluation (e.g., evaluating image-captioning or vision-RAG apps).

## Difficulty
Medium

RyanFlowerYes:
# Description

Adds multi-modal (text + image/audio/video) input support to the Google (Gemini)
feedback provider, completing the existing `TODO` in
`src/providers/google/trulens/providers/google/provider.py`.

Previously, `Google._create_chat_completion` assumed `message["content"]` was
always a plain string and wrapped it in a single text `Part`. There was no way to
pass media through the provider, so users wanting multi-modal evaluations had to
subclass `Google` and call `google_client.models.generate_content` directly,  as
shown in `examples/expositional/models/google/gemini_multi_modal_evaluation.ipynb`.
That workaround bypasses `self.endpoint` entirely, meaning those calls receive no
cost tracking, token counting, or rate-limit pacing.

`content` may now be either:

- a `str` (unchanged behavior), or
- a list of content parts:
  - `"some text"`
  - `{"type": "text", "text": str}`
  - `{"type": "media", "data": bytes, "mime_type": str}`

Media parts are converted to `types.Part.from_bytes` and cover images, audio, and
video, any MIME type Gemini accepts. Example:

```python
provider = Google()
result = provider._create_chat_completion(
    messages=[{"role": "user", "content": [
        {"type": "text", "text": "Does this restaurant have outdoor seating?"},
        {"type": "media", "data": image_bytes, "mime_type": "image/png"},
    ]}],
    response_format=ImageFaithfulnessScore,
)
```

Also added a few tests to validate my implementation in 'tests/unit/test_google_multimodal.py'. Happy to change it up based on maintainer feedback!

- **Existing text evaluations are unaffected.** String content takes the same path
  as before and produces an identical payload. This is explicitly covered by tests.
- **Design decision, explicit `mime_type`.** Media parts require `mime_type`
  rather than defaulting or inferring it. Gemini cannot decode inline bytes without
  it, and a wrong default (e.g. assuming `image/png` for a WAV) fails in confusing
  ways. Missing `mime_type` raises a `ValueError`.
- **Design decision, one `media` type, not `image`.** The example notebook
  demonstrates image, audio, and video evaluations that differ only in MIME type.
  A single `media` part covers all three and extends to future modalities without a
  schema change.
- To my knowledge, there was no pre-existing multi-modal message convention anywhere in TruLens core
  (`llm_provider.py` only ever emits `content: str`), so this schema is new. Happy
  to align it with a different shape if maintainers prefer one.
- `types.Content` and `types.Part` are now used consistently for both the `messages`
  and `prompt` paths, replacing the hand-built dicts.

**Follow-up (not in this PR):** inline media is limited to ~20MB per request. Larger
files require the Gemini Files API (`client.files.upload` + `Part.from_uri`), which
needs caching to avoid re-uploading per evaluation call and does not exist on Vertex
AI clients. Happy to take that on in a separate PR.

## Type of change

  not work as expected)
- [x] New Tests

fixes #2403


joshreini1 (approved):
This is great - thank you @RyanFlowerYes . Can you also update the google multimodal example to show using the improvements you've made?

PR #2603, opened 2026-07-15
fix(otel): make instrument_method idempotent (closes #2477)

### LINKED ISSUE #2477: Add Tests for instrument_method() on Third-Party Classes
## Description

`instrument_method()` lets users instrument classes they don't own (e.g., third-party retrievers). While `@instrument()` has good test coverage in `test_otel_instrument.py`, there's no dedicated test coverage for `instrument_method()`.

## Proposed Change

Add tests to `tests/unit/test_otel_instrument.py` (or a new file) covering:

- Basic `instrument_method()` on a dummy class
- Span type assignment via `span_type` parameter
- Attribute mapping via `attributes` dict
- Lambda-based attribute extraction
- Verifying the original class behavior is unchanged after instrumentation
- Instrumenting the same method twice (idempotency)

## Reference

- `src/core/trulens/core/otel/instrument.py` — `instrument_method()` function
- Existing `test_otel_instrument.py` for test infrastructure / fixtures

## Running Tests

```bash
TEST_OPTIONAL=true poetry run pytest tests/unit/test_otel_instrument.py -v
```

aqilaziz:
Resubmission of #2551, closed on 2026-06-24 without explanation as part of a batch-close on my end. Closes #2477 (still open, unassigned).

## Real bug, still present on current main

`instrument_method()` has no idempotency guard — calling it twice on the same class/method (which happens in practice when instrumenting third-party classes from multiple entry points) double-wraps the method, producing duplicate spans for a single call.

Verified this is still an active bug on current `main` (not something already fixed since June) by temporarily reverting to the un-fixed version and re-running the new test:

```
AssertionError: 2 != 1
```

(two spans emitted for one instrumented call, instead of one)

## Fix

`instrument_method()` now checks for the `TRULENS_INSTRUMENT_WRAPPER_FLAG` marker before wrapping — if the method is already instrumented, it's a no-op.

## Verification

```
$ pytest tests/unit/test_otel_instrument.py -v
11 passed in 1.28s
```

Includes `test_instrument_method_twice_is_idempotent`, which I confirmed genuinely catches the bug (not just passes trivially) — reverting the fix locally makes it fail with `2 != 1`, matching the bug description above; re-applying the fix passes clean.

### DIFF
--- a/src/core/trulens/core/otel/instrument.py
+++ b/src/core/trulens/core/otel/instrument.py
@@ -144,7 +144,9 @@ def _set_span_attributes(
                 func_exception,
             )
         # Set function call attributes.
-        set_function_call_attributes(span, ret, func_name, func_exception, all_kwargs)
+        set_function_call_attributes(
+            span, ret, func_name, func_exception, all_kwargs
+        )
     # Resolve the attributes.
     if instance is not None:
         args_with_self_possibly = (instance,) + args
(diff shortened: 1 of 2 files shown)


joshreini1 (approved):
Thanks for identifying and fixing @aqilaziz !

PR #2580, opened 2026-06-28
test(feedback): add integration tests for feedback pipeline wiring

nuthalapativarun:
closes #2494

## What changed
- Added `tests/integration/test_feedback_pipeline_wiring.py`
- Uses a `MockLLMProvider` that intercepts `generate_score` / `generate_score_and_reasons` so the real template-rendering and prompt-assembly path executes without a live endpoint
- Covers 8 cases across 4 feedback methods:
  - `relevance`: returns float in [0, 1], assembles `PromptResponseRelevance` system prompt, includes both inputs in user prompt
  - `context_relevance`: returns float in [0, 1], assembles `ContextRelevance` system prompt
  - `sentiment`: returns float in [0, 1], assembles `Sentiment` system prompt
  - `groundedness_measure_with_cot_reasons`: returns `(float, dict)` tuple with score in [0, 1]

## Testing
```
python -m pytest tests/integration/test_feedback_pipeline_wiring.py -v
# 8 passed
```

PR #2604, opened 2026-07-16
test: add feedback pipeline wiring integration coverage (closes #2494)

### LINKED ISSUE #2494: Integration Test: Feedback Function Pipeline Wiring
## Description

Existing tests (`test_otel_rag_triad.py`) use mock feedback functions that bypass template rendering, prompt assembly, and score parsing entirely. We need tests that verify the full pipeline — from `Feedback()` object creation through template rendering, prompt assembly, and score normalization — works correctly for each major feedback function.

## Proposed Change

Create `tests/integration/test_feedback_pipeline_wiring.py` with curated (input, mock_response, expected_score) triples:

1. Define a `MockLLMProvider` that returns different deterministic responses based on input patterns
2. Test each major feedback function through the full pipeline:
   - **Groundedness**: source + statement → mock returns score → verify normalized output
   - **Relevance**: prompt/response pair → mock returns score → verify output
   - **Context Relevance**: query/context → mock returns score → verify output
   - **Sentiment**: text → mock returns score → verify output
3. Asserts:
   - Correct prompts are assembled (system + user) for each feedback function
   - The input data is correctly interpolated into the template
   - Score parsing produces the expected normalized float
   - The `Feedback` object correctly routes inputs via `on_input()`, `on_output()`, `on_context()`

## Files to Reference

- `src/feedback/trulens/feedback/llm_provider.py` — `relevance()`, `groundedness_measure_with_cot_reasons()`, `context_relevance()`, `sentiment()`
- `tests/unit/test_feedback_criteria_and_additional_instructions.py` — `MockLLMProvider` pattern

## Running Tests

```bash
poetry run pytest tests/integration/test_feedback_pipeline_wiring.py -v
```

aqilaziz:
Resubmission of #2552, closed on 2026-06-24 without explanation as part of a batch-close on my end. Closes #2494 (still open, unassigned).

Same fix needed as #2603/#2597: `MockLLMProvider` passed `endpoint=None`, and `groundedness_measure_with_cot_reasons()` asserts `self.endpoint is not None` before reaching the mocked scoring methods. Fixed with `DummyEndpoint` (trulens' own test double).

## Verification

```
$ pytest tests/integration/test_feedback_pipeline_wiring.py -v
4 passed in 22.70s
```

### DIFF
--- a/tests/integration/test_feedback_pipeline_wiring.py
+++ b/tests/integration/test_feedback_pipeline_wiring.py
@@ -1,0 +1,130 @@
+"""Integration tests for feedback-function pipeline wiring."""
+
+from typing import Dict, Optional, Tuple
+import unittest
+
+from trulens.core import Metric
+from trulens.feedback import llm_provider
+
+
+class MockLLMProvider(llm_provider.LLMProvider):
+    """Mock provider that records prompts and returns deterministic scores."""
+
+    model_config = {"extra": "allow"}
+
+    last_system_prompt: Optional[str] = None
+    last_user_prompt: Optional[str] = None
+
+    def __init__(self, **kwargs):
+        super().__init__(endpoint=None, model_engine="mock-model", **kwargs)
+
+    def generate_score(
+        self,
+        system_prompt: str,
+        user_prompt: Optional[str] = None,
+        min_score_val: int = 0,
+        max_score_val: int = 3,
+        temperature: float = 0.0,
+    ) -> float:
+        self.last_system_prompt = system_prompt
+        self.last_user_prompt = user_prompt
+        return 2 / 3
+
+    def generate_score_and_reasons(
+        self,
+        system_prompt: str,
+        user_prompt: Optional[str] = None,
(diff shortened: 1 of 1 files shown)


joshreini1 (approved):
Solid resubmission -- the key improvement over #2580 is exercising the full Metric.run() path rather than calling the provider method directly, which means this actually tests the pipeline wiring end-to-end. Using DummyEndpoint is the right call since groundedness_measure_with_cot_reasons asserts endpoint is not None. Note: #2580 covers similar ground -- once this merges we can close that one. Thanks for the contribution!

joshreini1 (approved):
Solid resubmission -- the key improvement over #2580 is exercising the full Metric.run() path rather than calling the provider method directly, which tests the pipeline wiring end-to-end. Using DummyEndpoint is the right call since groundedness_measure_with_cot_reasons asserts endpoint is not None. Note: #2580 covers similar ground -- once this merges we can close that one. Thanks for the contribution!

PR #2582, opened 2026-06-29
Fix #2406: Implement HuggingFaceLocal Stub Methods

Mr-Neutr0n:
Fixes #2406

Implemented all 8 `HuggingfaceLocal` stub methods in `src/providers/huggingface/trulens/providers/huggingface/provider.py` — each maps to a HuggingFace model via the existing `_retrieve_tokenizer_and_model` pattern (with `AutoModelForTokenClassification` added for PII NER, and special `trust_remote_code`/separate tokenizer handling for the Vectara hallucination model). Added `model_class` parameter to `_retrieve_tokenizer_and_model` and 7 model path constants extracted from the existing API URL constants.

Local test infra unavailable in CI sandbox.

---
This change was prepared with AI assistance under human direction and review.

PR #2606, opened 2026-07-17
docs: add HuggingFace local eval quickstart (closes #2470)

### LINKED ISSUE #2470: Add HuggingFace Local Eval Cookbook
## Description

The existing HuggingFace example (`examples/expositional/models/local_and_OSS_models/local_vs_remote_huggingface_feedback_functions.ipynb`) uses the old API. We need a modern cookbook using `@instrument()` + OTEL tracing with a local HuggingFace model.

## Proposed Change

Create `examples/expositional/models/huggingface/huggingface_local_quickstart.ipynb` that demonstrates:

1. Installing `trulens-providers-huggingface` and a small local model
2. Using `@instrument()` to trace a simple pipeline
3. Running feedback evaluations locally (no API calls)
4. Viewing traces and scores in the dashboard

## Reference

- Provider: `src/providers/huggingface/trulens/providers/huggingface/provider.py` (extends `core_provider.Provider`, not `LLMProvider`)
- This is a good opportunity to show the non-LLM provider path

aqilaziz:
Resubmission of #2553, closed on 2026-06-24 without explanation as part of a batch-close on my end. Closes #2470 (still open, unassigned).

Rebased onto current `main` via cherry-pick, clean apply. Verified notebook is valid JSON and the imports (`TruApp`, `Metric`, `Selector`, `TruSession`, `instrument`, `SpanAttributes`) match the same set already verified across #2586-#2604.


examples/expositional/models/huggingface/huggingface_local_quickstart.ipynb
```diff
@@ -0,0 +1,165 @@
+{
```
joshreini1, line 1: <pre class="ql-syntax" spellcheck="false">Line #1.    def simple_answer_length_score(response: str) -&gt; float:
</pre><p>Instead of a custom local python function, show the use of a huggingface model running locally. Right now you're just showing a custom metric which already is documented.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/truera/trulens/pull/2606/discussion/#comment-3615254451'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

sfc-gh-jreini:
Checking in — this has been open for a few weeks. My inline comment from July 20 asked for a substantive change: instead of a custom local Python scoring function, show a HuggingFace model actually running locally (e.g. a sentence-transformers or transformers pipeline used as the judge). That is what makes this notebook a HuggingFace local eval quickstart rather than a generic custom-metric example.

Can you update the notebook with that change? Happy to help if you are unsure what pattern to use.

sfc-gh-jreini (changes requested):
The instrumentation pattern is clean and the notebook structure is good. One issue: the title promises 'HuggingFace Local Evaluation' but the feedback cell uses a custom toy metric (`simple_answer_length_score`) that doesn't use HuggingFace at all.

PR #2582 (HuggingfaceLocal stub implementations) merged on Aug 7, so the `HuggingfaceLocal` provider now has working methods. Please add a cell showing at least one real `HuggingfaceLocal` feedback function — for example:

```python
from trulens.providers.huggingface import HuggingfaceLocal
provider = HuggingfaceLocal()
f_context_relevance = Metric(
    implementation=provider.context_relevance,
    name="Context Relevance",
    selectors={
        "question": Selector.select_record_input(),
        "context": Selector.select_context(),
    },
)
```

The custom metric example is fine as a secondary fallback (showing users how to write a custom `Metric` when a provider method doesn't exist), but it shouldn't be the only feedback example in a notebook titled 'HuggingFace Local Evaluation'.

PR #2607, opened 2026-07-20
build: add Python 3.13 support for snowflake connector and cortex provider

sfc-gh-jreini:
## Summary
- Raises the Python cap from `<3.13` to `<3.14` for **trulens-connectors-snowflake** and **trulens-providers-cortex**, enabling install on Python 3.13.
- Underlying deps now support 3.13: `snowflake-snowpark-python` (`<3.15`), `snowflake-connector-python`, `snowflake-sqlalchemy`, and `snowflake-ml-python` (`<3.14`, ref [snowflake-ml-python#172](https://github.com/snowflakedb/snowflake-ml-python/issues/172)).
- Updates package `pyproject.toml`, root develop-path markers, conda `meta.yaml` recipes, regenerates `poetry.lock`, and enables the `py313-static` CI matrix entry.
- Drops the stale `<3.13` cap from non-snowflake conda recipes (core, feedback, dashboard, otel, langchain, langgraph) to match their already-uncapped pyprojects.

### Why `<3.14` (not fully uncapped)
Root `python = "^3.10"` means poetry must resolve deps across the whole `>=3.10,<4.0` range for these packages. `snowflake-ml-python` caps at `<3.14`, so `<3.14` is the correct ceiling that includes 3.13 while keeping resolution solvable.

## Test plan
- [x] `poetry lock --no-update` resolves cleanly
- [x] `poetry check` passes for root, connector, and cortex packages
- [x] pre-commit hooks pass (pyproject-fmt, poetry-check)
- [ ] CI green on the new `py313-static` matrix leg (basic + optional + snowflake unit tests)


sfc-gh-jreini:
Note on risk area: the connector side (`trulens-connectors-snowflake`) is fully clear — its only deps (`snowflake-snowpark-python`, `snowflake-sqlalchemy`) support 3.13.

The one thing to watch is **`trulens-providers-cortex`**, which depends on `snowflake-ml-python`. That package installs on 3.13 (pure-python wheel, `requires-python = <3.14`) but 3.13 isn't officially GA'd yet ([snowflake-ml-python#172](https://github.com/snowflakedb/snowflake-ml-python/issues/172) still open). The real risk is its transitive C-extension deps (e.g. numba/llvmlite) having 3.13 wheels — the new `py313-static` CI leg will validate this.

If cortex fails on 3.13 in CI, the clean fallback is to keep cortex pinned at `<3.13` and still land connector 3.13 support, since the two are independent.

A5Wagyu32 (approved):

PR #2329, opened 2026-01-15
Unify feedback and metricconfig implementations into new Metric API

sfc-gh-jreini:
# Description

Both Snowflake and OSS users now use the same Metric class with the same API - documentation, examples, and code are portable across environments.

The new metric api works in both OSS and Snowflake

Screenshot 2026-01-28 at 11 33 15 AM
Screenshot 2026-01-28 at 9 49 39 AM

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to
  not work as expected)
- [ ] New Tests
- [x] This change includes re-generated golden test results
- [x] This change requires a documentation update

----

> [!IMPORTANT]
> This pull request unifies the feedback and metricconfig implementations into a new Metric API, replacing the Feedback class and updating documentation, tests, and codebase for a consistent metric handling interface.
> 
>   - **Metric API**:
>     - Introduces `Metric` class in `metric/metric.py` replacing `Feedback` and `MetricConfig`.
>     - `Metric` class handles evaluation metrics with a unified interface.
>     - Deprecates `Feedback` class, re-implemented as an alias for `Metric`.
>   - **Documentation**:
>     - Updates documentation to reflect new `Metric` API.
>     - Adds `metric_migration.md` for migrating from `Feedback` to `Metric`.
>     - Updates examples and guides to use `Metric` class.
>   - **Code Changes**:
>     - Refactors codebase to use `Metric` instead of `Feedback`.
>     - Updates tests to use `Metric` class.
>     - Adds deprecation warnings for `Feedback` and `MetricConfig`.
>   - **Schema and Models**:
>     - Updates `FeedbackDefinition` to support `Metric` attributes.
>     - Plans to rename schema classes from `Feedback*` to `Metric*` in future.
>   - **Backward Compatibility**:
>     - Maintains backward compatibility with deprecation warnings.
>     - Allows gradual migration by supporting old `Feedback` methods in `Metric`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=truera%2Ftrulens&utm_source=github&utm_medium=referral)<sup> for 9dafd1464b383b321d370c2aba34217d73f43f83. You can [customize](https://app.ellipsis.dev/truera/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

PR #2609, opened 2026-07-22
feat(core): add BatchEvaluator for offline dataset evaluation (#2444)

### LINKED ISSUE #2444: Batch/Offline Evaluation Mode: Evaluate Datasets Without a Live App
## Summary

Add a first-class mode to run metrics over a pre-collected dataset without requiring a live app or recording session. Essential for CI/CD eval pipelines, regression testing, and benchmark suites.

## Motivation

Common workflows that need this:
- Run evals in CI against a golden test set
- Benchmark prompt changes across 1000+ examples
- Evaluate historical production logs offline
- Compare two models on the same dataset

Currently users must either replay through a live app or hack together virtual records. Neither is ergonomic for batch workloads.

## Proposed Approach

In batch mode, there are no live traces/spans. Instead, dataset columns are mapped to metric implementation arguments via `Selector`-based configuration. The `Selector` abstraction is extended (or a `DatasetSelector` variant is introduced) to select from tabular data rather than span attributes.

```python
from trulens.core import BatchEvaluator, Metric, Selector

provider = OpenAI()

# Metrics use Selectors that point to dataset columns
relevance = Metric(
    name="answer_relevance",
    implementation=provider.relevance,
    selectors={
        "prompt": Selector.from_column("query"),
        "response": Selector.from_column("answer"),
    },
)

groundedness = Metric(
    name="groundedness",
    implementation=provider.groundedness,
    selectors={
        "source": Selector.from_column("contexts"),
        "statement": Selector.from_column("answer"),
    },
)

# Custom metric with arbitrary arg names
my_metric = Metric(
    name="custom_quality",
    implementation=my_quality_fn,  # def my_quality_fn(doc: str, summary: str) -> float
    selectors={
        "doc": Selector.from_column("original_document"),
        "summary": Selector.from_column("generated_summary"),
    },
)

evaluator = BatchEvaluator(metrics=[relevance, groundedness, my_metric])

# From a DataFrame — Selectors map metric args to columns
results = evaluator.evaluate(df)

# From a list of dicts
results = evaluator.evaluate([
    {"query": "...", "answer": "...", "contexts": ["..."], "original_document": "...", "generated_summary": "..."},
    ...
])

# Optional: override column mapping at eval time
results = evaluator.evaluate(df, column_map={"user_question": "query"})
```

### Design Principles

1. **Selector-driven**: Reuse the `Selector` abstraction — extend it with `Selector.from_column(col_name)` for tabular data sources
2. **Flexible arg names**: No hardcoded input/output/context. Metrics declare what args they need; Selectors map those to dataset columns.
3. **Works with any metric signature**: Custom metrics with arbitrary parameter names work the same as built-in ones
4. **Parallel execution**: Configurable concurrency with rate limiting
5. **Persist optionally**: Results can be written to TruSession storage with synthetic trace metadata for dashboard viewing

## Acceptance Criteria

- [ ] `Selector.from_column(column_name)` (or equivalent) for selecting from tabular data
- [ ] `BatchEvaluator` class with `evaluate()` method
- [ ] Metric arg names are arbitrary — driven entirely by Selector configuration
- [ ] Supports pandas DataFrame and list-of-dict inputs
- [ ] Optional `column_map` for renaming dataset columns at eval time
- [ ] Parallel execution with rate limiting
- [ ] Results include per-row scores, explanations, and latencies
- [ ] Optional: persist to storage backend for dashboard viewing
- [ ] Works with all existing Metric implementations (including custom metrics with arbitrary signatures)
- [ ] Example notebook: "Evaluate RAG on BEIR dataset"

## Open Questions

- Should `Selector.from_column()` be a new static method on the existing `Selector` class, or a separate `DatasetSelector` subclass?
- How do we handle list-valued columns (e.g., contexts) — should `collect_list` semantics carry over from span-based Selectors?
- Can we unify this with the span-based Selector so the same Metric definition works both in live tracing and batch mode?

## Related

- examples/quickstart/batch_evaluation.ipynb (existing batch example to build on)
- TRAIL benchmarking work (PR #2314)
- `Selector` class in `src/core/trulens/core/feedback/selector.py`

desusaiteja:
## Description

Closes #2444.

Adds a first-class way to run metrics over a pre-collected dataset without a live app or recording session — the workflow you want for CI evals, regression checks, and benchmarking across many examples.

Today you have to replay data through a live app or hand-build virtual records. This adds a `BatchEvaluator` that takes a DataFrame (or list of dicts) and scores it directly:

```python
from trulens.core import BatchEvaluator, Metric, Selector

provider = OpenAI()

relevance = Metric(
    name="answer_relevance",
    implementation=provider.relevance_with_cot_reasons,
    selectors={
        "prompt": Selector.from_column("query"),
        "response": Selector.from_column("answer"),
    },
)

results = BatchEvaluator(metrics=[relevance]).evaluate(df)
```

Each metric argument is bound to a dataset column with the new `Selector.from_column(...)`, reusing the existing `Selector` abstraction rather than introducing a parallel one. `evaluate()` returns a DataFrame with the original columns plus a score, `_explanation`, and `_latency` column per metric.

### What's included
- `Selector.from_column(column, collect_list=..., ignore_none_values=...)` for tabular sources.
- `BatchEvaluator.evaluate()` — parallel execution, DataFrame and list-of-dict inputs, optional `column_map` to rename columns at eval time, and `collect_list` handling for list-valued columns (e.g. contexts).
- Works with any metric, including custom functions with arbitrary argument names.
- 18 unit tests (run fully offline, no LLM calls) and an `offline_batch_evaluation` quickstart notebook.

### Notes for developers
- Additive, backwards-compatible. Span-based selectors are untouched; a dataset selector just carries a `dataset_column` and is mutually exclusive with the span fields.
- Optional dashboard persistence (an "optional" item in the issue) is intentionally left out for now, since it would mean synthesizing OTEL spans/events — better as a focused follow-up.
- The `api.trulens.3.11.yaml` golden isn't regenerated here (needs `make write-api` on 3.11); it's already missing `Metric`/`Selector` on main.

## Type of change

- [x] New Tests


src/core/trulens/core/batch.py
```diff
@@ -0,0 +1,361 @@
...
+logger = logging.getLogger(__name__)
+
+# A single row of a dataset, as a mapping from column name to value.
+Row = Mapping[str, Any]
+
+# Accepted dataset input types.
+DatasetLike = Union[pd.DataFrame, Sequence[Row]]
+
+
+class BatchEvaluator:
```
sfc-gh-jreini, line 78: Heads-up, not a blocker. The issue marks persistence/dashboard as optional, and that's fine -- this in-memory, no-span path is a legitimate gap `Run`/`LOG_INGESTION` doesn't cover: `LOG_INGESTION` is span-mediated (`dataset_spec` maps span-attribute paths to columns and `_create_virtual_spans` synthesizes OTEL spans), so it can't take arbitrary metric arg names the way this does.

Two doc asks so users pick the right tool:

1. Make the "no persistence / no dashboard" scope explicit in the notebook -- one line that results stay in the returned DataFrame (nothing is written to the dashboard/event table) and that the persisted + dashboard path is `Run` with `mode=Mode.LOG_INGESTION`.
2. Add docstrings for the new public surface -- `BatchEvaluator`, `BatchEvaluator.evaluate`, and `Selector.from_column` -- including the same "use `Run`/`LOG_INGESTION` for dashboard results" pointer. `evaluate`'s docstring currently doesn't note that nothing is persisted.

No code change intended here -- just don't let a user discover via surprise that their batch eval doesn't show up in the leaderboard.

src/core/trulens/core/batch.py
```diff
@@ -0,0 +1,361 @@
...
+                or not selector.is_dataset_selector
+            ):
+                raise ValueError(
+                    f"Metric {metric.name!r} selector for argument "
+                    f"{arg_name!r} is not a dataset selector. In batch mode, "
+                    "use `Selector.from_column(column_name)` for every "
+                    "argument."
+                )
+
+    def _unique_metric_names(self) -> List[str]:
```
sfc-gh-jreini, line 146: Small nit: the first duplicate stays un-suffixed and the second starts at `_1` (`["overlap","overlap"]` -> `["overlap","overlap_1"]`). Fine, but slightly surprising for code that iterates result columns expecting a stable prefix. Either suffix all duplicates (`overlap_0`, `overlap_1`) or note the behavior in the docstring. Not a blocker.

src/core/trulens/core/batch.py
```diff
@@ -0,0 +1,361 @@
...
+        for combination in combinations:
+            ins = dict(zip(arg_names, combination))
+            try:
+                result_and_meta = metric(**ins)
+            except SkipEval as e:
+                logger.debug("Skipping metric %s: %s", metric.name, e)
+                continue
+
+            if isinstance(result_and_meta, tuple):
+                assert len(result_and_meta) == 2, (
```
sfc-gh-jreini: This `assert` validates a value returned by the user's metric, but asserts are stripped under `python -O`. Since it guards a public contract ("metric returns a float, a float-dict, or `(value, meta)`"), use an explicit `raise ValueError(...)`/`TypeError` so the error survives optimization and reads better than a bare `AssertionError`.

src/core/trulens/core/batch.py
```diff
@@ -0,0 +1,361 @@
...
+
+    @staticmethod
+    def _aggregate_scores(metric: Metric, scores: List[Any]) -> Any:
+        """Aggregate one or more metric scores using the metric's aggregator."""
+        if len(scores) == 1:
+            return scores[0]
+        agg = metric.agg if metric.agg is not None else np.mean
+        try:
+            return agg(scores)
+        except Exception as e:  # pylint: disable=broad-except
```
sfc-gh-jreini, line 326: Aggregator failures become a silent `NaN` with only a `logger.warning`. The headline use case here is CI eval pipelines, and a silently-swallowed failure yielding `NaN` can pass a gate the wrong way -- e.g. `results["x"].mean() >= 0.7` is `False` on `NaN` (fine), but a "no error raised" gate or a `not isna` check would accept a totally broken metric. At minimum surface the exception type/message in the row's `_explanation`; consider re-raising when no `agg` is configured. `NaN`-on-error is fine; *silent* is the concern.

src/core/trulens/core/feedback/selector.py
```diff
@@ -542,6 +565,46 @@ def __init__(
...
+            collect_list=collect_list,
+            ignore_none_values=ignore_none_values,
+        )
+
+    @property
+    def is_dataset_selector(self) -> bool:
+        """Whether this selector reads from a tabular dataset column."""
+        return self.dataset_column is not None
 
     def describes_same_spans(self, other: Selector) -> bool:
```
sfc-gh-jreini, line 632: Real latent bug. This doesn't compare `dataset_column` (or `collect_list`/`ignore_none_values`). For two dataset selectors the span fields are all `None`, so any two `from_column` selectors compare equal regardless of which column they point at -- `Selector.from_column("query")` and `Selector.from_column("answer")` are "same spans."

It matters beyond batch: `_group_kwargs_by_selectors` in `compute_feedback_by_span_group` (`computer.py`) uses `describes_same_spans` to group metric kwargs. If dataset selectors ever flow into the live feedback path, that grouping silently collapses distinct columns. Add `and self.dataset_column == other.dataset_column` (and probably `collect_list`/`ignore_none_values`) to the comparison.

sfc-gh-jreini (changes requested):
Nice reuse of `Selector` -- the `dataset_column` field with mutual-exclusion against span fields is clean and backwards-compatible, and arbitrary-arg metrics + `collect_list` carryover are exactly what #2444 asked for.

I read `compute_feedback_by_span_group` and it's span-mediated (consumes trace events, groups via `describes_same_spans`), so the standalone direct loop here is the right call -- not a candidate for reuse.

The issue marks persistence/dashboard as optional, which I'm fine with. The one ask on that front is doc-shaped: make the no-persistence scope explicit in the notebook + docstrings so users aren't surprised, and point them to `Run`/`LOG_INGESTION` for the dashboard path (see comment on the class).

Two real correctness items (`describes_same_spans`, silent NaN) + two small nits below.

Also: `api.trulens.3.11.yaml` golden isn't regenerated and now falls further behind (`BatchEvaluator`, `from_column`) -- worth a `make write-api` pass here or a tracked follow-up.

src/core/trulens/core/feedback/selector.py
desusaiteja: Good catch, this was a real bug. Fixed by comparing `dataset_column` in `describes_same_spans`, with regression tests covering different-column, same-column, and dataset-vs-span selector cases. I left `collect_list`/`ignore_none_values` out of the comparison since the method answers "same source?" rather than "same extraction options", and adding them would also change grouping behavior for existing span selectors. Happy to include them if you'd prefer.

src/core/trulens/core/batch.py
desusaiteja: I agree, silent was the problem. Two changes: if a user-configured agg raises, the score stays NaN but the row's `_explanation` now carries the exception type/message plus the raw per-item scores, so nothing is lost. If the default aggregator fails (no agg configured), it now raises a `TypeError` with a hint to set an agg or use `collect_list=True`, per your suggestion: that case means the metric's return type can't be aggregated at all, so failing loudly is right. One note worth noting: `Metric.__init__` defaults agg to `np.mean`, so the "user configured?" check uses the serialized aggregator field. Tests added for both paths.

src/core/trulens/core/batch.py
desusaiteja: Done. Replaced with a `TypeError` that names the metric and the actual tuple length. Verified it still fires under `python -O`.

src/core/trulens/core/batch.py
desusaiteja: Kept the first occurrence un-suffixed so the common unique-name case produces clean column names, and documented the `overlap/overlap_1` behavior in both `_unique_metric_names` and evaluate's Returns section as suggested.

src/core/trulens/core/batch.py
desusaiteja: Added the scope note in all the places you listed: the notebook intro now states results live only in the returned `DataFrame`, and the `BatchEvaluator`, evaluate, and `Selector.from_column` docstrings each point to Run with `mode=Mode.LOG_INGESTION` as the persisted/dashboard path.

desusaiteja:
While regenerating the API golden for this PR, I hit an unrelated import failure on main that I wanted to flag:

`trulens/benchmark/benchmark_frameworks/tru_benchmark_experiment.py` fails to import because line 65 references `core_feedback.AggCallable`, but `AggCallable` moved from `trulens.core.feedback.feedback` to `trulens.core.metric.metric` in #2329. Since the annotation is evaluated at import time, any code path that imports the module breaks — including a full `make write-api`, which is why I merged the new golden entries selectively instead of doing a full regeneration.

Looks like a one-line fix — the file already imports `trulens.core.metric.metric` as `core_metric`, so it's just `core_feedback.AggCallable` → `core_metric.AggCallable` — though re-exporting `AggCallable` from the old location might be preferable if that import path is considered public.

What would you advise? I'm happy to fix it in this PR if you're fine with the small out-of-scope diff, open a separate one-line PR, or leave it to you — whichever you prefer.

desusaiteja:
@joshreini1  - While regenerating the API golden for this PR, I hit an unrelated import failure on main that I wanted to flag:

`trulens/benchmark/benchmark_frameworks/tru_benchmark_experiment.py` fails to import because line 65 references `core_feedback.AggCallable`, but `AggCallable` moved from `trulens.core.feedback.feedback to trulens.core.metric.metric`. Since the annotation is evaluated at import time, any code path that imports the module breaks, including a full make write-api, which is why I merged the new golden entries selectively instead of doing a full regeneration.

Please advise if you want me to open a separate PR for the fix or you prefer including that in this PR itself?

joshreini1 (commented):
You can include that fix in this PR as well - thanks for flagging!

desusaiteja:
Done, included the fix in this PR. `tru_benchmark_experiment.py` now uses `core_metric.AggCallable` since the module already imports `trulens.core.metric.metric`. Verified on Python 3.11 that the module imports cleanly again and the relevant unit suites pass (`test_batch_evaluator`, `test_static`, `test_otel_selector`, `test_feedback`). This also unblocks a full `make write-api` regeneration, which the import failure was previously preventing.

sfc-gh-jreini (approved):
All the items from my July 22 review are addressed — happy to approve.

**`describes_same_spans` correctness** — `dataset_column` is now included in the equality check, and `test_describes_same_spans_compares_dataset_column` covers the same column / different column / dataset-vs-span cases cleanly.

**Silent NaN** — the aggregation logic now distinguishes between a user-configured aggregator failure (logs a warning, returns `float("nan")`) and a default-aggregator failure (raises `TypeError` with a clear message pointing at `collect_list=True`). That distinction is documented in the docstring and tested.

**Golden file** — regenerated to include `BatchEvaluator`, `Metric`, `Selector`, and the new `trulens.core.batch` / `trulens.core.metric` modules.

**AggCallable import** — fixed in the latest commit.

**Persistence scope** — documented in the class docstring and prominently in the notebook note block, with a pointer to `Run` + `LOG_INGESTION` for users who want dashboard-visible results.

The `_validate_metric` guard at construction time (all selectors must be dataset selectors) is the right defence-in-depth here — it catches a likely user error before any I/O happens. Nice addition. Thanks for sticking with this one through multiple rounds!

Issue #2441, opened 2026-05-04
Async-Safe Recording: TaskGroup, asyncio.gather, Concurrent Spans


sfc-gh-jreini:
## Summary

Ensure `@instrument` and `Recording` context managers work correctly with Python async concurrency primitives (`asyncio.gather()`, `TaskGroup`, concurrent coroutines) without trace corruption or lost spans.

## Motivation

Real-world LLM apps heavily use async: parallel retrieval, concurrent tool calls, fan-out/fan-in patterns. If multiple instrumented coroutines run concurrently within the same recording, spans can get mis-parented or lost due to context propagation issues.

## Proposed Approach

1. Audit current `ContextVar` / OTEL context propagation under `asyncio.gather()`
2. Ensure each concurrent branch gets a proper copy of the span context
3. Add explicit support for `TaskGroup` (Python 3.11+)
4. Add span attribute indicating concurrency (e.g., `ai.observability.concurrent_branch_id`)
5. Test with realistic patterns: parallel retrievals, parallel tool calls, parallel evals

## Acceptance Criteria

- [ ] `asyncio.gather(*[instrumented_coro() for _ in range(N)])` produces N sibling spans under correct parent
- [ ] No span leakage between concurrent branches
- [ ] Works with both sync and async metrics
- [ ] Integration test with LangGraph parallel nodes
- [ ] Documentation example showing concurrent instrumentation

## Notes

This likely requires careful use of `context.attach()` / `context.detach()` and possibly `copy_context()` at gather boundaries.

xuyuanfan:
Hi @sfc-gh-jreini I can take this one. I’ll start with regression tests for asyncio.gather() and  TaskGroup focusing on span parenting and context isolation across concurrent tasks.
I will then trace the existing  contextVar and openTelemetry context propagation and make the smallest necessary change at the instrumentation boundary.  I will also cover the LangGraph parallel-node case and add a concise async example.
I expect to open a draft pr with the initial tests and findings within 1–2 days. let me know if there’s any related work I should coordinate with.

joshreini1:
Thanks @xuyuanfan !

sfc-gh-jreini:
@xuyuanfan — checking in! You mentioned opening a draft PR within a day or two back on July 22 — any updates? Happy to assign you if you're still working on it. If life got busy that's totally fine too, just let me know and I can reopen the issue to others.

xuyuanfan:
Thanks for checking in!

[...]

PR #2610, opened 2026-07-22
fix(otel): preserve metadata for cancelled async spans

xuyuanfan:
## Summary

Refs #2441

Handle async span finalization when a task is cancelled.

Previously, `asyncio.CancelledError` skipped the TruLens span finalization path because it is not an `Exception`. This caused cancelled spans to miss function call metadata even though the span itself was properly ended by OpenTelemetry.

This change ensures cancelled async spans run the same finalization logic while preserving the original cancellation behavior.

## Changes

- Add explicit handling for `asyncio.CancelledError` in the async instrumentation wrapper.
- Run `_finalize_span()` during cancellation.
- Preserve the original cancellation exception.

## Tests

Added async concurrency tests covering:
- `asyncio.gather()` sibling span relationships
- `TaskGroup` cancellation behavior
- cancelled span metadata finalization
- context cleanup after cancellation

### DIFF
--- a/src/core/trulens/core/otel/instrument.py
+++ b/src/core/trulens/core/otel/instrument.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import asyncio
 import inspect
 import logging
 import types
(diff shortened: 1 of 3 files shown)


sfc-gh-jreini (approved):
Thanks for the contribution @xuyuanfan !

xuyuanfan:
> 感谢您的贡献@xuyuanfan！

Thanks for the review and merge.

PR #2617, opened 2026-07-24
fix(dashboard): don't split app tag strings into characters in tag filter

### LINKED ISSUE #1689: [BUG] Using multiple tags for Record and App
**Bug Description**
Tags can only be a single string. I tried different formats (comma separated etc.) but the whole string is always split into separate characters, i.e. `tags="tag1,tag2"` leads to the tags `"t", "a", "g", "1", "2", ","`.

**Expected behavior**
Input values should be of type `str | list[str]`.
`tags=["tag1", "tag2"]` should lead to the tags `"tag1", "tag2"`

**Environment:**

- Python Version: 3.12
- TruLens version: 1.2.10

feiiiiii5:
## Summary

Fixes #1689. In the dashboard's **Advanced Filters → tags** multiselect, each app's tag string was exploded into individual characters — a tag like `production` showed up as the options `p`, `r`, `o`, `d`, `u`, `c`, `t`, `i`, `o`, `n` instead of `production`.

`AppDefinition.tags` / `Record.tags` is a single string (`trulens.core.schema.types.Tags`). The tag-option collection in `render_app_version_filters` did:

```python
tags = set()
for _, app_version in app_versions_df.iterrows():
    tags |= set(app_version["tags"])   # set() on a str iterates its characters
tags = sorted(tags)
```

`set("production")` is `{"p", "r", "o", "d", "u", "c", "t", "i", "o", "n"}` — exactly the garbled single-character tags in the report. Even a single, valid tag is affected.

## Fix

Extract the collection into a small, testable helper `get_unique_app_tags` that gathers **whole** tag strings (and skips empty strings so the default `""` tag doesn't become a blank filter option):

```python
tags = get_unique_app_tags(app_versions_df)
```

Scope is intentionally minimal: the metadata filters and the downstream tag match (`any(tag in x for tag in selected_tags)`) are unchanged — only the character-splitting is fixed, consistent with the existing single-string tag model. (Broader `list[str]` multi-tag support would be a separate, larger enhancement.)

## Test

Adds `tests/unit/streamlit/test_dashboard_utils.py::TestGetUniqueAppTags`:
- regression: tag strings are not split into characters (issue #1689)
- a single tag stays whole
- duplicates de-duplicated and sorted
- empty tag strings skipped
- all-empty yields no options

`ruff check` is clean on both changed files.

### DIFF
--- a/src/dashboard/trulens/dashboard/utils/dashboard_utils.py
+++ b/src/dashboard/trulens/dashboard/utils/dashboard_utils.py
@@ -508,6 +508,24 @@ def _handle_reset_filters(
             del st.query_params[query_param_key]
 
 
+def get_unique_app_tags(app_versions_df: pd.DataFrame) -> List[str]:
+    """Return the sorted unique app tags for the tag filter multiselect.
+
+    Each app version's ``tags`` value is a single string (see
+    ``trulens.core.schema.types.Tags``), so tags are collected as whole
+    strings. Using ``set(app_version["tags"])`` here would iterate the string
+    into individual characters (e.g. ``"prod"`` -> ``{"p", "r", "o", "d"}``),
+    which is what surfaced the garbled single-character tags in issue #1689.
+    Empty tag strings are skipped so they do not show up as a blank option.
+    """
+    tags: set = set()
+    for _, app_version in app_versions_df.iterrows():
+        tag = app_version["tags"]
+        if tag:
+            tags.add(tag)
+    return sorted(tags)
+
+
 def render_app_version_filters(
     app_name: str,
     other_query_params_kv: Optional[Dict[str, str]] = None,
@@ -541,10 +559,7 @@ def render_app_version_filters(
     with col1.popover("Advanced Filters", width="stretch"):
         # get tag options
         st.header("Advanced Filters")
-        tags = set()
-        for _, app_version in app_versions_df.iterrows():
-            tags |= set(app_version["tags"])
-        tags = sorted(tags)
+        tags = get_unique_app_tags(app_versions_df)
         # select tags
 
         selected_tags = _render_filter_multiselect(
--- a/src/dashboard/trulens/dashboard/utils/dashboard_utils.py
+++ b/src/dashboard/trulens/dashboard/utils/dashboard_utils.py
@@ -2,7 +2,7 @@ import argparse
 import json
 import os
 import sys
-from typing import Any, Callable, Dict, List, Optional, Union
+from typing import Any, Callable, Dict, List, Optional, Set, Union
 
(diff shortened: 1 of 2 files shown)


src/dashboard/trulens/dashboard/utils/dashboard_utils.py
```diff
@@ -508,6 +508,24 @@ def _handle_reset_filters(
             del st.query_params[query_param_key]
 
 
+def get_unique_app_tags(app_versions_df: pd.DataFrame) -> List[str]:
+    """Return the sorted unique app tags for the tag filter multiselect.
+
```
joshreini1, line 513: `tags: set = set()` doesn't constrain what goes in the set. Change it to `tags: Set[str] = set()`. Without the type parameter, a future reader (or type-checker) can't tell whether this holds strings, ints, or something else — and a type-checker won't catch it if you accidentally add a non-string. `Set` is already imported in this file so it's a one-word change.

src/dashboard/trulens/dashboard/utils/dashboard_utils.py
```diff
@@ -508,6 +508,24 @@ def _handle_reset_filters(
             del st.query_params[query_param_key]
```
joshreini1, line 509: Other helpers in this file use a leading underscore for internal functions (e.g. `_handle_reset_filters`, `_render_filter_multiselect`). If this isn't meant to be part of the public API, rename it `_get_unique_app_tags` to match that convention.

joshreini1 (changes requested):
Thanks for fixing this longstanding bug - this is a great and well explained fix.

Two quick nits to address and then it is good to merge.

feiiiiii5:
Thanks @joshreini1! Both nits addressed in `88e6c64`:

- Renamed → `_get_unique_app_tags` (leading underscore, matching internal helper convention)
- Annotated → `tags: Set[str] = set()` (typed annotation, `Set` already imported)

Ready for re-review.

feiiiiii5:
Correction to my last comment: `Set` was **not** already imported in `dashboard_utils.py` — that caused the pre-commit lint failure (`F821 Undefined name Set`) on the last CI run. Fixed in `1ee3650` by adding `Set` to the existing `typing` import line. `ruff check` + `ruff format --check` now pass locally on the touched files.

feiiiiii5:
@joshreini1 gentle ping — both nits were addressed in commits 88e6c64 and 1ee3650 (renamed to `_get_unique_app_tags` + added `Set` import). CI is green. Ready for re-review when you have a moment.

joshreini1 (approved):
Thanks for addressing - fix looks good!

Issue #2627, opened 2026-07-24
Generic LLM-call auto-instrumentation


joshreini1:
## Description

Auto-instrumentation is explicit per-framework `Default.CLASSES`/`METHODS` lists. There's no generic "instrument any LLM call" path — if your framework isn't on the list, you fall back to manual `@instrument`. The existing provider cost tracking already hooks OpenAI/LiteLLM/Google clients; a generalization of that is the foundation.

## Proposed Change

Auto-instrument any LLM call by hooking SDK client classes at the client/completion layer, emitting `gen_ai.*` conventions independent of any framework.

1. Hook completion methods on detected SDK clients (OpenAI, Anthropic, Google GenAI, Bedrock, LiteLLM) and emit `SpanType.GENERATION` spans with `gen_ai.*` attributes + cost. Generalize the existing `_track_costs` mechanism from cost-only to full span emission.
2. Emit content as span **events** (per #2621) and use `SpanKind.CLIENT` (per #2623).
3. Ensure mandatory `gen_ai.system` + `gen_ai.request.model` are always set (a common SDK-wrapper failure).
4. One-line enable: `session.auto_instrument_llm()` (or on by default) instrumenting detected client classes without a framework wrapper.

## Usage Example

```python
from trulens.core import TruSession
session = TruSession()
session.auto_instrument_llm()

# no framework wrapper, no manual @instrument — still produces a GENERATION span
from openai import OpenAI
client = OpenAI()
resp = client.chat.completions.create(model="gpt-4o-mini", messages=[...])
```

## Files to Reference
- `src/core/trulens/core/instruments.py` — `Default.CLASSES`/`METHODS` lists
- `src/core/trulens/experimental/otel_tracing/core/session.py` — existing `_track_costs` (~lines 171-412)

## Acceptance Criteria
- A bare `openai.chat.completions.create` call (no framework, no manual `@instrument`) produces a GENERATION span with `gen_ai.*` attributes + cost + content as events + `SpanKind.CLIENT`. Same for Anthropic/Google/LiteLLM.

## Dependencies
- Coordinates with (but does not hard-depend on) #2621 (GenAI content-as-span-events) and #2623 (span-kinds + resource-attributes) — it consumes their content-event emission and `SpanKind.CLIENT`.

maitraymukeshkumarmodi-aiml:
[...]

PR #2633, opened 2026-07-24
feat: set SpanKind per GenAI conventions and add resource attributes

### LINKED ISSUE #2623: Correct span kinds + resource attributes
## Description

Trulens spans don't set OTel `SpanKind` (all default/INTERNAL) and don't populate standard resource attributes beyond `service.name`. The GenAI conventions RECOMMEND `SpanKind.CLIENT` for outbound LLM calls, and backends use resource attributes (`service.version`, `deployment.environment`) to identify and filter telemetry.

## Proposed Change

1. **`SpanKind.CLIENT`** for `GENERATION` spans (outbound LLM calls over instrumented protocols); `INTERNAL` for in-process work.
2. **Resource attributes**: add `service.version` (app version) and `deployment.environment` alongside the existing `service.name`. (These are OTel export attributes for backend identification — not a dashboard gating concept.)

## Files to Reference
- `src/core/trulens/experimental/otel_tracing/core/span.py` — span builder; set `SpanKind`
- `src/core/trulens/experimental/otel_tracing/core/session.py` — tracer provider `Resource.create({service.name})`; add resource attributes

## Acceptance Criteria
- A GENERATION span has `SpanKind.CLIENT`; the trace resource carries `service.name`, `service.version`, and `deployment.environment`.

## Dependencies
- `SpanKind.CLIENT` is consumed by the generic LLM-call auto-instrumentation issue. Foundational.

Payal2000:
## Summary

Closes #2623

- **Exporter: propagate `span.kind` from SDK spans** — `convert_readable_span_to_proto` and `construct_event` previously hardcoded `SPAN_KIND_INTERNAL`. They now read `span.kind` via a `_SPAN_KIND_TO_PROTO` mapping. Without this change, setting `SpanKind` at span creation would have been silently discarded in both proto export and event construction.
- **Set `SpanKind.CLIENT` for GENERATION spans** — outbound LLM calls now get `SpanKind.CLIENT` per the [OTel GenAI semantic conventions](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/). All other span types remain `SpanKind.INTERNAL`.
- **Virtual/log-ingestion path** — `run.py:879` now passes `span_type=span_type_enum` to `create_function_call_context_manager`, so virtual `generation` spans also get `SpanKind.CLIENT`. Previously `SpanKind` was chosen at span creation but the span type was only applied afterward via `set_general_span_attributes`, which was too late to affect `SpanKind`.
- **Add `service.version` resource attribute** — resolved from `trulens-core` distribution metadata, guarded against `PackageNotFoundError` so tracing still starts in vendored/frozen environments. `ResourceAttributes.APP_VERSION` already carries the user's app version, so `service.version` fills the SDK-version gap without overlap.
- **Add opt-in `deployment.environment`** — set via `TRULENS_DEPLOYMENT_ENVIRONMENT` or `DEPLOYMENT_ENVIRONMENT` env vars. When unset, `Resource.create()` still honours the standard `OTEL_RESOURCE_ATTRIBUTES` variable.
- **Import placement constraint** — `_set_up_tracer_provider()` must resolve its imports at module level. Importing `trulens.core.utils.imports` lazily inside the function changes module-loading order and causes `test_otel_apps_table` to fail when run after the instrument tests (9 vs 10 entries in `tru_class_info.bases`). Noting it here since nothing in the code makes the constraint obvious.

## Test plan

- [x] `test_generation_span_has_client_span_kind` — GENERATION spans get `SpanKind.CLIENT`
- [x] `test_non_generation_span_has_internal_span_kind` — RETRIEVAL and default spans stay `INTERNAL`
- [x] `test_generation_span_exports_as_client_kind_in_proto` — proto conversion carries `SPAN_KIND_CLIENT`
- [x] `test_virtual_generation_span_has_client_kind` — virtual/log-ingested generation spans get `SpanKind.CLIENT`, root spans stay `INTERNAL`
- [x] `test_resource_attributes` — `service.name` and `service.version` present on provider resource
- [x] `test_deployment_environment_absent_by_default` — absent when no env var is set
- [x] `test_deployment_environment_otel_resource_attributes_passthrough` — `OTEL_RESOURCE_ATTRIBUTES=deployment.environment=production` passes through via `Resource.create()`
- [x] `test_otel_apps_table` — no regression in app table serialisation (verified baseline vs branch)
- [x] Golden files — parsed every CSV under `tests/unit/static/golden/`; all span records have `kind=1` (INTERNAL) and no golden test app produces GENERATION spans, so no regeneration needed

## Open question

Should `RETRIEVAL` spans also use `SpanKind.CLIENT`? Kept as `INTERNAL` here to match the acceptance criteria, but I'd lean toward `CLIENT` in a follow-up — vector DB queries are outbound calls over the network, same as LLM calls. Happy either way; flagging rather than expanding scope.

Unblocks #2627, which consumes `SpanKind.CLIENT` for generic LLM-call auto-instrumentation.

### DIFF
--- a/src/core/trulens/core/otel/function_call_context_manager.py
+++ b/src/core/trulens/core/otel/function_call_context_manager.py
@@ -6,6 +6,7 @@ from opentelemetry.baggage import get_baggage
 from opentelemetry.baggage import remove_baggage
 from opentelemetry.baggage import set_baggage
 import opentelemetry.context as context_api
+from opentelemetry.trace import SpanKind
 from opentelemetry.trace import get_current_span
 from opentelemetry.trace.span import Span
 from trulens.experimental.otel_tracing.core.session import TRULENS_SERVICE_NAME
@@ -34,8 +35,13 @@ class UseCurrentSpanFunctionCallContextManager:
 
 
 class CreateSpanFunctionCallContextManager:
-    def __init__(self, span_name: str) -> None:
+    def __init__(
+        self,
+        span_name: str,
+        span_type: str = SpanAttributes.SpanType.UNKNOWN,
+    ) -> None:
         self.span_name = span_name
+        self.span_type = span_type
         self.span_context_manager = None
         self.token = None
         self._started_record = False
@@ -78,11 +84,18 @@ class CreateSpanFunctionCallContextManager:
                 recording.add_record_id(record_id)
 
         self._started_record = started_record
-        # Create span.
+        # Create span.  Use SpanKind.CLIENT for GENERATION spans
+        # (outbound LLM calls) per the OTel GenAI semantic conventions;
+        # all other span types are in-process work → INTERNAL.
+        span_kind = (
+            SpanKind.CLIENT
+            if self.span_type == SpanAttributes.SpanType.GENERATION
+            else SpanKind.INTERNAL
+        )
         self.span_context_manager = (
             trace.get_tracer_provider()
             .get_tracer(TRULENS_SERVICE_NAME)
-            .start_as_current_span(name=self.span_name)
+            .start_as_current_span(name=self.span_name, kind=span_kind)
         )
         span = self.span_context_manager.__enter__()
         if started_record:
@@ -126,11 +139,15 @@ class CreateSpanFunctionCallContextManager:
 
 
 def create_function_call_context_manager(
-    create_new_span: bool, span_name: str
+    create_new_span: bool,
+    span_name: str,
+    span_type: str = SpanAttributes.SpanType.UNKNOWN,
 ) -> Union[
     UseCurrentSpanFunctionCallContextManager,
     CreateSpanFunctionCallContextManager,
 ]:
     if create_new_span:
-        return CreateSpanFunctionCallContextManager(span_name)
+        return CreateSpanFunctionCallContextManager(
+            span_name, span_type=span_type
+        )
     return UseCurrentSpanFunctionCallContextManager()
--- a/src/core/trulens/core/otel/instrument.py
+++ b/src/core/trulens/core/otel/instrument.py
@@ -349,7 +349,9 @@ class instrument:
             span_end_callbacks = kwargs.pop(TRULENS_SPAN_END_CALLBACKS, [])
             func_name_for_call = _func_name_for_instance(instance)
             with create_function_call_context_manager(
-                self.create_new_span, func_name_for_call
+                self.create_new_span,
+                func_name_for_call,
+                span_type=self.span_type,
             ) as span:
                 ret = None
                 func_exception: Optional[Exception] = None
@@ -395,7 +397,9 @@ class instrument:
             span_end_callbacks = kwargs.pop(TRULENS_SPAN_END_CALLBACKS, [])
             func_name_for_call = _func_name_for_instance(instance)
             with create_function_call_context_manager(
-                self.create_new_span, func_name_for_call
+                self.create_new_span,
+                func_name_for_call,
+                span_type=self.span_type,
             ) as span:
                 ret = None
                 func_exception: Optional[Exception] = None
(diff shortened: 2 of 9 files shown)


joshreini1:
Thanks for the detailed writeup here. I reviewed the normal `@instrument(span_type=GENERATION)` path, exporter conversion, resource attrs, and the direct `create_function_call_context_manager(...)` call sites.

The main implementation looks right for standard decorated generation spans. Before this PR, `create_function_call_context_manager(...)` only accepted `(create_new_span, span_name)`, so all spans were created without an explicit OTel kind and therefore defaulted to `INTERNAL`. This PR adds a `span_type` argument and correctly passes `self.span_type` from `instrument.py`, so `@instrument(span_type=SpanAttributes.SpanType.GENERATION)` now creates `SpanKind.CLIENT`, and the exporter now preserves that instead of hardcoding `INTERNAL`.

I found one remaining gap in the virtual/log-ingestion path. It still calls `create_function_call_context_manager(...)` with the old two-argument shape, even though it has already resolved the semantic span type. That means virtual `generation` spans still get created as `INTERNAL` and only receive their `generation` span-type attribute afterward, which is too late to affect `SpanKind`.

I left a targeted inline comment for that. I’d like to see that small follow-up fixed before merging so “GENERATION spans are CLIENT” is true across both normal instrumentation and virtual spans.

src/core/trulens/core/otel/function_call_context_manager.py
```diff
@@ -126,11 +139,15 @@ def __exit__(self, exc_type, exc_val, exc_tb):
 
 
 def create_function_call_context_manager(
```
joshreini1, line 141: This PR correctly handles the normal `@instrument(span_type=GENERATION)` path, but one non-`@instrument` span creation path still misses the new `SpanKind.CLIENT` behavior.

Before this PR, `create_function_call_context_manager(...)` only accepted two args:

```python
create_function_call_context_manager(create_new_span, span_name)
```

This PR adds the new optional `span_type` argument, and `instrument.py` now passes it correctly:

```python
create_function_call_context_manager(
    self.create_new_span,
    func_name_for_call,
    span_type=self.span_type,
)
```

The remaining gap is virtual span creation in `src/core/trulens/core/run.py:879`. That call site still uses the old two-arg form:

```python
with create_function_call_context_manager(
    True, f"{span_type}_virtual"
) as child_span:
    set_general_span_attributes(child_span, span_type_enum)
```

That matters because `SpanKind` is chosen when the OTel span is created. If `span_type == "generation"`, `span_type_enum` is `SpanAttributes.SpanType.GENERATION`, but it is only applied later through `set_general_span_attributes(...)`. At that point the span has already been created with the default `UNKNOWN -> INTERNAL` behavior.

Can you pass `span_type_enum` into the context manager there?

```python
with create_function_call_context_manager(
    True,
    f"{span_type}_virtual",
    span_type=span_type_enum,
) as child_span:
    set_general_span_attributes(child_span, span_type_enum)
```

It would also be useful to add a small test with a virtual/log-ingested `generation.*` field in `dataset_spec` and assert that the resulting readable span/event kind is `SPAN_KIND_CLIENT`. That covers the non-`@instrument` path and prevents this from regressing.

joshreini1:
Future work note, not a blocker for this PR: there are a couple of direct `create_function_call_context_manager(...)` calls in LangGraph stream wrapping that still use the old two-arg form. Those create graph-node spans, so they should remain `INTERNAL` and are not a correctness issue for the `GENERATION -> CLIENT` change.

After this lands, it would be worth doing a small cleanup pass over direct `create_function_call_context_manager(...)` call sites and passing an explicit `span_type` everywhere the semantic span type is already known. That would make the new API harder to misuse and avoid this class of gap if we add more span-kind mapping later.

src/core/trulens/core/otel/function_call_context_manager.py
Payal2000: Fixed in run.py:879 
now passes span_type=span_type_enum. Added test_virtual_generation_span_has_client_kind covering this path.

joshreini1 (approved):
Thank you for the contribution - LGTM!

PR #2636, opened 2026-07-24
fix(openai): extract judge score from custom_tool_call output items (closes #2631)

### LINKED ISSUE #2631: Feedback scores are unreliable for gpt-5 models — score is parsed from the wrong text
**What happens**

When you use any `gpt-5*` model as a judge (e.g. `provider = OpenAI(model_engine="gpt-5-nano")`, feedback functions like `context_relevance` often log a flood of warnings:

```
Rating must be in [0, 3].
Rating must be in [0, 3].
...
Multiple valid rating values found in the string: {"id":"resp_...
```

…and the returned score can simply be wrong. In one call I traced, the model answered `score: 3` but TruLens returned `0`.

**Why it happens**

For `gpt-5*` models, TruLens asks OpenAI for the score through a structured tool call. The model answers correctly — the response contains:

```json
{"type": "custom_tool_call", "name": "base_feedback_json", "input": "{\"score\": 3}"}
```

But when TruLens reads the response back (in `_call_with_capability_fallbacks` in the OpenAI provider), it looks for an item with `type == "tool"`. The OpenAI SDK never uses that name — the type is `"custom_tool_call"`. So TruLens misses the answer, gives up, and dumps the **entire response JSON** into its score parser. The parser then regex-scans that blob and picks a number out of it — sometimes the real score, sometimes a `temperature`, `top_p`, or token count. That's what produces the warnings and the wrong scores.

**Which models are affected**

I ran the same judging call 3 times per model and counted how often the raw blob leaked through:

| Model | Leaks | |
|---|---|---|
| gpt-4o-mini, gpt-4.1-mini, gpt-4.1-nano | 0/3 | ✅ fine (different code path) |
| gpt-5-nano | 2/3 | ❌ |
| gpt-5-mini | 1/3 | ❌ |
| gpt-5.4-nano, gpt-5.4-mini | 3/3 | ❌ |
| gpt-5.6-luna | 2/3 | ❌ |

It's intermittent because it only hits when the model answers via the tool call (vs. plain text) — which makes it easy to miss: everything looks fine until it isn't.

**Reproduce**

```python

provider = OpenAI(model_engine="gpt-5.4-mini")  # any gpt-5* model
provider.context_relevance(
    question="What was revenue in FY2024?",
    context="Revenue was $2.3 billion in FY2024.",
)
# -> warning flood + unreliable score
```

**Versions:** trulens / trulens-providers-openai 2.8.1 (same code on `main`), openai SDK with Responses API, Python 3.12, macOS.

Anikhet:
# Description

Closes #2631.

For all `gpt-5*` judge models, feedback scores were silently unreliable. The provider routes `gpt-5*` models through the Responses API with a grammar-constrained custom tool (`base_feedback_json`), and the model answers correctly — the response contains an output item:

```json
{"type": "custom_tool_call", "name": "base_feedback_json", "input": "{\"score\": 3}"}
```

But the extraction loop in `_call_with_capability_fallbacks` only matched `item.type == "tool"`, a type the OpenAI SDK never emits (the literal is `"custom_tool_call"`, see `openai.types.responses.ResponseCustomToolCall`). With no `output_text` item present either (gpt-5 responses contain only `reasoning` + the tool call), extraction fell through to `response.model_dump_json()` — the entire response object — which was then regex-scanned for a rating. Symptoms: `Rating must be in [0, 3].` warning floods, `Multiple valid rating values found in the string: {"id":"resp_...` warnings, and wrong scores (in one traced call the model returned `score: 3` and TruLens parsed `0`).

Measured leak rates before the fix (same judging call, 3 times per model): `gpt-4o-mini`/`gpt-4.1-mini`/`gpt-4.1-nano` 0/3 (chat-completions path, unaffected); `gpt-5-nano` 2/3; `gpt-5-mini` 1/3; `gpt-5.4-nano` 3/3; `gpt-5.4-mini` 3/3; `gpt-5.6-luna` 2/3. The failure is per-call: it occurs whenever the model answers via the tool call rather than plain text.

The fix accepts both item types in the extraction loop, so the tool-call payload (`{"score": N}`) is returned verbatim and the parser sees exactly one valid rating.

- The existing CFG test dummy (`_DummyResponsesWithCreate` in `tests/unit/providers/test_openai_capabilities.py`) emitted `type = "tool"` — mirroring the code's wrong expectation rather than the real API, which is why tests passed. The dummy now emits the realistic `"custom_tool_call"` type and includes a `reasoning` item before the tool call, matching real gpt-5 response shape.
- Added a parametrized regression test (`test_cfg_extracts_tool_call_item_types`) covering both item types, and an end-to-end test (`test_generate_score_end_to_end_via_cfg_tool_call`) asserting `generate_score` returns the correctly normalized score — the user-visible symptom.
- Verified the new tests fail on the unfixed code and pass with the fix (13→14 tests, all green with `--run_optional_tests`).

## Type of change

- [x] New Tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -298,7 +298,7 @@ test-unit:
 		echo "Attempting to run anyway (expect failures)..."; \
 		$(PYTEST) tests/unit/test_otel*.py; \
 	fi
-	$(PYTEST) $(shell ls tests/unit/test_*.py | grep -v test_otel)
+	$(PYTEST) $(shell ls tests/unit/test_*.py tests/unit/providers/test_*.py | grep -v test_otel)
 # Tests in the e2e folder make use of possibly costly endpoints. They
 # are part of only the less frequently run release tests.
 test-e2e:
--- a/src/feedback/trulens/feedback/llm_provider.py
+++ b/src/feedback/trulens/feedback/llm_provider.py
@@ -3192,6 +3192,10 @@ class LLMProvider(core_provider.Provider):
                     min_score_val=0,
                     max_score_val=1,
                 )
+                # generate_score returns (score, reason) when the judge
+                # responds with structured JSON; compare on the score alone.
+                if isinstance(score, tuple):
+                    score = score[0]
             except Exception:
                 score = 0  # assume not abstention if abstention scoring fails
             return score
@@ -3208,6 +3212,10 @@ class LLMProvider(core_provider.Provider):
                 min_score_val=0,
                 max_score_val=1,
             )
+            # generate_score returns (score, reason) when the judge
+            # responds with structured JSON; compare on the score alone.
+            if isinstance(score, tuple):
+                score = score[0]
             return score
 
         if filter_trivial_statements:
--- a/src/providers/openai/trulens/providers/openai/provider.py
+++ b/src/providers/openai/trulens/providers/openai/provider.py
@@ -201,9 +201,13 @@ class OpenAI(llm_provider.LLMProvider):
                 if hasattr(response, "output"):
                     out_items = getattr(response, "output")
                     for item in out_items:
-                        if getattr(item, "type", None) == "tool" and hasattr(
-                            item, "input"
-                        ):
+                        # Responses API emits custom tool calls with type
+                        # "custom_tool_call" (see openai.types.responses.
+                        # ResponseCustomToolCall); "tool" kept for safety.
+                        if getattr(item, "type", None) in (
+                            "tool",
+                            "custom_tool_call",
+                        ) and hasattr(item, "input"):
                             return getattr(item, "input")
                     texts: list[str] = []
                     for item in out_items:
--- a/tests/unit/providers/test_litellm_capabilities.py
+++ b/tests/unit/providers/test_litellm_capabilities.py
@@ -109,7 +109,10 @@ def test_temperature_success_and_cached(monkeypatch):
 
 @pytest.mark.optional
 def test_reasoning_effort_fallback_and_cached(monkeypatch):
-    provider, dummy = _make_provider(monkeypatch, model_engine="o1-mini")
+    # Use an explicit LiteLLM provider prefix so model routing does not
+    # require a real provider lookup ("o1-mini" alone raises
+    # "LLM Provider NOT provided" in newer litellm versions).
+    provider, dummy = _make_provider(monkeypatch, model_engine="openai/o1-mini")
     dummy.fail_on = {"reasoning_effort": True}
 
     out = provider._create_chat_completion(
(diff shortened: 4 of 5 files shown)


Anikhet:
The plan was a one line fix plus a test, but while testing it I hit a second bug, so there's an extra commit.

Here's what happened: once the extraction fix worked, gpt-5 judges started returning proper scores for the first time, and that exposed a crash further down. The answerability metric assumes generate_score gives back a plain number, but for structured JSON responses it actually returns a (score, reason) pair, so the metric blew up with a TypeError when comparing it to 0.5. The second commit unwraps the score before that comparison.

While I was at it, I found out why the original bug was never caught: CI doesn't run anything in tests/unit/providers/, since the Makefile only collects tests one level up. I added the folder, and made the test mocks match the shape of a real gpt-5 response so this can't silently regress again.

joshreini1:
Thanks for the thorough fix here. The `custom_tool_call` path and the added regression coverage look like the right direction, and I appreciate that you also made the provider tests visible in CI instead of leaving them silently uncollected.

CI looks blocked by that test collection change exposing an existing LiteLLM test issue rather than by the OpenAI regression tests themselves. The failing optional-unit tests are:

- `tests/unit/providers/test_litellm_capabilities.py::test_reasoning_effort_fallback_and_cached`
- `tests/unit/providers/test_litellm_capabilities.py::test_reasoning_effort_success_and_cached`

Both construct `LiteLLM(model_engine="o1-mini")`, which now fails during provider detection with the CI LiteLLM version:

```text
litellm.exceptions.BadRequestError: LLM Provider NOT provided.
You passed model=o1-mini
```

I think the PR-specific path to get CI green is to update those two tests to use an explicit LiteLLM provider prefix, e.g. `model_engine="openai/o1-mini"`, or alternatively monkeypatch `get_llm_provider` if the intent is to keep those tests fully isolated from LiteLLM’s model routing behavior.

There are also unrelated OTEL failures showing up in the basic unit suite, especially:

```text
tests/unit/test_otel_async_concurrency.py::TestOtelAsyncConcurrency::test_cancelled_finalization_failure_preserves_cancelled_error
```

Those files are not touched by this PR, so I’d treat them separately unless they reproduce only on this branch after the LiteLLM test fix.

Anikhet:
@joshreini1 @sfc-gh-jreini thanks for the detailed diagnosis! Pushed a commit fixing both LiteLLM tests with the explicit provider prefix (openai/o1-mini) as you suggested. _is_reasoning_model() parses the id after the /, so they still exercise the reasoning-model path.

Left the OTEL failures alone since they're unrelated to this PR

joshreini1 (approved):
Very valuable fix to the root causes of the problems, and unlocking usage of the latest gpt-5 models including the new gpt-5.6 sol/terra/luna models. Thank you!

Issue #2415, opened 2026-04-23
MCP Tool Instrumentation Example


joshreini1:
## Summary
MCP (Model Context Protocol) is a hot topic but TruLens's MCP support is undiscoverable.

## What
Create `examples/cookbooks/mcp_instrumentation.ipynb` showing how to instrument MCP tool calls and evaluate tool selection/quality.

## Difficulty
Medium

bhargavikalicheti:
Hi @joshreini1  I’d like to work on this. I plan to add a self-contained MCP instrumentation cookbook using a local deterministic MCP server, demonstrate tool-call tracing and evaluation of tool selection/result quality, and include dashboard inspection steps. Should the example use the official MCP Python SDK directly, and is examples/cookbooks/mcp_instrumentation.ipynb still the preferred location?

joshreini1:
Using the official MCP SDK is great, and that location is also good. Thanks!

bhargavikalicheti:
@joshreini1 - Please review the PR.

PR #2638, opened 2026-07-25
feat: mcp tool instrumentation

bhargavikalicheti:
# Description

Adds a self-contained cookbook demonstrating how to trace and evaluate Model
Context Protocol (MCP) tool calls with TruLens.
The example uses the official MCP Python SDK to start a deterministic local MCP
server over stdio, discover its tools through ClientSession, and invoke them
from a LangGraph agent. TruLens records MCP-specific span attributes, including
the server name, tool name, arguments, output, error status, and execution time.
The cookbook also demonstrates Tool Selection and Tool Calling evaluations and
provides instructions for inspecting the resulting trace and evaluation results
in the TruLens dashboard.
This change addresses #2415.
https://github.com/truera/trulens/issues/2415

The local MCP server provides deterministic weather lookup and temperature
conversion tools. It does not require network access or external service
credentials.
OpenAI access is used only by the LangGraph agent and the LLM-based evaluation
metrics. The official MCP server startup, tool discovery, and instrumented tool
invocation were validated independently of the model.
The MCP tools are exposed to LangGraph through thin StructuredTool wrappers.
The underlying calls continue to use the official MCP SDK through
ClientSession.call_tool().

## Type of change

- [ X] New feature (non-breaking change which adds functionality)
  not work as expected)
- [X ] New Tests
- [X ] This change requires a documentation update


joshreini1 (commented):
One improvement to make that would add a lot of value here: `mcp_attributes` receives the return value of `call_mcp_tool`, and `call_mcp_tool` returns only the joined text parts from the MCP result. That means this line records the flattened text output, not the raw MCP SDK response or any structured response metadata:

```python
SpanAttributes.MCP.OUTPUT_CONTENT: str(ret or exception)
```

If the intent is just to record the text passed back to LangGraph, I think the current shape is fine, but the notebook/README should describe it as tool output text rather than raw MCP response metadata.

If the intent is to capture structured MCP response metadata, preserve the SDK result before flattening it and pass both through the instrumented return value. I tested this locally with a no-network instrumentation-only run and the dashboard showed the MCP spans with structured `output_content` as expected.

One possible shape:

```python
from dataclasses import dataclass
import json


@dataclass
class MCPToolCallResult:
    text: str
    raw_result: object

    def __str__(self) -> str:
        # Keeps accidental stringification readable.
        return self.text


def mcp_attributes(ret, exception, *args, **kwargs):
    raw_result = getattr(ret, "raw_result", None)

    if raw_result is not None:
        output_content = json.dumps(
            {
                "is_error": getattr(raw_result, "isError", None),
                "content": [
                    {
                        "type": getattr(content, "type", None),
                        "text": getattr(content, "text", None),
                    }
                    for content in getattr(raw_result, "content", [])
                ],
            },
            default=str,
        )
    else:
        output_content = str(exception) if exception is not None else ""

    return {
        SpanAttributes.MCP.TOOL_NAME: kwargs["tool_name"],
        SpanAttributes.MCP.SERVER_NAME: "local-weather",
        SpanAttributes.MCP.INPUT_ARGUMENTS: json.dumps(
            kwargs["arguments"], default=str
        ),
        SpanAttributes.MCP.OUTPUT_CONTENT: output_content,
        SpanAttributes.MCP.OUTPUT_IS_ERROR: (
            exception is not None
            or bool(getattr(raw_result, "isError", False))
        ),
    }


@instrument(
    span_type=SpanAttributes.SpanType.MCP,
    attributes=mcp_attributes,
)
async def call_mcp_tool(
    *, tool_name: str, arguments: dict
) -> MCPToolCallResult:
    result = await mcp_session.call_tool(tool_name, arguments)
    text_parts = [
        content.text
        for content in result.content
        if getattr(content, "type", None) == "text"
    ]
    return MCPToolCallResult(
        text="\n".join(text_parts),
        raw_result=result,
    )


def as_langchain_tool(mcp_tool):
    async def invoke(**arguments):
        result = await call_mcp_tool(
            tool_name=mcp_tool.name,
            arguments=arguments,
        )
        return result.text

    return StructuredTool.from_function(
        coroutine=invoke,
        name=mcp_tool.name,
        description=mcp_tool.description or "",
        args_schema=mcp_tool.inputSchema,
    )
```

I’d also use `json.dumps(..., default=str)` for `INPUT_ARGUMENTS` instead of `str(...)`, since it makes the trace easier to inspect and more useful to evaluators.

joshreini1:
The pre-commit check is failing because generated formatting/lint changes need to be applied and committed.

Please run this locally from the repository root:

```bash
poetry install --with dev
poetry run pre-commit run --all-files
```

If the hooks modify files, review the diff and commit those changes back to this PR branch:

```bash
git status
git diff
git add <modified-files>
git commit -m "chore: apply pre-commit fixes"
git push
```

After that, the pre-commit CI job should rerun on the updated commit.

bhargavikalicheti:
> `mcp_attributes`

@joshreini1 Thanks for the detailed suggestion. I updated the instrumented return value to preserve the raw MCP SDK result while returning only the joined text to LangGraph. output_content now contains structured JSON with the SDK error status and content entries, and input_arguments uses json.dumps(..., default=str). I also updated the notebook and README descriptions and validated the structured trace path with a no-network MCP run.

bhargavikalicheti:
@joshreini1 Hi! just checking if you are okay with this and ready to merge please?

joshreini1 (approved):
Thank you for the contribution! It is great to have this new detailed example showing deep MCP instrumentation.

PR #2647, opened 2026-07-28
docs: rebuild the trulens.org home page (closes #2413)

### LINKED ISSUE #2413: Refresh trulens.org Home Page
## Summary
The home page at [`docs/overrides/home.html`](https://github.com/truera/trulens/blob/main/docs/overrides/home.html) is a static HTML template that's fallen behind the current feature set.

## Specific Issues
- **Stale feature list**: The "metrics" section lists only RAG-era metrics (Context Relevance, Groundedness, etc.) but doesn't mention agentic evaluations (ToolSelection, PlanAdherence, etc.), MCP instrumentation, or inline evals
- **Missing OTEL story**: The "Interoperable tracing" section is vague — should highlight OpenTelemetry by name with a code snippet or diagram showing span types
- **No provider showcase**: Doesn't list the 7+ supported LLM providers (OpenAI, Anthropic via LiteLLM, Bedrock, Cortex, Google/Gemini, HuggingFace, LangChain)
- **Outdated "How it works" diagram**: The SVG shows a simplified flow that doesn't reflect the current OTEL-based architecture
- **"Use cases" section is generic**: Lists "Agents, RAG, Summarization, Co-pilots" without showing what TruLens actually evaluates for each
- **Footer says "TruLens 2025"**: Should be updated
- **No dashboard screenshots**: The page doesn't show what the Streamlit dashboard looks like
- **Banner links to Feb 2026 blog post**: Should link to latest release or feature

## What
Update the HTML template with current features, add dashboard screenshots, refresh the "How it works" diagram, and modernize the metrics/use-cases sections. Also update the CSS at `docs/stylesheets/style.css` if needed.

## Difficulty
Medium

joshreini1:
## Summary

Rebuilds the trulens.org home page. Closes #2413.

The old page opened on an abstract explainability image and asked the reader to
infer what TruLens does. The new one is a product tour: it names the failure it
finds, shows the actual product doing it, then shows the evidence that the judges
can be trusted. Fifteen bands, four of them live widgets rather than screenshots.

## Before
home-before

## After
home-after

The page is 11,850px against the old 5,317px, and roughly two thirds of the added
height is product: four interactive widgets, a benchmarks band with sourced
numbers, and an integrations wall where every mark links to a working example.

## What changed

**Structure.** Reordered so the argument precedes the product: the problem, then
tracing, then scoring, then the evidence the scores are trustworthy, then
customization, then comparison. Fifteen bands, five hard-edged dividers, no
border radius anywhere.

**Four live widgets replace four static screenshots.** A trace waterfall with
expandable spans and a Tree/Timeline switch, the feedback-results dataframe
behind `st.pills`, the leaderboard with sortable columns and per-version toggles,
and the quality/cost scatter. Each is JS-hydrated into an empty container with a
`<noscript>` screenshot fallback, and the page now renders with scripting off
(the `.reveal` scroll animation previously left it near-blank).

**The widget data is internally consistent, not decorative.** The record root
span aggregates its children, so one record costs $0.01074 and the leaderboard's
ten records cost $0.107 rather than double that. Context relevance is scored once
per retrieved chunk with the pill computed as their mean, so the headline cannot
drift from its rows. The five leaderboard versions form a real Pareto frontier:
v5-mini is 3.8x cheaper than the baseline and scores better, so the untuned
baseline is strictly dominated rather than merely beaten.

**Benchmarks band with four sourced stat cards.** Agent GPA at 95% (267 of 281)
against a 55% baseline, Groundedness F1 of 0.81 against Bespoke-MiniCheck-7B over
an 11,000-example LLM-AggreFact holdout, AIMultiple NDCG@5 of 0.93, and a 4.2:1
adversarial win-loss. Every card links to its source. Where AIMultiple ranks
TruLens second rather than first, the card says so.

**Adopters wall held to an evidence standard.** Every logo links to first-person
evidence, three unsourced marks were removed, and `ADOPTERS.md` records each one
with its source. Snowflake was taken off the wall because it maintains TruLens.

**Integrations wall** rebuilt with real vendor marks, ordered, with every mark
linking to a runnable example rather than to a category page.

**Hero lede** now leads with the failure and the cost pressure rather than the
mechanism: "TruLens finds where your agent fails and where you can cut cost
without losing quality. Open source, OpenTelemetry-native." The `<head>` keeps
the mechanism sentence as its meta description, so OpenTelemetry and
"benchmarked judges" are retained for search without spending hero space.

**`style.css` is 2,071 lines changed and net smaller** — the old page's stylesheet
carried a lot that the new component set does not need.

## Site-level fixes found while reviewing the page

These are in `mkdocs.yml` and affect every page, not just the home page.

`site_url` was never set. MkDocs builds both the sitemap's `<loc>` entries and
Material's `rel="canonical"` from it, so the deployed `sitemap.xml` was a
109-byte empty `<urlset>` and not one page on the site carried a canonical.
Setting it yields 337 sitemap URLs and canonicals on 348 of 351 built pages.

`exclude_docs: overrides/` fixes a duplicate the missing canonicals were hiding.
That directory is the theme's `custom_dir` but also sits under `docs_dir`, so the
raw template was copied out as a static file and
`https://www.trulens.org/overrides/home.html` served a second, near-identical
copy of the home page with no canonical on either.

`site_description` still advertised "Explain Deep Neural Nets", from the
TruLens-Explain era. It is the fallback meta description on every page that does
not set its own.

## Incidental

`docs/reference/` gained the Google provider, TruGraph and TruGEPA, which were
missing from the API reference index. One-line fix to
`docs/evaluation/batch_evaluation.md`.

## Metadata, accessibility and the no-JS fallbacks

Added after the first review pass, all four items that were previously listed here
as follow-ups.

**Head.** The page now has a `rel="canonical"`. Setting `site_url` gave one to
every other page on the site, but this template never extends `base.html`, so
Material's canonical block never ran for it and the most important page was the
one page without one. `og:image` was a root-relative path where OG requires an
absolute URL, so the card fell back to no image; `og:title` was the bare word
"TruLens"; `og:type`, `og:site_name` and `twitter:card` were absent.
`og:description` now carries the same sentence as the hero.

**Social card and README banner.** Shared links and the GitHub repo page both led
with a neural-net explainability diagram from the TruLens-Explain era, correctly
sized for a card at 1200x628 but describing a different product. Both now use a
new 1200x630 card built from this page's own tokens and the same squid-eye mark
that closes band 15. `tools/make_og_card.py` generates it, committed alongside the
PNG because the card has a tagline baked into pixels and would otherwise be an
unreproducible binary the moment that tagline changes. Type is sized for the case
that dominates: an unfurl is often rendered at a third of full size, where a 112px
wordmark survives and body copy does not.

`Neural_Network_Explainability.png` stays in the repo despite now being
unreferenced. `README.md` is the PyPI `long_description` for both the root and
core packages, so that absolute URL is embedded in the published page of every
past release, and deleting the file would retroactively break those.

**Structured data.** A `SoftwareApplication` node with a nested `Organization`
publisher, so the license, repository, language and owner are stated rather than
inferred. Stable facts only, no scores or counts, because anything numeric there
would duplicate a claim rendered further down and drift the moment it is edited.

**Accessibility.** All eleven tab buttons across five tablists had `role="tab"`
and `aria-selected` but no `aria-controls`, and none of the eleven panels was a
tabpanel. Each tab now has an id and `aria-controls`; each panel has a matching
id, `role="tabpanel"`, `aria-labelledby` back to its tab, and `tabindex="0"`. The
three unlabelled tablists gained `aria-label`s. Added a single `<main>` around all
fifteen bands with the nav, announce bar and footer outside it, and a skip link as
the first focusable element. The metrics band's h2-to-h4 jump is fixed, so no
heading skips more than one level. 31 of 71 images now lazy-load; the 40 adopters
marquee images are deliberately left eager, because the track is far wider than
the viewport and scrolls horizontally, so lazy clones would load late or never and
leave gaps in the wall as it travels.

**Fallback screenshots.** All four regenerated. `leaderboard.png` had two rows
named `v1-top1` and `v2-top3`, a column headed Answer Relevance, and no Total Cost
(USD) column; `metrics.png` showed Answer Relevance 1.00 and Context Relevance
0.58. Each shot is now a capture of precisely the element the noscript rule hides,
which differs per widget and the old set did not respect: the leaderboard keeps its
`.dash-bar` with scripting off so its shot is the table alone, while band 5 has no
`.dash-bar` and hides all of `.fb`, so that shot carries its own heading. Captured
at `device_scale_factor` 2 so each renders at exactly half its intrinsic width.
Alt text moved with the images; two still said "two app versions".

## `llms.txt`

Adds `docs/llms.txt`, which MkDocs copies to the site root the same way it already
copies `robots.txt`, so it serves at `/llms.txt` as `text/plain` with no config
change. It follows the llmstxt.org spec: one H1, a blockquote summary, orienting
notes, then H2 file lists of 40 curated links out of the 337 pages in the sitemap.
The spec's `## Optional` section, which consumers may skip when they need a shorter
context, holds the cookbook, the blog and the contributing guides.

Hand-authored rather than generated. `mkdocs-llmstxt` would mean a new docs
dependency and a `poetry.lock` change days after the last one moved, and it emits a
full-markdown dump of selected pages rather than a curated index, which suits this
corpus poorly when roughly a third of it is notebooks. No `llms-full.txt`: the spec
does not define one, and concatenating ~100 notebooks produces a multi-megabyte
file that nothing reliably consumes.

The cost of hand-authoring is drift, so it gets a guard. Nothing today would catch
a stale link, because `lychee` only reads `site/**/*.html` and the strict build does
not look inside static files. `tools/check_llms_txt.py` asserts the spec shape and
pins every internal link against the built `sitemap.xml`, which only became usable
when `site_url` was set earlier in this PR. It runs as a new step in
`ci-docs-pr.yaml` immediately after the build that produces `site/`, and that
pipeline already triggers on any change under `docs/`.

`robots.txt` also gains a `Sitemap:` pointer, worth having now that the sitemap has
337 entries rather than zero.

Worth stating plainly, because it cuts against how I first pitched this: `llms.txt`
is a community convention with no standards-body backing, and as of 2026 no major
model provider has publicly committed to reading it in production. Google says
explicitly that no special AI-readable file is needed for its AI surfaces. The case
for shipping it is that the tools which *do* fetch it skew toward developer tooling,
and that the file doubles as a curated map a human or an agent can paste into a
context window, which a 337-page sitemap cannot do.

## README

Refreshed against the rebuilt page, and four defects fixed that were not just
staleness:

- The Open In Colab badge pointed at `examples/quickstart/langchain_quickstart.ipynb`,
  which does not exist. It points at `quickstart.ipynb` now.
- The contributing link pointed at `/contributing/`, which 404s. There is no bare
  contributing index; the page is `/contributing/development/`.
- The workflow section led with `TruLens_Architecture.png`, the 798 KB "how it
  works" diagram this PR's home page removed as outdated. #2413 listed it as a gap
  and the README was the last place still showing it.
- The install block opened with `pip install trulens-core` while `install.md` says
  `pip install trulens`.

The opening still sold TruLens as a way to "systematically evaluate and track your
LLM experiments", which is the pre-agent positioning, so it now leads with the same
sentence as the hero. Carries over the three things the page has and the README had
none of: the benchmark evidence as a table with every source linked, the adopters by
name, and two real product screenshots (the trace waterfall and the leaderboard) in
place of the removed diagram. Also drops "the TruLens 2.8 Run API" to "the Run API",
since we are on 2.10.

## Known follow-ups, not in this PR

- **Deploy the docs when this merges.** Four README URLs resolve only once this
  lands. Three images (`og-card.png`, `home/trace.png`, `home/leaderboard.png`) need
  the docs site redeployed; the `ADOPTERS.md` link resolves as soon as this is on
  `main`. The docs site is the only mechanism for the images, not a preference:
  `raw.githubusercontent.com` serves a 130-byte LFS pointer as `text/plain` rather
  than the image, and README is the PyPI `long_description` for the root and core
  packages, where relative paths do not resolve either.
- The tabs use `role="tab"` without roving tabindex or arrow-key navigation. Every
  tab is reachable by Tab and activates on Enter or Space, so the pattern is
  functional but not the full ARIA recommendation.
- With scripting off, the three code-block tab rows are visible but inert. Left
  that way deliberately: the default panel is shown and its tab reads as selected,
  so the row still tells the reader other variants exist. That is different from
  the leaderboard's "click a column to sort" hint, which promised an interaction
  and is hidden.
- No `/faq/` page. A page answering "is it open source", "does it work with
  LangGraph", "how does it compare to Ragas or DeepEval" would rank for those
  queries and carry `FAQPage` schema, which cannot live on a page without the
  visible Q&A.
- `.install` in `style.css` is dead now that the hero pill is gone.
- `Chain_Explore.png`, `Evaluations.png` and `Leaderboard.png` are unreferenced,
  but they were already unreferenced on `main`, so they are not this PR's to
  remove.

## Test plan

- [x] Full-page render verified at 1400px, 1024px, 768px and 390px
- [x] Page renders with JavaScript disabled; all four widget fallbacks present and
      no duplicated tab rows
- [x] All four widgets exercised: span expansion, Tree/Timeline switch, metric
      pills, column sort both directions, version toggles
- [x] Every outbound link returns 200, including all four benchmark sources and
      every integration and adopter mark
- [x] Leaderboard cost column sorts correctly and the frontier holds: v5, v2, v3,
      v4 each best at their price, v1 strictly dominated
- [x] `mkdocs build` exits 0; `sitemap.xml` has 337 entries, up from 0
- [x] `rel="canonical"` on 348 of 351 built pages
- [x] `/overrides/` no longer published; built page count 352 to 351
- [x] Home page still renders from the custom template after `exclude_docs`
- [x] All four `<noscript>` screenshots regenerated; verified with JavaScript
      disabled that each loads, each JS container is hidden, and no heading or
      sort hint is duplicated
- [x] Page is 11,905px with JS off against 11,850px with JS, so no fallback is
      mis-sized
- [x] 11 of 11 tabs wired to a `role="tabpanel"` with matching `aria-labelledby`;
      no duplicate ids; all five tab groups still switch
- [x] Skip link is first in the focus order and moves from -9999px to 0 on focus
- [x] Heading outline has no skip greater than one level
- [x] JSON-LD parses; `og:image` absolute; canonical present
- [x] New card serves as `image/png` at 1200x630; `og:image`, width, height and alt
      all render on the page; `tools/make_og_card.py` reproduces it byte-identically
- [x] Every README URL checked: all resolve except the four that require this PR to
      land; no relative links or raw HTML, so the PyPI `long_description` stays valid
- [x] `/llms.txt` serves at the site root as `text/plain`; both external links in it
      return 200
- [x] `make check-llms-txt` passes, and fails as expected on a renamed link, a
      missing blockquote summary, a bullet that is not a markdown link, a second H1,
      and an empty sitemap
- [x] `properdocs build --clean --strict` exits 0 with `llms.txt` present and
      `overrides/` still unpublished

The before/after GIFs above were captured before the metadata and accessibility
commits. Those changes render no differently with JavaScript on, so the GIFs are
still accurate.


A5Wagyu32 (approved):
This is an excellent rework of the site and makes the capabilities of TruLens so much more clear! Thank you for doing this.

Issue #2630, opened 2026-07-24
Sampling for online evaluation + clearer APIs


joshreini1:
## Description

By default, trulens computes metrics immediately after each trace is logged — so evaluation is already online and continuous. The gap is on high-traffic production apps: evaluating every trace is costly, and there's no easy sampling path to evaluate only a fraction of logged traces, nor clear APIs for controlling online evaluation.

## Proposed Change

1. **Sampling config**: a sampling policy (`sample_rate`, optional filter, throttle, cost budget) so you can evaluate only a fraction of logged traces — e.g. score 10% of production traces to control cost — applied to the existing immediate-after-log metric computation.
2. **Clearer APIs**: a clear, explicit entry point for enabling/disabling online evaluation and configuring sampling (today the immediate-eval default is implicit).
3. **Sampling context on results**: evaluated results carry the sampling context (`sample_rate`, sampled flag) so coverage is visible in the dashboard.
4. **Guardrails**: throttling so a high-traffic app doesn't overwhelm eval; per-app sampling rates.
5. **Cost budget — only when costs are actually tracked.** `cost_budget` is only enforceable when the metrics' provider reports tracked costs (OpenAI, LiteLLM, Google, Cortex today; other providers may not report costs). If `cost_budget` is set but a configured metric's provider does not report tracked costs, emit an **obvious, prominent warning** naming the metric + provider (at `configure_online_eval` time and in the dashboard) so the user knows the budget won't be enforced for that metric — do not silently ignore it.
6. **Alerting integration**: sampled online results can trigger alerts (tie to #2619's alerting viz).

## Usage Example

```python
from trulens.core import TruSession, Metric
from trulens.providers.openai import OpenAI
from trulens.apps.app import TruApp

session = TruSession()
provider = OpenAI()

# Metrics — by default these are computed immediately after each trace is logged
f_groundedness = (
    Metric(provider.groundedness_measure_with_cot_reasons, name="Groundedness")
    .on_input()
    .on_output()
)
f_relevance = (
    Metric(provider.relevance_with_cot_reasons, name="Answer Relevance")
    .on_input()
    .on_output()
)

# Wrap the app. By default every invocation evaluates all metrics right after the trace logs.
tru_app = TruApp(
    my_rag_app,                         # your RAG / app object
    app_name="prod_rag",
    app_version="v1",
    feedbacks=[f_groundedness, f_relevance],
)

# On a high-traffic app, evaluate only a sample of traces to control cost.
[...]

Issue #2619, opened 2026-07-24
Leaderboard: trends, significance, drift, alerting (all apps by default)


joshreini1:
## Description

The Leaderboard shows per-app-version aggregates + histograms. There's no time-trend view, no signal-vs-noise on score changes, no regression detection, and limited alerting visualization. (Some smoothed-bands + faded-points work is in flight on `josh/openai-agent-sdk-snowflake-tools` / `jreini/dashboard-query-perf-2372`; this issue scopes the full vision and reconciles with that work.)

## Proposed Change

**On for all apps by default — zero config.** A user iterating on any app version immediately sees whether v2 beat v1, without marking anything as "production." No environment/production concept is introduced: trends, significance, drift, and sweet-spot are always-on; alerting is simply per-app threshold configuration. Non-Snowflake (SQLite/Postgres) users benefit most, since the local dashboard is being deprecated for Snowflake users in favor of Snowsight.

1. **Trend view**: per-metric time-series (uses existing `get_records_and_feedback`); smoothed bands (reuse in-flight work) with faded individual points; toggle between per-app-version and per-conversation aggregation.
2. **Statistical significance / regression detection**: turn "score went up" into "score changed by Δ with a 95% confidence interval." Compute CIs from sample size + standard error per metric per app version; show score deltas between app versions and highlight individual records that regressed beyond a tolerance.
3. **Drift / sweet-spot detection**: rolling mean + variance bands; mark "sweet-spot" app versions (best score vs cost/latency); drift indicator on rows.
4. **Alerting viz**: per-app threshold configuration rendered as horizontal bands on that app's trend charts; flag rows breaching thresholds. Thresholds are the opt-in — no environment gating.
5. **Perf**: build on SQL-level aggregation (`get_leaderboard_aggregates`, PR #2372); add a time-bucketed + CI variant so trend queries stay cheap.

## Files to Reference
- `src/dashboard/trulens/dashboard/tabs/Leaderboard.py` — leaderboard tab
- `src/core/trulens/core/database/base.py` — `get_leaderboard_aggregates`, `get_records_and_feedback`

## Acceptance Criteria
- Leaderboard shows per-metric trends with CIs, score deltas + regression highlighting, drift/sweet-spot markers, and per-app alert bands for every app by default; degrades gracefully under SIS compatibility; trend queries use SQL aggregation. No environment/production mark is required.

## Dependencies
[...]

PR #2649, opened 2026-07-30
feat(core): add sampling for online evaluation

Payal2000:
## Summary

- Add `SamplingConfig` (frozen Pydantic model) and `SamplingController` (mutable, thread-safe) for controlling automatic post-ingest evaluation on high-traffic apps
- Add `TruSession.configure_online_eval()` API with `sample_rate` (global or per-app dict), `throttle` (max evals/min), and `cost_budget` (daily USD cap)
- Gate only the ingest path — explicit `compute_metrics()`/`compute_now()` always evaluate everything
- Deterministic hash-based sampling (SHA-256 of record_id salted with app_name) — idempotent across retries/replays
- `EVAL_DECISION` span type emitted for in-scope records (both evaluated and skipped) so `get_records_and_feedback()` projects `sampled`, `sample_rate`, `eval_decision_reason` columns
- Out-of-scope apps evaluate normally with no span overhead
- Batch backfills don't charge the daily cost budget (`contextvars.ContextVar` propagated via `copy_context()` at `ThreadPoolExecutor` dispatch)
- `reports_costs` property on `Provider` base (default `False`), overridden to `True` on OpenAI/LiteLLM/Google/Cortex

## New public API

- `trulens.core.SamplingConfig`
- `TruSession.configure_online_eval(sample_rate, throttle, cost_budget, feedbacks)`
- `TruSession.sampling_controller` (property + setter for test injection)
- `Provider.reports_costs` (property)

## Open acceptance criteria (follow-up)

- **Dashboard**: coverage visibility and cost-warning display — requires dashboard/frontend changes not in this PR
- **Alerting integration**: soft-coupled to #2619

## Test plan

- [x] 37 pure-Python unit tests: config validation, deterministic hashing, probabilistic sampling, per-app rates, throttle, cost budget, gate ordering, counters, context var propagation through ThreadPoolExecutor
- [x] 20 OTEL integration tests: configure_online_eval lifecycle, evaluator sampling gate, batch path exempt, skip path emits decision spans, out-of-scope app evaluates normally, decision span projection through DB round-trip (`sampled`/`sample_rate`/`eval_decision_reason` columns), null semantics for unconfigured sessions, batch path doesn't charge budget, controller injection via setter
- [x] Zero regressions on existing `test_otel_evaluator` and `test_otel_tru_session` tests
- [x] All pre-commit hooks pass (ruff, ruff-format)

Refs #2630

### DIFF
--- a/src/core/trulens/core/utils/evaluator.py
+++ b/src/core/trulens/core/utils/evaluator.py
@@ -11,7 +11,11 @@ import weakref
 
 import pandas as pd
 from trulens.core.otel.utils import is_otel_tracing_enabled
+from trulens.core.sampling import EvalDecisionReason
+from trulens.core.sampling import SamplingController
+from trulens.core.sampling import ingest_eval_active
 from trulens.core.session import TruSession
+from trulens.otel.semconv.trace import ResourceAttributes
 from trulens.otel.semconv.trace import SpanAttributes
 
 if TYPE_CHECKING:
@@ -190,7 +194,53 @@ class Evaluator:
                 record_ids,
                 self._processed_time,
             )
+
+            # Sampling is only applied on the automatic ingest path
+            # (in_evaluator_thread=True).  Explicit compute_now() calls
+            # always evaluate everything.
+            controller: Optional[SamplingController] = None
+            if in_evaluator_thread:
+                controller = TruSession()._sampling_controller
+
             for record_id, events in record_id_to_events.items():
+                # --- sampling gate (ingest path only) ---
+                if controller is not None:
+                    should_eval, sampling_meta = controller.should_evaluate(
+                        record_id=record_id,
+                        app_name=self._app_name,
+                    )
+                    reason = sampling_meta.get("eval_decision_reason")
+
+                    # NOT_CONFIGURED means sampling doesn't apply to
+                    # this app — fall through to evaluate with no span
+                    # overhead.  Only emit decision spans for records
+                    # that are actually in scope for sampling.
+                    if reason != EvalDecisionReason.NOT_CONFIGURED.value:
+                        _emit_sampling_decision_span(
+                            record_id=record_id,
+                            app_name=self._app_name,
+                            app_version=self._app_version,
+                            events=events,
+                            sampling_meta=sampling_meta,
+                        )
+
+                    if not should_eval:
+                        logger.debug(
+                            "Skipping evaluation for record_id=%s: %s",
+                            record_id,
+                            reason,
+                        )
+                        self._record_id_to_event_count[record_id] = len(events)
+                        continue
+
+                # Set the ingest flag so record_cost() in computer.py
+                # only charges the budget on the ingest path.  A batch
+                # backfill must not burn the daily budget.
+                token = (
+                    ingest_eval_active.set(True)
+                    if controller is not None
+                    else None
+                )
                 try:
                     self._app_ref().compute_feedbacks(
                         raise_error_on_no_feedbacks_computed=False,
@@ -201,6 +251,8 @@ class Evaluator:
                         f"Error computing feedbacks in evaluator thread (record_id={record_id}): {e}\n{traceback.format_exc()}"
                     )
                 finally:
+                    if token is not None:
+                        ingest_eval_active.reset(token)
                     self._record_id_to_event_count[record_id] = len(events)
                     TruSession().force_flush()
                 if in_evaluator_thread and self._stop_event.is_set():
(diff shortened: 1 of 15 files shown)


joshreini1:
Thank you for contributing! Very excited about this addition.

I think there is an important state-modeling issue here around sampled-out records and explicit backfill.

The PR says explicit `compute_metrics()` / `compute_now()` calls are not gated by sampling, but skipped records currently appear to be marked as processed in the same internal bookkeeping used for completed records. In `Evaluator._compute_feedbacks`, the sampled-out path sets `_record_id_to_event_count[record_id] = len(events)` before continuing. Later, even when `compute_now(record_ids=...)` is called explicitly, `_get_record_id_to_unprocessed_events` still filters out records whose event count has not increased. So a record that was `not_sampled` by the automatic evaluator can become unreachable for explicit feedback computation.

I reproduced this with a small local demo against this branch:

```text
Sampling counters: {'evaluated': 5, 'not_configured': 0, 'not_sampled': 7, 'throttled': 0, 'over_budget': 0}
Skipped before compute_now: 7
Rows missing feedback after compute_now: 7
Events: 58
Eval decision spans: 12
```

I think the automatic evaluator needs to distinguish at least two concepts:

- the evaluator has already seen this record/event-count and should not repeatedly reconsider it on the automatic ingest path
- feedback was actually computed for this record, vs skipped due to `not_sampled`, `throttled`, or `over_budget`

Sampling decisions should suppress repeated automatic work, but they should not permanently mark a record as ineligible for explicit backfill. Operationally, users will likely sample live traffic cheaply, then later backfill evaluations for skipped records when they see a regression, want higher confidence on a slice, or temporarily want to evaluate at a higher sample rate.

Concretely, `record_ids is not None` seems like it should be a force path: fetch those records and compute missing feedbacks regardless of the automatic evaluator's seen/skipped bookkeeping. Otherwise the public contract should be narrowed from “explicit compute always evaluates everything” to “explicit compute only evaluates records the automatic evaluator has not already marked processed,” which feels like the less useful behavior.

Payal2000:
> Thank you for contributing! Very excited about this addition.
> 
> I think there is an important state-modeling issue here around sampled-out records and explicit backfill.
> 
> The PR says explicit `compute_metrics()` / `compute_now()` calls are not gated by sampling, but skipped records currently appear to be marked as processed in the same internal bookkeeping used for completed records. In `Evaluator._compute_feedbacks`, the sampled-out path sets `_record_id_to_event_count[record_id] = len(events)` before continuing. Later, even when `compute_now(record_ids=...)` is called explicitly, `_get_record_id_to_unprocessed_events` still filters out records whose event count has not increased. So a record that was `not_sampled` by the automatic evaluator can become unreachable for explicit feedback computation.
> 
> I reproduced this with a small local demo against this branch:
> 
> ```
> Sampling counters: {'evaluated': 5, 'not_configured': 0, 'not_sampled': 7, 'throttled': 0, 'over_budget': 0}
> Skipped before compute_now: 7
> Rows missing feedback after compute_now: 7
> Events: 58
> Eval decision spans: 12
> ```
> 
> I think the automatic evaluator needs to distinguish at least two concepts:
> 
> * the evaluator has already seen this record/event-count and should not repeatedly reconsider it on the automatic ingest path
> * feedback was actually computed for this record, vs skipped due to `not_sampled`, `throttled`, or `over_budget`
> 
> Sampling decisions should suppress repeated automatic work, but they should not permanently mark a record as ineligible for explicit backfill. Operationally, users will likely sample live traffic cheaply, then later backfill evaluations for skipped records when they see a regression, want higher confidence on a slice, or temporarily want to evaluate at a higher sample rate.
> 
> Concretely, `record_ids is not None` seems like it should be a force path: fetch those records and compute missing feedbacks regardless of the automatic evaluator's seen/skipped bookkeeping. Otherwise the public contract should be narrowed from “explicit compute always evaluates everything” to “explicit compute only evaluates records the automatic evaluator has not already marked processed,” which feels like the less useful behavior.

This is fixed in the latest push (51a139e). Sampled-out records now go into a separate _sampled_out_record_ids set instead of _record_id_to_event_count, and compute_now(record_ids=...) bypasses both filters with force=True. Dedup is handled downstream by _remove_already_computed_feedbacks. Two regression tests cover the backfill path and the automatic-path filtering.

sfc-gh-jreini (changes requested):
Thank you for contributing @Payal2000 - this is a very powerful addition for monitoring use cases.

The implementation and tests look good.

Can you document the new capabilities in: https://github.com/truera/trulens/blob/main/docs/component_guides/evaluation/running_feedback_functions/with_app.md

Excited to get this in the hands of agent builders!

Payal2000:
> Thank you for contributing @Payal2000 - this is a very powerful addition for monitoring use cases.
> 
> The implementation and tests look good.
> 
> Can you document the new capabilities in: https://github.com/truera/trulens/blob/main/docs/component_guides/evaluation/running_feedback_functions/with_app.md
> 
> Excited to get this in the hands of agent builders!

Added a sampling section to that doc covering configure_online_eval(), per-app rates, throttle/cost budget, coverage inspection via get_records_and_feedback(), and backfilling skipped records.

src/core/trulens/core/utils/evaluator.py
```diff
@@ -186,11 +212,66 @@ def _compute_feedbacks(
...
+                        )
+
+                    if not should_eval:
+                        logger.debug(
+                            "Skipping evaluation for record_id=%s: %s",
+                            record_id,
+                            reason,
+                        )
+                        # Track as sampled-out, NOT as processed.
+                        # This lets compute_now() still reach these
```
joshreini1, line 262: The fix for the state-modeling issue is correct: sampled-out records are tracked in `_sampled_out_record_ids` and skipped from `_record_id_to_event_count`, making them reachable by explicit `compute_now(record_ids=...)`. Well done.

tests/unit/test_otel_sampling.py
```diff
@@ -0,0 +1,498 @@
...
+        record_ids = ["record_A", "record_B", "record_C"]
+        events_map = {rid: _make_mock_events(rid) for rid in record_ids}
+
+        # Step 1: automatic evaluator skips all records.
+        evaluator._get_record_id_to_unprocessed_events = MagicMock(
+            return_value=events_map
+        )
+        evaluator._compute_feedbacks(in_evaluator_thread=True)
+        mock_app.compute_feedbacks.assert_not_called()
+
```
joshreini1, line 230: `TestSampledOutRecordsReachableViaComputeNow` is an excellent regression test for the state-modeling issue raised in the discussion. The test explicitly verifies sampled-out records remain in `_sampled_out_record_ids` (not `_record_id_to_event_count`) and are reachable via `compute_now`.

joshreini1 (approved):
Strong implementation with excellent test coverage (37 unit + 20 OTEL tests). The state-modeling fix properly separates sampled-out records from processed records via `_sampled_out_record_ids`, making explicit backfills work correctly. Cost budget enforcement via context vars is elegant.

sfc-gh-jreini (approved):

PR #2653, opened 2026-07-31
fix(feedback): validate structured-JSON scores against the configured scale

feiiiiii5:
## Root Cause

The structured-JSON fast paths in `LLMProvider.generate_score` and `generate_score_and_reasons` normalize `parsed_json["score"]` without validating it against the configured `[min_score_val, max_score_val]` scale. The plain-string path already enforces this via `re_configured_rating` (out-of-range → `ParseError` — pinned by `test_generate_score_raises_parse_error[out_of_range]`), so the two response shapes disagree:

| response (scale 0-10) | before | after |
|---|---|---|
| `"42"` (string) | `ParseError` | `ParseError` (unchanged) |
| `{"score": 42}` | **silently returns 4.2** | `ParseError` |
| CoT JSON `score=100` | **silently returns 10.0** | `ParseError` |
| `[{"score": 2}, {"score": 42}]` | mean folds in 42 → 2.2 | 42 skipped w/ warning → 0.2 |
| CoT JSON `score="abc"` | sentinel −1.0 normalized to **−0.1** | raw `-1.0` (matches `generate_score`) |

Out-of-scale scores escape the documented [0, 1] output range and silently corrupt downstream aggregation (leaderboard means, threshold checks). The `-1.0` failure sentinel was additionally being normalized only in `generate_score_and_reasons`, diverging from `generate_score`.

## Fix

- Add module-private `_validate_score_range` mirroring `re_configured_rating`'s range semantics; apply it on both scalar JSON paths.
- List-averaging path skips out-of-range items with the same warning the string parser logs, instead of folding them into the mean.
- Normalize only on successful parse in `generate_score_and_reasons`, so the `-1.0` sentinel is returned raw.

No public API surface change (helper is module-private).

## Test

- [x] New `tests/integration/test_score_range_validation.py` (7 tests): out-of-range scalar JSON raises on both methods (above/below), in-range still normalizes, list filtering, sentinel regression.
- [x] Existing `tests/integration/test_score_parsing_normalization.py` + `tests/unit/test_feedback_score_generation.py`: 34 passed, 1 xfailed (pre-existing strict xfail for #2496) — no regressions.
- [x] `ruff check` / `ruff format --check` clean on touched files.

## Diff scope

2 files, +197/−13 (fix is ~40 lines; the rest is tests).

## AI Disclosure

AI assistance was used for candidate-pattern scanning and test drafting; the root-cause analysis, fix design (mirroring the existing string-path semantics), and all verification (local repro of the 4 cases above + test runs) were human-reviewed. Happy to adjust anything.

### DIFF
--- a/src/feedback/trulens/feedback/llm_provider.py
+++ b/src/feedback/trulens/feedback/llm_provider.py
@@ -229,9 +229,9 @@ class LLMProvider(core_provider.Provider):
         """
 
         assert self.endpoint is not None, "Endpoint is not set."
-        assert max_score_val > min_score_val, (
-            "Max score must be greater than min score."
-        )
+        assert (
+            max_score_val > min_score_val
+        ), "Max score must be greater than min score."
 
         llm_messages = [{"role": "system", "content": system_prompt}]
         if user_prompt is not None:
@@ -342,9 +342,9 @@ class LLMProvider(core_provider.Provider):
                 reason metadata dictionary.
         """
         assert self.endpoint is not None, "Endpoint is not set."
-        assert max_score_val > min_score_val, (
-            "Max score must be greater than min score."
-        )
+        assert (
+            max_score_val > min_score_val
+        ), "Max score must be greater than min score."
 
         llm_messages = [{"role": "system", "content": system_prompt}]
         if user_prompt is not None:
--- a/src/feedback/trulens/feedback/llm_provider.py
+++ b/src/feedback/trulens/feedback/llm_provider.py
@@ -38,6 +38,27 @@ logger = logging.getLogger(__name__)
 
 REASONING_MODEL_PREFIXES = ("o1", "o3", "o4", "gpt-5", "deepseek-r1")
 
+
+def _validate_score_range(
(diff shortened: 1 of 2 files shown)


joshreini1 (approved):
Good catch and clean fix. The structured-JSON paths were indeed inconsistent with the string-response path.

A few notes:

1. ** list branch (lines ~284-297):** You chose to  and silently drop out-of-range items rather than raise . This is a reasonable tradeoff — raising on an out-of-range item in a list would discard all other valid scores — but it's a behavioural asymmetry from the scalar paths worth calling out in a comment. Consider adding a short docstring note or inline comment.

2. ** restructuring:** The refactoring (moving normalisation into the `else` branch) is a neat improvement — avoids the -1.0 sentinel leaking into a valid normalisation call.

3. **Test location:** The test file lives under `tests/integration/` but does not call any external service — it's effectively a unit test. Consider moving it to `tests/unit/` so it runs in the normal `make test-unit` sweep without extra infra.

Approving — the core fix is correct and well-tested.

PR #2661, opened 2026-08-04
Fix release version derivation

joshreini1:
## Summary
- remove the release bump selector and derive the exact target version from `releases/rc-trulens-X.Y.Z`
- reject malformed branch names and targets that do not advance the current version
- document the direct-push branch protection requirements

## Test plan
- [x] Parse the Azure Pipeline YAML
- [x] Run pre-commit on all changed files
- [x] Verify patch, minor, major, and arbitrary future release branches
- [x] Verify malformed and non-release branches are rejected


sfc-gh-jreini:
redundant, making these changes directly in 2.11 release branch

PR #2672, opened 2026-08-05
fix(groundedness): fence statement and stop instructing judge to copy it verbatim

AUTHENSOR:
## Summary

The Groundedness user_prompt template (`templates/rag.py:66`) instructs the judge to set the `criteria` key to a verbatim copy of the STATEMENT. Since `STATEMENT: {hypothesis}` is the model-under-test output interpolated via `str.format`, the judge is guaranteed to reproduce attacker-controlled text in its JSON output. Combined with the JSON `score` key being trusted directly, a crafted answer can influence its own groundedness score.

## Fix (2 lines changed)

1. Wrap the hypothesis in `<statement>...</statement>` tags so the judge treats it as data.
2. Change `criteria: copy of the STATEMENT verbatim` to `criteria: a short label for the statement (do NOT copy it)`.

## Verification

- `ruff check`: all checks passed
- 1 file changed, 2 insertions, 2 deletions

### DIFF
--- a/src/feedback/trulens/feedback/templates/rag.py
+++ b/src/feedback/trulens/feedback/templates/rag.py
@@ -66,10 +66,10 @@ class Groundedness(Semantics, WithPrompt, CriteriaOutputSpaceMixin):
     user_prompt: ClassVar[str] = cleandoc(
         """SOURCE: {premise}
 
-        STATEMENT: {hypothesis}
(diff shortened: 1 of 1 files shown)


daniel-huang-1230 (changes requested):
AI slop that does not actually address the issue it claimed. - lets close this @sfc-gh-jreini

PR #2655, opened 2026-08-01
feat(feedback): Conversation-aware metric templates (#2624)

maitraymukeshkumarmodi-aiml:
#### Title:                                         
                                                  
feat(feedback): Conversation-aware metric         
templates (#2624)                                   
                                                  
### One-line summary for changelog:               
Add conversation-aware metric templates for multi-
turn chat evaluation (#2624)                        
                                                  
### Meaningful description:                       
This PR resolves #2624 by introducing multi-turn  
conversation evaluation metrics built against multi-
turn transcripts (`List[Record]` / turn             
dictionaries).                                      
                                                  
#### Key Additions:                               
1. **Transcript Renderer & Templates**: Created   
`src/feedback/trulens/feedback/templates/conversatio
n.py` providing:                                    
   - `conversation_to_prompt()`: Serializes       
`Record` objects and turn dicts into structured     
Human/Assistant/Tool transcripts.                   
   - `ConversationHelpfulness`: Evaluates overall 
helpfulness across turns (0-3 Likert).              
   - `SimpleCriteriaScore`: Evaluates multi-turn  
sessions against custom criteria (0-3 Likert).      
   - `TopicAdherence`: Evaluates topic precision, 
recall, and F1 score against reference topics.      
   - `AgentGoalAccuracy`: Evaluates binary goal   
completion (0/1) for agentic conversations.         
   - `CoherenceAcrossTurns`: Evaluates dialogue   
consistency and context retention across turns (0-3 
Likert).                                            
2. **LLMProvider Wiring & Re-exports**:           
   - Added provider methods                       
(`conversation_helpfulness`, `simple_criteria_score`,
`topic_adherence`, `agent_goal_accuracy`,           
`coherence_across_turns`) to `LLMProvider`.         
   - Re-exported conversation templates in        
`templates/__init__.py`.                            
3. **Unit Tests**:                                
   - Added                                        
`tests/unit/test_templates_conversation.py` covering
transcript formatting, provider invocations, and    
template structures (`6/6 PASSED`).                 
                                                  
### Checklist                                     
- [x] AI was used in creating this PR

PR #2673, opened 2026-08-06
fix(feedback): unwrap generate_score tuple in float-returning feedbacks

WatchTree-19:
### what

`generate_score` returns `float | tuple[float, dict]`: a `(score, reason)` tuple when the judge answers with structured json, a bare float otherwise. its docstring already tells callers to unwrap:

```python
score = result[0] if isinstance(result, tuple) else result
```

`evaluate_abstention` and `evaluate_answerability` do this. seven methods annotated `-> float` don't, so they return the raw tuple whenever a structured-output provider is used (the default `_create_chat_completion` path returns json text):

- `context_relevance`
- `citation_accuracy`
- `relevance`
- `sentiment`
- `stereotypes`
- `citation_attribution`
- `_langchain_evaluate` (backs the langchain criteria family: conciseness, correctness, coherence, harmfulness, ...)

so those feedbacks hand back `(0.8, {...})` where a float is expected, which breaks downstream aggregation/thresholding.

### repro

with `generate_score` returning a structured-json tuple, on current main:

```
citation_attribution(...)  ->  (1.0, {'reason': 'cited'})   # tuple
```

after this change:

```
citation_attribution(...)  ->  1.0                          # float
```

### fix

small `_extract_score` staticmethod, all nine call sites routed through it (the two existing inline guards folded in too). regression test added in `test_citation_attribution`. pinned ruff (0.5.5) format/check and the touched unit tests pass locally.

note: #2655 adds a few more `return self.generate_score(...)` methods with the same shape that would want the same wrap. the root cause is the dual return on `generate_score`; happy to instead make it always return a bare float and drop the tuple contract if you prefer that direction.

### DIFF
--- a/src/feedback/trulens/feedback/llm_provider.py
+++ b/src/feedback/trulens/feedback/llm_provider.py
@@ -206,7 +206,7 @@ class LLMProvider(core_provider.Provider):
         min_score_val: int = 0,
         max_score_val: int = 10,
         temperature: float = 0.0,
-    ) -> float | tuple[float, dict]:
+    ) -> float:
         """
         Base method to generate a score normalized to 0 to 1, used for evaluation.
 
@@ -218,15 +218,8 @@ class LLMProvider(core_provider.Provider):
             temperature (float): The temperature for the LLM response.
 
         Returns:
-            float | tuple[float, dict]: The normalized score on a 0-1 scale.
-            When the LLM returns a structured JSON response containing a ``score``
-            key, returns a ``(score, reason_dict)`` tuple where ``reason_dict``
-            contains the raw parsed JSON under a ``"reason"`` key. When the
-            response is a plain string, returns a bare ``float``. Callers should
-            handle both forms, e.g.::
-
-                result = provider.generate_score(system_prompt="...")
-                score = result[0] if isinstance(result, tuple) else result
+            float: The normalized score on a 0-1 scale. If reasons are needed,
+            use :meth:`generate_score_and_reasons` instead.
         """
 
         assert self.endpoint is not None, "Endpoint is not set."
@@ -277,7 +270,7 @@ class LLMProvider(core_provider.Provider):
                     max_score_val - min_score_val
                 )
 
-            return normalized_score, {"reason": parsed_json}
+            return normalized_score
 
         if isinstance(parsed_json, list):
             # If a list is returned, average the scores where possible.
@@ -303,7 +296,7 @@ class LLMProvider(core_provider.Provider):
                 )
             else:
                 normalized_score = -1.0
-            return normalized_score, {"reason": parsed_json}
+            return normalized_score
 
         if isinstance(response, feedback_output_schemas.BaseFeedbackResponse):
             score = response.score
@@ -3610,10 +3603,6 @@ class LLMProvider(core_provider.Provider):
                     min_score_val=0,
                     max_score_val=1,
                 )
-                # generate_score returns (score, reason) when the judge
-                # responds with structured JSON; compare on the score alone.
-                if isinstance(score, tuple):
-                    score = score[0]
             except Exception:
                 score = 0  # assume not abstention if abstention scoring fails
             return score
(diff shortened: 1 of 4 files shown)


sfc-gh-jreini (dismissed):
Clean, targeted fix. `_extract_score` static helper is the right abstraction — normalizes the `float | tuple[float, dict]` contract in one place and the seven affected methods all use it consistently. The existing inline `isinstance` checks in `evaluate_abstention` / `evaluate_answerability` are correctly replaced with the helper rather than left as duplicates.

Regression test is well-constructed — wiring `LLMProvider._extract_score` onto the mock ensures the real unwrapping logic is exercised, not a stub.

Thanks for the contribution!

sfc-gh-jreini (dismissed):
Clean fix. _extract_score helper is the right abstraction — normalizes the float | tuple[float, dict] contract in one place. All seven affected methods use it consistently, and the existing inline isinstance checks in evaluate_abstention/evaluate_answerability are correctly replaced rather than left as duplicates. Regression test is solid — wiring LLMProvider._extract_score onto the mock exercises the real unwrapping logic. Thanks for contributing!

sfc-gh-jreini (changes requested):
Thanks for the detailed bug report and the offer at the end of your description — "happy to instead make it always return a bare float and drop the tuple contract" — that is the right direction.

The root cause is that generate_score has an inconsistent return type (float | tuple[float, dict]). Adding _extract_score and patching all seven callers works, but it treats the symptom rather than the source. The existing pattern in this codebase is:

- foo() calls generate_score() and returns float
- foo_with_cot_reasons() calls generate_score_and_reasons() and returns Tuple[float, Dict]

citation_attribution (and the other six methods) should follow the same split: generate_score returns a plain float, and if you need the reasons you call generate_score_and_reasons.

Preferred fix: update generate_score to always return a bare float. The structured-JSON path that can produce a tuple should either (a) silently drop the reason dict and return only the score, or (b) route callers that want reasons to generate_score_and_reasons instead. Either way the return type becomes float unconditionally.

Please revert the _extract_score addition and the seven call-site changes, and fix generate_score directly. The regression test pattern is still good — update it to verify generate_score returns a bare float even when the provider returns structured JSON.

sfc-gh-jreini (changes requested):
The latest commit (c2bbc34) only updates the test mock binding — it doesn't address the core feedback from my previous review.

To reiterate: the fix I'm looking for is to make `generate_score` always return a bare `float`. The `_extract_score` wrapper at all seven call sites treats the symptom, not the root cause.

The existing pattern in this codebase is:
- `foo()` → calls `generate_score()` → returns `float`
- `foo_with_cot_reasons()` → calls `generate_score_and_reasons()` → returns `Tuple[float, Dict]`

`generate_score` itself should never return a tuple. Please fix the structured-JSON parsing path inside `generate_score` to strip the reason dict and return only the score, then remove `_extract_score` and the seven call-site changes.

The regression test concept is still valuable — keep it, but update it to verify that `generate_score` returns a bare float even when the underlying provider responds with structured JSON.

WatchTree-19:
done, went with the root-cause fix. generate_score now always returns a bare float (drops the reason dict on the structured-json path), so _extract_score and the seven call-site changes are gone, and the two inline tuple guards in evaluate_abstention / evaluate_answerability are removed too.

also un-xfailed test_generate_score_structured_json_returns_float (#2496), which now passes, and updated test_score_parsing_pipeline and test_openai_capabilities to expect a float. rebased onto main.

sfc-gh-jreini (approved):
The fix is exactly right. `generate_score` now returns a bare `float` on all paths — the structured JSON dict/list branches no longer produce a tuple, and the two `isinstance(score, tuple)` workarounds in `evaluate_abstention` / `evaluate_answerability` are correctly removed rather than left as dead code. The test updates are consistent: `test_generate_score_structured_json_returns_float` no longer needs `xfail`, and the integration / unit assertions drop the tuple unpacking and the stale `reason` assertions. Clean follow-through on the requested change. Thanks!

sfc-gh-jreini (approved):
The fix is exactly right. `generate_score` now returns a bare `float` on all paths — the structured JSON dict/list branches no longer produce a tuple, and the two `isinstance(score, tuple)` workarounds in `evaluate_abstention` / `evaluate_answerability` are correctly removed rather than left as dead code. The test updates are consistent: `test_generate_score_structured_json_returns_float` no longer needs `xfail`, and the integration / unit assertions drop the tuple unpacking and the stale `reason` assertions. Clean follow-through on the requested change. Thanks!

PR #2688, opened 2026-08-14
Add Ollama provider (trulens-providers-ollama)

### LINKED ISSUE #2405: Add Ollama Provider (Local Models)
## Summary
Ollama is already reachable through the LiteLLM provider, but a dedicated provider is still valuable because:

1. **Removes the `litellm` dependency** — Ollama users shouldn't need to install a 100+ provider routing layer just to talk to localhost
2. **Enables Ollama-specific features** like model pull/list, custom Modelfile options, and keep-alive control that LiteLLM doesn't expose
3. **Simplifies onboarding** — `from trulens.providers.ollama import Ollama` is more discoverable than configuring a LiteLLM model string
4. **Sensible defaults** (localhost:11434, no API key) instead of requiring users to understand LiteLLM's config format

## What
Create `src/providers/ollama/` with an `Ollama` class. Use the `ollama` Python package or the OpenAI-compatible endpoint.

## Reference
Follow the pattern in existing providers under `src/providers/`.

## Difficulty
Easy-Medium

connectsudhindra-gif:
## Summary

Closes #2405.

Adds a dedicated `Ollama` feedback provider (`trulens-providers-ollama`) so users can run feedback functions against a local (or remote) [Ollama](https://ollama.com/) server without pulling in the full LiteLLM routing layer or configuring an API key.

Unlike routing through LiteLLM or Ollama's OpenAI-compatible `/v1` API, this provider talks to Ollama's **native** `/api/chat` endpoint directly. That turned out to be necessary, not just nicer: Ollama's OpenAI-compatible surface silently ignores top-level `options`/`keep_alive` fields (verified live â€” requests through it had zero effect on `num_ctx` or model residency in `ollama ps`), so a provider that only wrapped the OpenAI-compatible endpoint couldn't actually deliver the Ollama-specific value the issue asks for.

- `Ollama` still reuses `OpenAI.__init__` for construction (request pacing, moderation endpoint, serialization), preconfigured with `base_url="http://localhost:11434/v1"` (overridable via `OLLAMA_BASE_URL`/`base_url`) and a placeholder API key (Ollama needs none, but the OpenAI client requires a non-empty string).
- `_create_chat_completion` posts directly to `{base}/api/chat`, translating `temperature`/`seed` into Ollama's nested `options`, and `response_format` into Ollama's native JSON-schema `format` parameter for real schema-constrained structured output.
- `options` (Modelfile-style parameters: `num_ctx`, `mirostat`, `repeat_penalty`, ...) and `keep_alive` are supported both at the provider level and per-call, verified to actually take effect via `ollama ps`.
- `list_models()` / `pull_model()` for model management, via Ollama's native `/api/tags` and `/api/pull`.
- Token usage from Ollama's native response (`prompt_eval_count`/`eval_count`) is fed into the existing cost-tracking callback, since bypassing the openai SDK client also bypasses its automatic usage instrumentation.
- Registered the new package in the root `pyproject.toml` providers group, `mkdocs.yml` API reference nav, `docs/reference/providers/index.md`, and `docs/contributing/optional.md`.

## Test plan

- [x] `tests/unit/providers/test_ollama_capabilities.py` (10 tests): default `base_url`/`api_key`/`model_engine`, `OLLAMA_BASE_URL` env var handling, native `/api/chat` request construction, `options`/`keep_alive` provider-level + per-call merging, `temperature`/`seed`/`reasoning_effort` translation, structured-output JSON-schema `format`, usage/cost recording, `list_models`, `pull_model` (success + failure).
- [x] Verified live against a real Ollama server (Docker `ollama/ollama` image, `llama3.2` model pulled):
  - Plain chat completion and `provider.relevance(...)` (1.0 relevant vs. 0.0 irrelevant) work end-to-end.
  - `options={"num_ctx": 8192}` and `keep_alive="9m"` confirmed applied via `ollama ps` (context size and residency changed accordingly) â€” and confirmed *not* applied when sent through the OpenAI-compatible endpoint instead, which is why this uses the native API.
  - Token usage correctly recorded on `provider.endpoint.global_callback.cost` after a native call.
  - `list_models()` / `pull_model()` against the live server.
- [x] Existing `test_openai_capabilities.py` suite still passes (no regression; this provider no longer shares the chat-completion code path with `OpenAI`, only construction).


docs/contributing/optional.md
```diff
@@ -28,6 +28,8 @@ Tools for instrumenting LLM apps built with other frameworks:
 Integrations for running feedback functions with different LLM providers:
 
 - `trulens-providers-openai` - OpenAI models
+- `trulens-providers-ollama` - Local (or remote) Ollama models via its
+  OpenAI-compatible API
```
joshreini1: Change 'OpenAI-compatible API' to 'native API'. The PR summary explicitly states the provider uses Ollama's native /api/chat endpoint because the OpenAI-compatible /v1 surface silently ignores options/keep_alive — this line contradicts that core premise and will mislead users about what the provider actually does.

src/providers/ollama/pyproject.toml
```diff
@@ -0,0 +1,40 @@
...
+  "Development Status :: 3 - Alpha",
+  "License :: OSI Approved :: MIT License",
+]
+
+[tool.poetry.dependencies]
+python = "^3.9"
+trulens-core = { version = "^2.0.0" }
+trulens-feedback = { version = "^2.0.0" }
+trulens-providers-openai = { version = "^2.0.0" }
+openai = ">=1.52.1,<2.0.0"
```
joshreini1, line 34: Add langchain-community (or langchain-core) as an explicit dependency. The provider imports LLMResult from langchain_core.outputs at provider.py:14, but this pyproject.toml doesn't declare it. The OpenAI provider (which this extends) declares langchain-community≥0.3.29 for the same import — follow that pattern here so the package boundary is clean and doesn't rely on undeclared transitive deps.

src/providers/ollama/trulens/providers/ollama/provider.py
```diff
@@ -0,0 +1,304 @@
...
+        else:
+            raise ValueError("`prompt` or `messages` must be specified.")
+
+        model_name = kwargs.pop("model", self.model_engine)
+
+        # Ollama exposes generation parameters (temperature, seed, top_p,
+        # ...) nested under `options`, not as top-level request fields like
+        # the OpenAI API.
+        effective_options: Dict[str, Any] = dict(self.options or {})
+        if options:
```
joshreini1, line 248: Document which OpenAI-style parameters are translated vs. ignored. Only temperature/seed/top_p move into Ollama's nested options dict; common params like max_tokens (Ollama calls it num_predict), frequency_penalty, presence_penalty are silently ignored. Add a docstring note or class-level comment listing supported OpenAI params and their Ollama equivalents (or state that only the three are translated and others should be passed via the options dict directly).

joshreini1 (changes requested):
Well-implemented provider with native Ollama API integration, proper options merging, usage tracking via LLMResult, and comprehensive unit tests. Three issues block merge: (1) docs/contributing/optional.md incorrectly states the provider uses the OpenAI-compatible API when the entire premise is using the native /api/chat endpoint instead, (2) provider.py imports LLMResult from langchain_core but pyproject.toml lacks an explicit langchain-core or langchain-community dependency (the OpenAI provider declares langchain-community≥0.3.29 for this same import), (3) parameter translation only covers temperature/seed/top_p but silently ignores other OpenAI-style params like max_tokens, worth documenting which params are supported. Verdict: Request changes — 1 blocker (incorrect docs), 2 should-fix (missing dependency, parameter doc gap).

docs/contributing/optional.md
connectsudhindra-gif: PR #2688 now shows the new commit on top.

Summary:
- Ran the unit tests relevant to this change: all 24 tests pass (test_ollama_capabilities.py — 10, test_openai_capabilities.py — 14). 
- Broader provider tests (anthropic/bedrock/cortex/google/huggingface/langchain/litellm) failed only because their packages weren't installed in this scratch test venv — unrelated to this change.
- Reviewed both commits before pushing: my new commit a90bbf2d7. The two pre-existing commits (9c383af69, 7a8d7f3f5), I left those untouched (rewriting would force-push over the exact commit the reviewer's approval is pinned to) and pushed only the new clean commit on top.
- Pushed a90bbf2d7 to connectsudhindra-gif/trulens:add-ollama-provider — PR #2688 now reflects the fix.

connectsudhindra-gif:
Fixed pls review and merge

connectsudhindra-gif:
@joshreini1 , pls merge, let me know if you need other adjustment thank you

joshreini1 (approved):
Well-implemented native Ollama provider with proper options merging, usage tracking via LLMResult, and comprehensive unit tests. Thanks for contributing!!

Issue #2442, opened 2026-05-04
Streaming LLM Response Support: Token Tracking, TTFT, Partial Output Capture


sfc-gh-jreini:
## Summary

Add first-class support for streaming LLM responses in TruLens instrumentation: track tokens as they arrive, measure time-to-first-token (TTFT), accumulate costs incrementally, and capture partial outputs for evaluation.

## Motivation

Most production LLM apps use streaming for UX reasons. Currently TruLens only captures the final aggregated response. Key metrics like TTFT and streaming throughput (tokens/sec) are invisible. Additionally, streaming responses may be evaluated mid-stream for guardrails.

## Proposed Approach

1. Add `GENERATION` span attributes for streaming:
   - `ai.observability.generation.is_streaming` (bool)
   - `ai.observability.generation.time_to_first_token_ms` (float)
   - `ai.observability.generation.tokens_per_second` (float)
   - `ai.observability.generation.chunks_received` (int)
2. Instrument common streaming patterns:
   - OpenAI `stream=True` (via provider auto-instrumentation)
   - LangChain `astream()` / `stream()`
   - Generic async generator pattern
3. Allow metrics to evaluate partial/in-progress outputs (for streaming guardrails)
4. Cost tracking should accumulate as chunks arrive (not just at end)

## Acceptance Criteria

- [ ] TTFT captured for streaming LLM calls
- [ ] Token throughput (tokens/sec) recorded as span attribute
- [ ] Cost updates incrementally during stream
- [ ] Works with OpenAI provider streaming
- [ ] Example notebook demonstrating streaming instrumentation
- [ ] Dashboard shows TTFT in latency views

## Related

- Streaming guardrails (examples/quickstart/blocking_guardrails.ipynb)

Abelo9996:
I'd like to take this on!

joshreini1:
Thanks @Abelo9996 !!

PR #2689, opened 2026-08-14
Add GENERATION span attributes for streaming (semconv)

connectsudhindra-gif:
## Summary

Part of #2442 (streaming LLM instrumentation). This is PR 1 of several small, independently-revertable PRs toward that issue, split so each piece can land (or be reverted) on its own.

Adds the four new `GENERATION` span attributes the issue asks for, as a pure additive semconv change with **no behavior change**:

- `ai.observability.generation.is_streaming`
- `ai.observability.generation.time_to_first_token_ms`
- `ai.observability.generation.tokens_per_second`
- `ai.observability.generation.chunks_received`

Nothing populates these yet — that's the next PR (wiring them up for the OpenAI provider's `stream=True` path, extending the existing chunk-handling in `OpenAIEndpoint._handle_response`).

## Scope of the overall issue

Issue #2442 bundles several genuinely separate efforts: semconv attributes (this PR), OpenAI streaming instrumentation, LangChain `stream()`/`astream()` instrumentation, a generic async-generator pattern, mid-stream/partial-output evaluation for guardrails, and a dashboard latency-view UI change. Splitting these into separate PRs keeps each one small, reviewable, and safe to revert independently rather than landing (or blocking on) one large, multi-part change.

## Test plan

- [x] Verified the new attribute keys resolve correctly, e.g. `SpanAttributes.GENERATION.IS_STREAMING == "ai.observability.generation.is_streaming"`.
- [x] No existing tests reference `trulens.otel.semconv.trace.SpanAttributes.GENERATION` (confirmed via repo-wide search), and this package isn't covered by the `tests/unit/static` API golden-snapshot tests, so nothing needed regenerating.

### DIFF
--- a/src/otel/semconv/trulens/otel/semconv/trace.py
+++ b/src/otel/semconv/trulens/otel/semconv/trace.py
@@ -47,6 +47,19 @@ class GenAIAttributes:
         TEMPERATURE = GEN_AI_SCOPE + ".request.temperature"
         """Sampling temperature requested."""
 
+        STREAM = GEN_AI_SCOPE + ".request.stream"
+        """Whether the request was made in streaming mode. Conditionally
+        required: only set if and only if the request is streaming (per
+        spec, an unset value is assumed to mean non-streaming)."""
+
+    class RESPONSE:
+        """Attributes describing the GenAI response."""
+
+        TIME_TO_FIRST_CHUNK = GEN_AI_SCOPE + ".response.time_to_first_chunk"
+        """Time to first chunk in a streaming response, in seconds, measured
+        from request issuance to the first chunk being received. Recommended
+        when the request was a streaming request."""
+
     class USAGE:
         """Token usage reported by the GenAI provider."""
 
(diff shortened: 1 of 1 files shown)


src/otel/semconv/trulens/otel/semconv/trace.py
```diff
@@ -425,6 +425,22 @@ class GENERATION:
 
         base = BASE_SCOPE + ".generation"
 
+        IS_STREAMING = base + ".is_streaming"
```
joshreini1: Replace IS_STREAMING with the official gen_ai.request.stream attribute (boolean) already defined in the OTel GenAI spec. The spec explicitly defines this as the standard way to flag streaming requests on spans and events. Duplicating it under the ai.observability.* namespace creates confusion and breaks interoperability with OTel-native tooling.

src/otel/semconv/trulens/otel/semconv/trace.py
```diff
@@ -425,6 +425,22 @@ class GENERATION:
...
+        """Whether this generation call used streaming (e.g. `stream=True`)."""
+
+        TIME_TO_FIRST_TOKEN_MS = base + ".time_to_first_token_ms"
```
joshreini1: Change TIME_TO_FIRST_TOKEN_MS to align with the official gen_ai.response.time_to_first_chunk attribute (double, unit=seconds, not milliseconds). The OTel GenAI spec defines this attribute for both span attributes and the gen_ai.client.inference.operation.details event. Using milliseconds breaks interoperability and makes this incompatible with the standard Histogram metric gen_ai.client.operation.time_to_first_chunk (which also uses seconds with explicit bucket boundaries).

src/otel/semconv/trulens/otel/semconv/trace.py
```diff
@@ -425,6 +425,22 @@ class GENERATION:
...
+        """Milliseconds between issuing the request and receiving the first
+        streamed chunk. Only set when `IS_STREAMING` is `True`."""
+
+        TOKENS_PER_SECOND = base + ".tokens_per_second"
```
joshreini1, line 456: Drop TOKENS_PER_SECOND and CHUNKS_RECEIVED — these are derived metrics that conflict with the OTel GenAI instrumentation model. The spec expects tokens/sec to be computed from gen_ai.client.token.usage + gen_ai.client.operation.duration, and inter-chunk latency to be reported as Histogram metrics (gen_ai.client.operation.time_per_output_chunk), not as span attributes. Span attributes should capture request/response metadata, not aggregated performance metrics. If TruLens needs these for internal evaluation, compute them from the official attributes rather than storing redundant custom ones.

src/otel/semconv/trulens/otel/semconv/trace.py
```diff
@@ -425,6 +425,22 @@ class GENERATION:
```
joshreini1, line 453: Should streaming metadata live on a gen_ai.client.inference.operation.details event (lines 101-105) rather than as span attributes? The OTel GenAI spec defines that event for capturing per-request metadata including gen_ai.response.time_to_first_chunk and gen_ai.request.stream. The GenAIEvents class already exists in this file (lines 95-114) and includes the event name constant. If these will be populated on the next PR's OpenAI instrumentation, clarify whether they'll be set as span attributes, event attributes, or both.

joshreini1 (changes requested):
The semconv additions don't align with the official OTel GenAI semantic conventions for streaming instrumentation. The spec defines gen_ai.request.stream (boolean) and gen_ai.response.time_to_first_chunk (double, seconds) as the canonical streaming attributes, and recommends Histogram metrics (not span attributes) for per-chunk latency and throughput. This PR adds custom ai.observability.generation.* attributes with wrong units (milliseconds vs seconds) and derived metrics (tokens_per_second, chunks_received) that conflict with the spec's instrumentation model. Request changes: 3 blockers.

src/otel/semconv/trulens/otel/semconv/trace.py
connectsudhindra-gif: Fixed

PR #2690, opened 2026-08-14
Instrument OpenAI stream=True with TTFT/chunks/throughput

connectsudhindra-gif:
## Summary

Part of #2442 (streaming LLM instrumentation), PR 2 of several. **Stacked on #2689** (adds the `GENERATION.IS_STREAMING` / `TIME_TO_FIRST_TOKEN_MS` / `TOKENS_PER_SECOND` / `CHUNKS_RECEIVED` span attributes with no behavior change) -- this PR is what actually populates them, for the OpenAI provider's sync `stream=True` path. The diff below includes #2689's commit until that one merges first; the second commit (`Instrument OpenAI stream=True ...`) is the new content here.

`OpenAICostComputer.handle_response` is the existing hook that already runs on every instrumented `openai.*.create()` return value (it's what currently populates `COST.*` attributes for both streaming and non-streaming calls). This PR extends it:

- **`IS_STREAMING` / `TIME_TO_FIRST_TOKEN_MS`**: known synchronously. The existing code already blocks on the first chunk to read its `model` for cost attribution, so timing that call gives TTFT for free.
- **`CHUNKS_RECEIVED` / `TOKENS_PER_SECOND`**: these need the *full* stream, which is only consumed later by the caller, after this attribute hook has already returned. To handle that, the stream's iterator is wrapped to count chunks and, once exhausted, write both attributes directly onto the span that was current when the request was issued (captured via `trace.get_current_span()` at hook time). `TOKENS_PER_SECOND` is only set when the caller requested `stream_options={"include_usage": True}` -- OpenAI doesn't send per-chunk token counts otherwise, and I didn't want to fabricate a token estimate from chunk count.

### A real limitation, not papered over

This only works when the caller consumes the stream *before* the enclosing instrumented span closes -- the common pattern (an instrumented method that iterates the stream to build its return value). If the raw stream is instead handed off elsewhere uncontained (e.g. returned directly for pass-through streaming to an HTTP response), the span has already ended by the time the stream finishes, and `CHUNKS_RECEIVED`/`TOKENS_PER_SECOND` are silently not recorded for that call (`IS_STREAMING`/TTFT still are, since those are captured synchronously). Verified by test (`test_stream_completion_after_span_ended_is_a_safe_noop`) that this degrades gracefully rather than raising.

### Explicitly out of scope here

- The `AsyncOpenAI` client path goes through a separate, ad-hoc instrumentation function in `experimental/otel_tracing/core/session.py` rather than `OpenAICostComputer.handle_response`, so async streaming isn't covered by this PR.
- LangChain `stream()`/`astream()`, the generic async-generator pattern, mid-stream/partial-output evaluation, the dashboard latency view, and the example notebook are each separate follow-ups tracked against #2442.

## Test plan

- [x] New `tests/unit/providers/test_openai_streaming_span_attributes.py` (5 tests), using a real `InMemorySpanExporter`/`TracerProvider` (not just mocks) to prove attributes actually land on the exported span:
  - `IS_STREAMING`/`TIME_TO_FIRST_TOKEN_MS` present immediately in the returned attributes dict.
  - `CHUNKS_RECEIVED`/`TOKENS_PER_SECOND` correctly recorded on the span once the stream is fully consumed within an open span.
  - `TOKENS_PER_SECOND` correctly omitted when no usage chunk is present.
  - Consuming the stream after the span has already ended is a safe no-op (no exception, attributes simply absent).
  - Non-streaming responses are unaffected (no streaming attributes added).
- [x] Existing `test_openai_capabilities.py` (14 tests) and `test_async_openai_capabilities.py` (10 tests) still pass -- no regression.


tests/unit/providers/test_openai_streaming_span_attributes.py
```diff
@@ -0,0 +1,147 @@
...
+        list(stream)  # fully consume while span is still open
+
+    spans = otel_setup.get_finished_spans()
+    assert len(spans) == 1
+    attrs: Dict[str, Any] = spans[0].attributes
+    assert attrs[SpanAttributes.GENERATION.CHUNKS_RECEIVED] == 3
+    assert attrs[SpanAttributes.GENERATION.TOKENS_PER_SECOND] > 0
+
+
+@pytest.mark.optional
```
joshreini1, line 98: Rename to `test_tokens_per_second_omitted_without_usage` - the test verifies that CHUNKS_RECEIVED is present (line 108 asserts `== 2`) and only TOKENS_PER_SECOND is absent, making the current name wrong and confusing for future maintainers.

src/providers/openai/trulens/providers/openai/endpoint.py
```diff
@@ -660,3 +673,67 @@ def new_iter():
...
+                    state["completion_tokens"] / duration_s,
+                )
+        except Exception:
+            logger.debug(
+                "Could not record streaming span attributes; the span may "
+                "have already ended.",
+                exc_info=True,
+            )
+
+    if isinstance(stream, openai.AsyncStream):
```
joshreini1: Remove the AsyncStream branch (lines 731-733) - async is explicitly out of scope per the PR description and routes through separate instrumentation in experimental/otel_tracing, making this untested dead code. Add it later when async is unified, with tests, following YAGNI.

src/providers/openai/trulens/providers/openai/endpoint.py
```diff
@@ -660,3 +673,67 @@ def new_iter():
 
     stream._iterator = new_iter()
     return stream
+
+
+def _instrument_stream_span_attributes(stream, first_chunk_received_at: float):
+    """Wrap `stream`'s iterator to record `CHUNKS_RECEIVED` and, when usage
+    is available (e.g. the caller passed ``stream_options={"include_usage":
+    True}``), `TOKENS_PER_SECOND` on the span that is current *right now* --
+    which must be called before this function, i.e. synchronously as soon as
```
joshreini1: Change 'which must be called before this function' to 'which must be captured before this function returns' - the current wording suggests calling order, but the actual constraint is that get_current_span() must execute while the caller's span is still current, preserving the span reference for later attribute writes.

joshreini1 (changes requested):
Solid implementation of streaming instrumentation - wrapping the stream iterator to record attributes is clean, and the approach correctly handles the limitation where spans must still be open when streams are consumed. One blocker: a test name directly contradicts its assertions. Also consider removing untested AsyncStream handling since async is explicitly out of scope. **Request changes: 1 blocker (misleading test name), 1 should-fix (untested async branch).**

tests/unit/providers/test_openai_streaming_span_attributes.py
connectsudhindra-gif: PR #2690 (openai-streaming-instrumentation) — Rebased onto #2689's new base; endpoint.py and its test now use GenAIAttributes.REQUEST.STREAM/RESPONSE.TIME_TO_FIRST_CHUNK (seconds, not ms) instead of the removed keys.

PR #2691, opened 2026-08-14
Demonstrate streaming span attributes in the streaming quickstart

connectsudhindra-gif:
## Summary

Part of #2442 (streaming LLM instrumentation), PR 3 of several. **Stacked on #2690** (populates the streaming span attributes) and #2689 (adds them) -- the diff below includes both until they merge first; the last commit (`Demonstrate streaming span attributes ...`) is the new content here.

Extends the existing `examples/quickstart/streaming_apps.ipynb` (rather than adding a new notebook, since it already demonstrates exactly the right pattern -- an `@instrument`-decorated method consuming an OpenAI `stream=True` completion) with a section that queries recorded events for the new `GENERATION` span attributes (`is_streaming`, TTFT, `tokens_per_second`, `chunks_received`) and prints them.

Also fixes a comment that PR #2690 made stale: `stream_options={"include_usage": True},  # not yet tracked by trulens` -- that's no longer true, so it now explains what the option is for instead.

## Test plan

- [x] Ran the notebook's actual code path end-to-end as a script against a local Ollama server (OpenAI-compatible endpoint, since I don't have an OpenAI API key handy) and confirmed all four attributes populate correctly:
  ```
  is_streaming: True
  time_to_first_token_ms: 2.44
  tokens_per_second: 22.3
  chunks_received: 4
  ```
- [x] Verified the notebook file is still valid JSON and the diff is otherwise minimal (no unrelated re-encoding of existing cells).


tests/unit/providers/test_openai_streaming_span_attributes.py
```diff
@@ -0,0 +1,158 @@
...
+        list(stream)  # fully consume while span is still open
+
+    spans = otel_setup.get_finished_spans()
+    assert len(spans) == 1
+    attrs: Dict[str, Any] = spans[0].attributes
+    assert attrs[SpanAttributes.GENERATION.CHUNKS_RECEIVED] == 3
+    assert attrs[SpanAttributes.GENERATION.TOKENS_PER_SECOND] > 0
+
+
+@pytest.mark.optional
```
joshreini1, line 118: Rename to `test_tokens_per_second_omitted_without_usage`. The test actually verifies chunks_received IS recorded (asserts ==2) but tokens_per_second is not—the current name contradicts the assertion.

src/otel/semconv/trulens/otel/semconv/trace.py
```diff
@@ -425,6 +425,22 @@ class GENERATION:
 
         base = BASE_SCOPE + ".generation"
 
+        IS_STREAMING = base + ".is_streaming"
+        """Whether this generation call used streaming (e.g. `stream=True`)."""
+
+        TIME_TO_FIRST_TOKEN_MS = base + ".time_to_first_token_ms"
+        """Milliseconds between issuing the request and receiving the first
```
joshreini1: Change docstring from "issuing the request" to "the streaming response being available" for accuracy. The implementation measures from when handle_response gets the stream object (after HTTP returns) to when next() yields the first chunk, not from initial request send.

src/providers/openai/trulens/providers/openai/endpoint.py
```diff
@@ -660,3 +673,67 @@ def new_iter():
...
+                span.set_attribute(
+                    SpanAttributes.GENERATION.TOKENS_PER_SECOND,
+                    state["completion_tokens"] / duration_s,
+                )
+        except Exception:
+            logger.debug(
+                "Could not record streaming span attributes; the span may "
+                "have already ended.",
+                exc_info=True,
+            )
```
joshreini1, line 740: Change `duration_s > 0` to `duration_s > 0.01` (or similar threshold). Prevents misleadingly high tokens/sec on very fast streams (e.g. single chunk completing in 0.0001s would show 100k tokens/sec). A 10ms minimum ensures the metric represents realistic throughput.

joshreini1 (changes requested):
Clean streaming instrumentation with comprehensive test coverage. Correctly captures TTFT, chunks, and throughput by wrapping the stream iterator. One blocker: test at line 108 is named `test_chunks_received_omitted_without_usage` but asserts chunks_received IS present—rename to `test_tokens_per_second_omitted_without_usage`. Two should-fix items: clarify TTFT docstring (measures stream-available-to-first-chunk, not request-issuance) and add minimum duration guard in tokens/sec calculation to avoid misleading metrics on very fast streams. Request changes: 1 blocker.

PR #2692, opened 2026-08-14
Instrument the generic streaming pattern (covers LangChain too)

connectsudhindra-gif:
## Summary

Part of #2442 (streaming LLM instrumentation). **Stacked on #2689** (adds the semconv attributes; that commit is included in the diff until it merges). Independent of #2690/#2691 -- this is the "generic async generator pattern" item from the issue, not OpenAI-specific.

Any `@instrument(span_type=SpanAttributes.SpanType.GENERATION)`-decorated function that returns a sync or async generator now gets `IS_STREAMING`, `TIME_TO_FIRST_TOKEN_MS`, and `CHUNKS_RECEIVED` recorded automatically -- useful for custom LLM integrations that aren't OpenAI or LangChain (e.g. a raw SSE call to some other provider's API).

### Why this didn't need the deferred-span-write trick from #2690

The OpenAI provider's raw `Stream` object is usually handed back to application code that consumes it *after* the instrumentation hook that fires on `create()` returning -- the span is often already closed by then, which is why #2690 has to capture the span and write to it later, best-effort.

Here it's simpler: the existing (pre-dating this change) generator handling in `convert_to_generator`/`async_generator_wrapper` already keeps the span open for the generator's *entire* lifetime and only finalizes it once exhausted (verified this holds even under partial consumption + GC, via `test_generation_generator_partial_consumption_records_partial_count`). So the new attributes are just computed inline in the same `finally` block that already runs after consumption -- no deferred-write mechanism needed.

### Bonus: this also turned out to cover LangChain's `stream()`/`astream()` for free

`trulens.apps.langchain.tru_chain` already registers `BaseLanguageModel`/`BaseChatModel`'s `stream`/`astream` with `span_type=GENERATION`, and that routes through this exact same generic mechanism -- LangChain's `stream`/`astream` are literally a sync generator function and an async generator function, no different from any other instrumented streaming function. I verified this live (real `TruChain` + `langchain_core`'s `GenericFakeChatModel`, not a stub) before adding `TestOtelTruChainStreaming` in `tests/unit/test_otel_tru_chain.py` to pin it down. **No LangChain-specific instrumentation code was needed or added** -- this closes that item from the issue's "Instrument common streaming patterns" list as a side effect of the generic design, not a separate implementation.

### Deliberately not included

No `TOKENS_PER_SECOND` here, unlike the OpenAI-specific work -- a generic chunk's contents aren't known to be token-shaped, so there's no honest way to compute a token rate without provider-specific knowledge of the response (see #2690 for how that's done when the provider tells you the real completion-token count).

## Test plan

- [x] 4 new tests in `tests/unit/test_otel_instrument.py`: sync generator, async generator, a non-GENERATION generator correctly getting *no* streaming attributes (confirms the gating), and partial consumption + GC recording only the partial chunk count.
- [x] 2 new tests in `tests/unit/test_otel_tru_chain.py` (`TestOtelTruChainStreaming`): real `TruChain` + `GenericFakeChatModel`, sync `.stream()` and async `.astream()`, confirming `IS_STREAMING`/`TIME_TO_FIRST_TOKEN_MS`/`CHUNKS_RECEIVED` on real exported spans.
- [x] Full existing suite (`tests/unit/test_otel_instrument.py`, 23 tests) passes -- no regression to the pre-existing generator/span-lifecycle behavior this builds on.

### DIFF
--- a/src/core/trulens/core/otel/instrument.py
+++ b/src/core/trulens/core/otel/instrument.py
@@ -6,6 +6,7 @@ from collections.abc import Sequence
 from contextvars import ContextVar
 import inspect
 import logging
+import time
 import types
 from types import TracebackType
 from typing import (
@@ -285,6 +286,40 @@ def _set_span_attributes(
         )
 
 
+def _record_generic_streaming_attributes(
+    span: Span,
+    span_type: SpanAttributes.SpanType,
+    chunk_count: int,
+    request_started_at: float,
+    first_chunk_at: float | None,
+) -> None:
+    """Record IS_STREAMING/TIME_TO_FIRST_TOKEN_MS/CHUNKS_RECEIVED for any
+    GENERATION-typed `@instrument`-decorated function that returns a
+    generator or async generator -- the "generic async generator pattern",
+    as opposed to provider-specific instrumentation (e.g. the OpenAI
+    endpoint's) which knows enough about chunk contents to also compute
+    TOKENS_PER_SECOND from real token counts. No token count is fabricated
+    here since a generic chunk's contents aren't known to be token-shaped.
+    """
+    if span_type != SpanAttributes.SpanType.GENERATION or chunk_count == 0:
+        return
+    try:
+        span.set_attribute(SpanAttributes.GENERATION.IS_STREAMING, True)
+        span.set_attribute(
+            SpanAttributes.GENERATION.CHUNKS_RECEIVED, chunk_count
+        )
+        if first_chunk_at is not None:
+            span.set_attribute(
+                SpanAttributes.GENERATION.TIME_TO_FIRST_TOKEN_MS,
+                (first_chunk_at - request_started_at) * 1000,
+            )
+    except Exception:
+        logger.debug(
+            "Could not record generic streaming span attributes.",
+            exc_info=True,
+        )
+
+
 def _finalize_span(
     span: Span,
     span_type: SpanAttributes.SpanType,
@@ -422,13 +457,21 @@ class instrument:
             ) as span:
                 ret = None
                 func_exception: Exception | None = None
+                was_generator = False
+                chunk_count = 0
+                request_started_at = time.monotonic()
+                first_chunk_at: float | None = None
                 # Run function.
                 try:
                     result = func(*args, **kwargs)
                     if isinstance(result, types.GeneratorType):
+                        was_generator = True
                         yield "is_generator"
                         ret = []
                         for curr in result:
+                            chunk_count += 1
+                            if first_chunk_at is None:
+                                first_chunk_at = time.monotonic()
                             ret.append(curr)
                             yield curr
                     else:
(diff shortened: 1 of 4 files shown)


tests/unit/test_otel_instrument.py
```diff
@@ -537,3 +537,102 @@ def call_llm(prompt: str) -> str:
...
+                pass
+
+        asyncio.run(consume())
+
+        spans = self.exporter.get_finished_spans()
+        self.assertEqual(len(spans), 1)
+        attrs = spans[0].attributes
+        self.assertTrue(attrs[SpanAttributes.GENERATION.IS_STREAMING])
+        self.assertEqual(attrs[SpanAttributes.GENERATION.CHUNKS_RECEIVED], 2)
+        self.assertGreaterEqual(
```
joshreini1, line 587: Add `self.assertNotIn(SpanAttributes.GENERATION.TOKENS_PER_SECOND, attrs)` after this assertGreaterEqual to match the sync test's verification—both tests should confirm generic streaming doesn't fabricate token-rate metrics, per the PR's stated design ("Deliberately not included: No TOKENS_PER_SECOND").

tests/unit/test_otel_instrument.py
```diff
@@ -537,3 +537,102 @@ def call_llm(prompt: str) -> str:
...
+        gen = stream_llm("hi")
+        for i, _ in enumerate(gen):
+            if i == 1:
+                break
+        del gen
+        gc.collect()
+
+        spans = self.exporter.get_finished_spans()
+        self.assertEqual(len(spans), 1)
+        self.assertEqual(
```
joshreini1, line 637: Add `self.assertTrue(spans[0].attributes[SpanAttributes.GENERATION.IS_STREAMING])` before this CHUNKS_RECEIVED check to verify partial consumption still marks the span as streaming—currently only tests the count, not the flag itself.

src/core/trulens/core/otel/instrument.py
```diff
@@ -285,6 +286,40 @@ def _set_span_attributes(
...
+) -> None:
+    """Record IS_STREAMING/TIME_TO_FIRST_TOKEN_MS/CHUNKS_RECEIVED for any
+    GENERATION-typed `@instrument`-decorated function that returns a
+    generator or async generator -- the "generic async generator pattern",
+    as opposed to provider-specific instrumentation (e.g. the OpenAI
+    endpoint's) which knows enough about chunk contents to also compute
+    TOKENS_PER_SECOND from real token counts. No token count is fabricated
+    here since a generic chunk's contents aren't known to be token-shaped.
+    """
+    if span_type != SpanAttributes.SpanType.GENERATION or chunk_count == 0:
```
joshreini1, line 304: The `chunk_count == 0` guard correctly prevents marking unconsume generators as streaming, but add a one-line comment explaining this—e.g., `# Don't mark as streaming if generator was created but never consumed`.

joshreini1 (changes requested):
Clean generic streaming instrumentation that correctly extends the existing generator-wrapping lifecycle. The timing logic is sound (monotonic clock, first-chunk capture), chunk counting handles partial consumption correctly, and the LangChain coverage is an elegant bonus. One blocker: the async streaming test at line ~588 is missing the `assertNotIn(TOKENS_PER_SECOND)` check that the sync test has—both should verify the deliberate absence of token-rate metrics for generic streams per the PR description. Request changes: 1 blocker, 1 should-fix.

PR #2693, opened 2026-08-14
Instrument AsyncOpenAI stream=True with TTFT/chunks/throughput

connectsudhindra-gif:
## Summary

Part of #2442 (streaming LLM instrumentation). **Stacked on #2690** (sync OpenAI streaming; that commit and #2689's are included in the diff until they merge). Covers a gap flagged when overall issue coverage was last reviewed: the sync-only OpenAI work didn't touch `AsyncOpenAI`, which goes through a separate instrumentation path.

### The bug this actually is

`AsyncOpenAI.post` is registered via `instrument_method(..., span_type=GENERATION)`, which defaults to `create_new_span=True`. That creates a span scoped exactly to the `await post(...)` call -- which always ends the instant `post()` resolves to the stream object, *before* the caller ever starts consuming it. This isn't a "sometimes it doesn't work" edge case like the sync provider's pass-through-streaming caveat -- it's structural: that span can never still be open by the time chunks are consumed, no matter what the calling code does.

Confirmed by testing against a live Ollama server before the fix: `IS_STREAMING` recorded correctly (it's known synchronously), but `TIME_TO_FIRST_TOKEN_MS`/`CHUNKS_RECEIVED`/`TOKENS_PER_SECOND` all came back `None` every time.

### The fix

- Added a `create_new_span` parameter to `instrument_method` (defaults to `True`, so every other caller is unaffected), and pass `create_new_span=False` for `AsyncOpenAI.post` so it piggybacks on the caller's already-open span instead -- the same pattern the sync client already uses successfully (`instrument_cost_computer`, also `create_new_span=False`).
- Added `instrument_async_openai_stream` in the OpenAI endpoint, which wraps the `AsyncStream`'s iterator and defers `TIME_TO_FIRST_TOKEN_MS`/`CHUNKS_RECEIVED`/`TOKENS_PER_SECOND` until the stream is actually consumed. It can't use the sync path's eager-first-chunk-read trick to measure TTFT (that needs `await`, and this fires from a plain sync `attributes` callback), so TTFT here is measured from "request resolved" to "caller pulled the first chunk" -- documented in the docstring as slightly less precise than the sync measurement if the caller does other work before starting to iterate.

Re-verified live after the fix: all four attributes populate correctly.

## Test plan

- [x] New `tests/unit/providers/test_async_openai_streaming_span_attributes.py`, using a **mocked httpx transport** (no network) that still exercises the *real* class-level `AsyncOpenAI.post` monkeypatch -- deliberately not an instance-level mock swap (`client.post = AsyncMock(...)`), since that would bypass the exact instrumentation this PR changes and wouldn't have caught the bug described above.
- [x] Verified live end-to-end against a real Ollama server (before and after the fix, to confirm the fix actually mattered).
- [x] Existing `test_otel_instrument.py` (23 tests), `test_tru_llama_workflow.py`, `test_openai_capabilities.py`, `test_async_openai_capabilities.py`, `test_openai_streaming_span_attributes.py` all still pass -- no regression from the `instrument_method` signature change or the `AsyncOpenAI.post` registration change.

### DIFF
--- a/src/providers/openai/trulens/providers/openai/endpoint.py
+++ b/src/providers/openai/trulens/providers/openai/endpoint.py
@@ -737,3 +737,83 @@ def _instrument_stream_span_attributes(stream, first_chunk_received_at: float):
             stream._iterator, wrap=_record_chunk, on_done=_on_done
         )
     return stream
+
+
+def instrument_async_openai_stream(
+    stream: "openai.AsyncStream", request_returned_at: float
+) -> "openai.AsyncStream":
+    """Async counterpart to `_instrument_stream_span_attributes`, for use
+    from a hook that -- unlike `OpenAICostComputer.handle_response` for the
+    sync client -- cannot `await` the first chunk itself to measure TTFT
+    synchronously (e.g. a plain sync `attributes` callback on an
+    already-awaited `AsyncOpenAI.post()` call). Records `IS_STREAMING`
+    immediately (returned as a caller-attached attribute, since that much is
+    known synchronously) and defers `TIME_TO_FIRST_TOKEN_MS` /
+    `CHUNKS_RECEIVED` / `TOKENS_PER_SECOND` to when the stream is actually
+    consumed, same graceful-no-op-if-span-already-ended caveat as above.
+
+    Note this measures TTFT from `request_returned_at` (when the coroutine
+    resolved to the stream object) to whenever the *caller* first pulls a
+    chunk -- which, unlike the sync path's eager first-chunk read, can
+    slightly overstate true network TTFT if the caller does other work
+    before starting to iterate the stream.
+    """
+    span = otel_trace.get_current_span()
+    if not span.get_span_context().is_valid or not isinstance(
+        stream, openai.AsyncStream
+    ):
+        return stream
+
+    state: Dict[str, Any] = {
+        "count": 0,
+        "first_chunk_at": None,
+        "last_chunk_at": request_returned_at,
+        "completion_tokens": None,
+    }
+
+    def _record_chunk(chunk):
+        now = time.monotonic()
+        state["count"] += 1
+        if state["first_chunk_at"] is None:
+            state["first_chunk_at"] = now
+            try:
+                span.set_attribute(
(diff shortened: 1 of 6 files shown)


src/providers/openai/trulens/providers/openai/endpoint.py
```diff
@@ -660,3 +673,147 @@ def new_iter():
 
     stream._iterator = new_iter()
     return stream
+
+
+def _instrument_stream_span_attributes(stream, first_chunk_received_at: float):
```
joshreini1, line 678: Add a type annotation for `stream` to match the style of `instrument_async_openai_stream` below: `stream: Union[openai.Stream, openai.AsyncStream]` (add `Union` to imports) or use the 3.10+ union syntax `openai.Stream | openai.AsyncStream` if the codebase supports it. Maintains signature consistency across instrumentation helpers.

src/providers/openai/trulens/providers/openai/endpoint.py
```diff
@@ -660,3 +673,147 @@ def new_iter():
...
+    count/throughput just won't be recorded for that call, same as if this
+    function did not exist.
+    """
+    span = otel_trace.get_current_span()
+    if not span.get_span_context().is_valid:
+        return stream
+
+    state: Dict[str, Any] = {
+        "count": 0,
+        "last_chunk_at": first_chunk_received_at,
```
joshreini1, line 700: Sync vs async throughput semantics differ: sync measures from when `handle_response` pulls the first chunk to when the caller receives the last (includes prepend overhead), while async measures from when the caller first iterates to last chunk received. Does this difference matter for downstream consumers comparing TOKENS_PER_SECOND across sync/async calls, or is the slight skew acceptable given the instrumentation constraints?

joshreini1 (commented):
Solid fix for a real async streaming bug: `AsyncOpenAI.post` now piggybacks on the caller's span via `create_new_span=False`, enabling TTFT/chunks/throughput capture when consumed. Well-documented, comprehensive mocked-transport tests, and the constraint-driven difference between sync (eager first-chunk TTFT) vs async (deferred measurement) is explained. Two minor style items and one question about throughput measurement semantics. Ready to merge after addressing the type annotation.

PR #2694, opened 2026-08-15
Fix streaming cost accumulation to use real token usage

connectsudhindra-gif:
## Summary

Part of #2442 (streaming LLM instrumentation): "Cost tracking should accumulate as chunks arrive (not just at end)". **Stacked on #2690** (sync OpenAI streaming; that commit and #2689's are included in the diff until they merge).

### The bug

`OpenAICallback.handle_generation_chunk` hard-codes `token_usage={}` on the `finish_reason == "stop"` chunk. Confirmed live against a real server: prompt/completion tokens were **always 0** for a streamed OpenAI call before this fix, even when the caller passed `stream_options={"include_usage": True}` -- **27/4 after** the fix, for the same request.

Root cause: OpenAI only sends real usage on a *separate* trailing chunk with empty `choices` (and only when `include_usage` is requested) -- sent right after, not on, the `"stop"` chunk. The existing code finalized the whole generation record the moment it saw `finish_reason == "stop"`, before that trailing chunk ever arrived, so the real numbers were never looked at. Non-streaming responses were unaffected; this was streaming-only.

### The fix

- `handle_generation_chunk` no longer finalizes on `"stop"` -- it just accumulates text, and stashes real usage from whichever chunk carries it (if any).
- A new `finalize_stream()` flushes whatever was accumulated into a real cost record -- real usage if it arrived, empty otherwise (same fallback as before, so behavior for callers who don't request `include_usage` is unchanged).
- The existing stream-iterator wrapping in `_handle_response` now calls `finalize_stream()` via `on_done`, once, when the stream is *actually* exhausted -- matching how a stream really ends, instead of guessing from a single chunk's shape.
- Also extended the per-span streaming-attributes helper from #2690 to write real `COST.NUM_PROMPT_TOKENS`/`NUM_COMPLETION_TOKENS`/`NUM_TOKENS` onto the span once known (previously stuck at whatever the synchronous, pre-consumption `COST.*` dict had -- i.e. zero -- for every streamed call), alongside the `CHUNKS_RECEIVED`/`TOKENS_PER_SECOND` it already recorded. Same graceful no-op if the span has already closed by the time the stream finishes.

## Test plan

- [x] `tests/unit/providers/test_openai_streaming_cost_accumulation.py` (4 new tests): real usage from a trailing chunk gets picked up correctly; the no-`include_usage` case still records the request (just with $0 cost, not silently dropped); `finalize_stream()` on nothing accumulated is a no-op; a full `OpenAICostComputer`-style end-to-end path produces real, non-zero cost for a known-priced model (`gpt-4o-mini`).
- [x] Fixed an under-specified `Mock` in the existing `test_openai_streaming_span_attributes.py` that this change's stricter usage handling exposed (a `Mock(completion_tokens=7)` with no other fields set was silently returning more mocks for `prompt_tokens`/`total_tokens` instead of the real ints a live OpenAI response always has) -- now asserts the new `COST.NUM_*` span attributes too.
- [x] Verified live against a real server end-to-end: prompt/completion tokens correctly non-zero after the fix, using a real streamed request with `stream_options={"include_usage": True}`.
- [x] Existing `test_otel_instrument.py`, `test_tru_llama_workflow.py`, `test_openai_capabilities.py`, `test_async_openai_capabilities.py` all still pass -- no regression to non-streaming cost tracking or anything else sharing this callback class.

### DIFF
--- a/src/providers/openai/trulens/providers/openai/endpoint.py
+++ b/src/providers/openai/trulens/providers/openai/endpoint.py
@@ -23,6 +23,7 @@ the involved classes will need to be adapted here. The important classes are:
 import inspect
 import logging
 import pprint
+import time
 from typing import (
     Any,
     Callable,
@@ -51,6 +52,7 @@ except ImportError:
         from langchain_core.outputs import LLMResult
 
 from langchain_community.callbacks.openai_info import OpenAICallbackHandler
+from opentelemetry import trace as otel_trace
 import pydantic
 from pydantic.v1 import BaseModel as v1BaseModel
 from trulens.core.feedback import endpoint as core_endpoint
@@ -125,12 +127,22 @@ class OpenAICostComputer:
         endpoint = OpenAIEndpoint()
         callback = OpenAICallback(endpoint=endpoint)
         model_name = ""
+        streaming_attributes: Dict[str, Any] = {}
 
         if hasattr(response, "__iter__") and not hasattr(response, "model"):
+            streaming_attributes[SpanAttributes.GENERATION.IS_STREAMING] = True
+            request_returned_at = time.monotonic()
             try:
                 first_chunk = next(response)
+                first_chunk_received_at = time.monotonic()
+                streaming_attributes[
+                    SpanAttributes.GENERATION.TIME_TO_FIRST_TOKEN_MS
+                ] = (first_chunk_received_at - request_returned_at) * 1000
                 model_name = first_chunk.model or ""
                 response = prepend_first_chunk(response, first_chunk)
+                response = _instrument_stream_span_attributes(
+                    response, first_chunk_received_at
+                )
             except Exception:
                 logger.exception(
                     "Exception occurred while consuming the first chunk from a streamed response."
@@ -156,6 +168,7 @@ class OpenAICostComputer:
         }
         if model_name:
             ret[SpanAttributes.COST.MODEL] = model_name
+        ret.update(streaming_attributes)
 
         return ret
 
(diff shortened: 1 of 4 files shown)


tests/unit/providers/test_openai_streaming_span_attributes.py
```diff
@@ -0,0 +1,155 @@
...
+    spans = otel_setup.get_finished_spans()
+    assert len(spans) == 1
+    attrs: Dict[str, Any] = spans[0].attributes
+    assert attrs[SpanAttributes.GENERATION.CHUNKS_RECEIVED] == 3
+    assert attrs[SpanAttributes.GENERATION.TOKENS_PER_SECOND] > 0
+    assert attrs[SpanAttributes.COST.NUM_PROMPT_TOKENS] == 12
+    assert attrs[SpanAttributes.COST.NUM_COMPLETION_TOKENS] == 7
+    assert attrs[SpanAttributes.COST.NUM_TOKENS] == 19
+
+
```
joshreini1, line 104: Rename test_chunks_received_omitted_without_usage to test_tokens_per_second_omitted_without_usage. The test asserts that CHUNKS_RECEIVED is present (equals 2), not omitted — only TOKENS_PER_SECOND is omitted. The current name is backwards and will confuse future readers.

src/providers/openai/trulens/providers/openai/endpoint.py
```diff
@@ -325,26 +343,52 @@ def handle_generation_chunk(self, response: Any) -> None:
...
-                    else:
-                        if hasattr(choice, "delta"):
-                            self.chunks.append({"text": choice.delta.content})
-
+                for choice in response.choices:
+                    if (
+                        choice.finish_reason != "stop"
+                        and hasattr(choice, "delta")
+                        and choice.delta.content
+                    ):
```
joshreini1, line 351: The check 'and choice.delta.content' skips falsy values (None, empty strings), which is safer than the old code that would have appended them. This is correct — just noting it's a subtle behavior improvement not called out in the PR description.

src/providers/openai/trulens/providers/openai/endpoint.py
```diff
@@ -660,3 +713,85 @@ def new_iter():
...
+                # span *before* any chunk was consumed, so it's stuck at
+                # zero/absent -- overwrite it now that real numbers exist.
+                prompt_tokens = getattr(usage, "prompt_tokens", 0) or 0
+                total_tokens = getattr(usage, "total_tokens", 0) or 0
+                span.set_attribute(
+                    SpanAttributes.COST.NUM_PROMPT_TOKENS, prompt_tokens
+                )
+                span.set_attribute(
+                    SpanAttributes.COST.NUM_COMPLETION_TOKENS,
+                    completion_tokens or 0,
```
joshreini1, line 779: The broad except Exception here is correct defensive programming — if span attribute recording fails (e.g., span already closed), the actual request shouldn't break. The debug-level logging is appropriate.

joshreini1 (commented):
Fix correctly addresses the root cause: OpenAI sends real token usage in a separate trailing chunk (when stream_options={"include_usage": True} is passed), not on the finish_reason=="stop" chunk. Deferring finalization until finalize_stream() is called after full stream exhaustion is the right approach. Tests are comprehensive and cover edge cases. One test name in test_openai_streaming_span_attributes.py is misleading and should be fixed. Otherwise ready to merge.

PR #2695, opened 2026-08-15
Add streaming (mid-stream) support to block_output guardrail

connectsudhindra-gif:
## Summary

Part of #2442: "Allow metrics to evaluate partial/in-progress outputs (for streaming guardrails)". Independent of the other streaming PRs -- based directly on `main`, no stacking.

`examples/quickstart/blocking_guardrails.ipynb` (referenced by the issue as "Related") has no streaming coverage today. `block_output` assumed `func()` returns a complete string synchronously -- calling it on a generator function would run the feedback function against the raw generator object itself, not any text, so streaming functions were effectively unusable with this guardrail before this change (not a behavior change for any previously-working case).

### What changed

`block_output` now detects a sync or async generator function and, every `check_every_n_chunks` chunks (default 20), re-evaluates the feedback against the *accumulated* output so far. If that trips the threshold, streaming stops -- yielding `return_value` once first, if one is configured -- otherwise chunks pass through unchanged. A stream shorter than one checkpoint still gets a single final check once it ends, purely so the `GUARDRAIL` span records a real evaluation, since nothing can be blocked after the stream has already fully passed through.

### Being upfront about what this can and can't do

This is a real safety constraint, not just an observability nicety, so the docstring says plainly: once a chunk is yielded it can't be un-sent. A violation is only ever caught starting from the *next* checkpoint onward, never retroactively -- there is no `check_every_n_chunks` value that guarantees zero unsafe output ever reaches the caller for a true token-by-token stream. Lower values catch problems earlier at the cost of more feedback calls during the stream (typically LLM-judge-backed, so real added latency/cost per checkpoint).

## Test plan

- [x] 6 new tests in `tests/unit/test_guardrails.py`: checkpoint cutoff with/without a fallback value, non-blocking stream passes everything through, a stream shorter than one checkpoint still gets a final (non-blocking, by then) check, and the async generator case.
- [x] Verified live end-to-end: `@instrument(span_type=GENERATION)` + `@block_output` stacked on a real streaming completion against a local Ollama server -- the exact pattern shown in the new streaming docstring example -- correctly cuts off after the configured checkpoint and yields the fallback (`['Once', ' upon', ' a', '[response withheld]']`).
- [x] Existing `test_guardrails.py`, `test_otel_guardrail.py`, `test_otel_guardrail_spans.py` (17 tests total) all still pass -- non-streaming `block_output`/`block_input`/`context_filter` behavior is unchanged.

### DIFF
--- a/src/core/trulens/core/guardrails/base.py
+++ b/src/core/trulens/core/guardrails/base.py
@@ -244,10 +244,36 @@ class block_input:
 class block_output:
     """Provides a decorator to block output based on a given feedback and threshold.
 
+    Also supports streaming (generator/async generator) functions: the
+    feedback is periodically re-evaluated against the *accumulated* output
+    so far as chunks arrive, every `check_every_n_chunks` chunks, so an
+    unsafe response can be cut off partway through rather than only being
+    caught (uselessly, since it would already be fully streamed to the
+    caller by then) after the whole thing has been generated.
+
+    !!! warning "Streaming guardrails can only block *future* chunks"
+        Once a chunk has been yielded it can't be un-sent -- if a violation
+        is detected at checkpoint N, chunks 1..N (whatever text they made
+        up) have already reached the caller. Lower `check_every_n_chunks`
+        catches problems earlier at the cost of more (typically
+        LLM-judge-backed) feedback calls during the stream; there's no
+        setting that guarantees zero unsafe output ever reaches the caller
+        for a true token-by-token stream. A final check also runs once the
+        stream ends (covering any tail shorter than `check_every_n_chunks`)
+        purely for the GUARDRAIL span's record -- by then nothing can be
+        blocked, since everything has already been yielded.
+
     Args:
         feedback: The feedback object to use for blocking. It must only take a single argument.
         threshold: The minimum feedback value required for a context to be included.
-        return_value: The value to return if the input is blocked. Defaults to None.
+        return_value: The value to return (or, for a streaming function,
+            yield once in place of further chunks) if the output is
+            blocked. Defaults to None.
+        check_every_n_chunks: For streaming (generator) functions only: how
+            often, in number of chunks yielded by the wrapped function, to
+            re-evaluate the feedback against the accumulated output so far.
+            Defaults to 20. Ignored for non-streaming functions, which are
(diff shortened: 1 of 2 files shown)


joshreini1 (commented):
Clean implementation of streaming support for block_output guardrails with excellent documentation. The logic correctly handles sync/async generators, accumulates chunks, and checks at configurable intervals. Tests comprehensively verify core behavior: early cutoff, tail checks, and pass-through cases. The docstring does an outstanding job explaining the fundamental limitation that already-yielded chunks can't be retracted. One validation gap: check_every_n_chunks isn't bounds-checked, so negative or zero values silently check every chunk (since since_last_check >= 0 is always True after incrementing), burning feedback budget unexpectedly. Verdict: 1 should-fix (parameter validation), 1 nit (type hint precision).

src/core/trulens/core/guardrails/base.py
```diff
@@ -277,47 +303,123 @@ def chat(self, question: str) -> str:
...
         self,
         feedback: core_metric.Metric,
         threshold: float,
         return_value: Optional[str] = None,
+        check_every_n_chunks: int = 20,
     ):
         self.feedback = feedback
         self.threshold = threshold
         self.return_value = return_value
+        self.check_every_n_chunks = check_every_n_chunks
```
joshreini1, line 338: Add validation for check_every_n_chunks after this line: `if check_every_n_chunks <= 0: raise ValueError("check_every_n_chunks must be > 0")`. Without this, negative or zero values silently check every single chunk (since `since_last_check >= 0` is always True after the first increment), which defeats the throttling purpose and burns feedback budget unexpectedly—typically LLM-judge calls that add real latency and cost per checkpoint.

src/core/trulens/core/guardrails/base.py
```diff
@@ -277,47 +303,123 @@ def chat(self, question: str) -> str:
...
+
+    def _evaluate(self, text) -> tuple:
```
joshreini1, line 340: Change return type hint from `tuple` to `tuple[float, bool]` for clarity—the method returns `(result: float, blocked: bool)` and the precision helps readers understand the unpacking in `_check`.

PR #2696, opened 2026-08-15
Show time-to-first-token in the Records page latency view

connectsudhindra-gif:
## Summary

Part of #2442: "Dashboard shows TTFT in latency views". This is the last of the code-level items in the issue. **Stacked on #2689** (adds the semconv attributes; that commit is included in the diff until it merges) -- this PR is purely a consumer of `SpanAttributes.GENERATION.IS_STREAMING`/`TIME_TO_FIRST_TOKEN_MS`, not a producer, so it needs those to exist but doesn't care which of the other streaming PRs populates them for a given record.

Adds a fourth metric card, **"Time to First Token (ms)"**, next to the existing Total tokens / Total cost / Latency cards on a record's detail view (Records page), whenever that record contains a streaming `GENERATION` span. Falls back to the original 3-column layout with no visual change for non-streaming records or when OTel tracing is off.

### Scope: per-record view only, not the Trends aggregate chart

The Trends page has an existing Average/P90/P99 latency chart backed by its own SQL/pandas time-bucketed aggregation pipeline. Extending *that* to include TTFT would mean writing new aggregation logic to pull a JSON span attribute (only present on streaming `GENERATION` spans) across potentially many events and bucket/percentile it the same way -- a materially larger, separate change. This PR covers the Records page per-record card only; the Trends chart is a reasonable follow-up, not bundled here.

### On visual verification

I don't have browser automation available in this environment, so I could not get a real browser screenshot of the rendered page -- flagging that explicitly rather than claiming full visual confirmation. What I *did* verify:

- Generated a real recorded run in a fresh sqlite DB and drove the actual `_render_record_metrics` function through Streamlit's own `AppTest` framework (headless, but exercises the real widget tree via the real rendering code, not a hand-rolled simulation), confirming the new metric card renders with the correct label and value, and that a non-streaming record still renders the original 3 cards cleanly with no error and no extra card.
- Along the way, this caught a real bug in my first draft: I'd initially based this branch on `main` directly (thinking it was independent), which meant `SpanAttributes.GENERATION.IS_STREAMING` didn't exist yet and the page crashed. Fixed by properly stacking on #2689 instead -- exactly the kind of thing a real browser screenshot wouldn't have caught any better than this did.

## Test plan

- [x] 4 new unit tests for `_get_record_ttft_ms` in `tests/unit/streamlit/test_streamlit_records.py`: found on a streaming span, OTel tracing disabled, no streaming span present, and the underlying query raising (graceful `None`, not a crash).
- [x] Live-verified via `AppTest` against a real sqlite DB + real recorded run, both with and without streaming attributes present.
- [x] Existing `tests/unit/streamlit/` (58 tests) and `tests/unit/test_dashboard_utils.py` suites pass -- no regression.


src/dashboard/trulens/dashboard/tabs/Records.py
```diff
@@ -174,7 +199,17 @@ def _render_record_metrics(
 
     app_specific_df = records_df[records_df["app_id"] == selected_row["app_id"]]
 
-    token_col, cost_col, latency_col, _ = st_columns([1, 1, 1, 3])
+    ttft_ms = _get_record_ttft_ms(selected_row)
+    if ttft_ms is not None:
+        token_col, cost_col, latency_col, ttft_col, _ = st_columns([
```
joshreini1, line 214: Collapse this to single-line format to match the existing 3-column case at line 214: `token_col, cost_col, latency_col, ttft_col, _ = st_columns([1, 1, 1, 1, 2])`. The multi-line style is more readable but creates formatting inconsistency within the same function.

joshreini1 (approved):
Clean conditional layout that adds TTFT when present and falls back gracefully for non-streaming records. Comprehensive tests cover all branches. One NIT: multi-line column list at line 204 is inconsistent with the single-line style at line 214, but not a blocker. Ready to merge.

src/dashboard/trulens/dashboard/tabs/Records.py
connectsudhindra-gif: @joshreini1 : I see its approved! You want me to fix this or it will me merged  ? 

(Collapse this to single-line format to match the existing 3-column case at line 214: token_col, cost_col, latency_col, ttft_col, _ = st_columns([1, 1, 1, 1, 2]). The multi-line style is more readable but creates formatting inconsistency within the same function.)

joshreini1: Please collapse then I will merge. Thanks!

connectsudhindra-gif: Done! I have to do this in 2 lines :)
Eval failed : ruff format (pinned v0.5.5, 80-char limit)

sfc-gh-jreini:
@connectsudhindra-gif can you resolve the merge conflict?

connectsudhindra-gif:
@sfc-gh-jreini : Pls review and merge.. Fixed merge conflicts..

#2696 (dashboard-ttft-latency-view) — the two conflicting files were trace.py and Records.py:
- trace.py: the GENERATION attribute conflict was cosmetic — both sides defined the same four attribute names with slightly different docstrings (main's had already absorbed this PR's earlier semconv commit via #2708). Kept main's wording.
- Records.py: this PR's _get_record_ttft_ms function duplicated _first_token_latency_ms, which already shipped on main via #2708 — and main's version is strictly better (reuses spans the caller already fetched instead of re-querying, and correctly picks the earliest streamed span instead of just the first one found). I dropped the redundant function and kept main's, then retargeted this PR's 4 unit tests onto _first_token_latency_ms (main had no tests for it, so that coverage is preserved, not lost) plus added one new test for the earliest-span-wins behavior.

connectsudhindra-gif:
@sfc-gh-jreini  pls merge

PR #2708, opened 2026-08-18
feat(otel): track streaming tokens, TTFT and throughput (#2442)

### LINKED ISSUE #2442: Streaming LLM Response Support: Token Tracking, TTFT, Partial Output Capture

Abelo9996:
Closes #2442

# Description

Adds first-class support for streaming LLM responses: time to first token, streaming throughput, chunk counts, and cost that stays current while a stream is being consumed.

This also fixes a defect found while implementing it. Cost tracking runs the moment the openai client hands back a `Stream`, which is as soon as the response headers arrive and before the model has generated anything, and nothing read the counts again once chunks started flowing. Every streamed call was therefore recorded with zero tokens and zero cost. The usage chunk that OpenAI sends for `stream_options={"include_usage": True}` was being dropped as well, because it arrives with an empty `choices` list and the chunk handler only looked inside `choices`. That is the reason for the `# not yet tracked by trulens` comment that had been sitting in `examples/quickstart/streaming_apps.ipynb`.

Four attributes are added to `SpanAttributes.GENERATION`, which until now held only its base scope:

- `ai.observability.generation.is_streaming`
- `ai.observability.generation.time_to_first_token_ms`
- `ai.observability.generation.tokens_per_second`
- `ai.observability.generation.chunks_received`

Streaming is measured both for OpenAI, where token counts are known, and for any instrumented generator, which covers LangChain `stream`/`astream` and plain async generators without framework-specific code. The record detail view in the dashboard now shows time to first token beside latency.

`StreamingSpanUpdater` in the openai endpoint captures the enclosing span up front and interposes itself on the stream's iterator, rewriting cost and streaming attributes as each chunk is consumed. This works because the enclosing span stays open for the length of the stream, and it means a stream abandoned part way through still records what it produced. The span is captured rather than looked up per chunk because chunks are consumed from inside the caller's own generator, where the current OTEL context is not guaranteed to still be the one the call was made under.

Time to first token is measured at the eager `next(response)` that cost tracking already performs in order to learn the model name. That call is what blocks on the model, so it is the real first-token wait rather than a proxy for it.

Where both the provider path and the generic generator path apply to the same span, the provider measurement wins. It sees chunks as they leave the network and knows how many tokens they carried, whereas counting yields would undercount an app that only yields content chunks. A provider reporting a *non*-streaming call does not suppress the generic path, since a generator wrapped around an ordinary call may still be streaming something of its own.

`tokens_per_second` is deliberately left unset rather than guessed when the completion token count is unknown, which is always the case for a bare generator and is the case for OpenAI unless the request opts into usage.

Point 3 of the proposed approach in the issue, evaluating partial or in-progress output for streaming guardrails, is not included here. It means running metrics against a buffer that is still being written, which is a change to the guardrails path rather than to instrumentation, and folding it in would make this diff hard to review. Happy to open a follow-up issue for it against `examples/quickstart/blocking_guardrails.ipynb`.

The semantic conventions reference is updated in this PR, so no separate documentation change is needed.

### Testing

`tests/unit/test_otel_streaming.py` adds 13 tests, none of which touch the network. They construct `openai.Stream` and `AsyncStream` objects directly, since the code under test reads only `_iterator`. Coverage includes sync and async streams, chunk counts, time to first token, throughput being absent when token counts are unknown, tokens and cost from the usage chunk, abandoned streams, and embedding responses not being labelled as generations.

The one worth calling out is `test_provider_measurements_survive_the_wrapping_generator`, which runs a stream through an instrumented generator that yields only the two content chunks and checks the span still reports all four chunks along with real throughput and token counts.

`tests/unit/test_otel_instrument.py`, `test_otel_span.py`, `test_otel_recording.py`, `test_exporter.py` and `tests/unit/providers/test_openai_capabilities.py` were also run with no regressions. ruff check and ruff format are clean at the version pinned in `.pre-commit-config.yaml`.

One gap worth stating: the LangChain `astream` path has not been exercised against a real LangChain app. It goes through the generic generator wrapper with no LangChain-specific code, but that part is untested end to end.

## Type of change

  not work as expected)
- [x] New Tests

### DIFF
--- a/docs/otel/semantic_conventions.md
+++ b/docs/otel/semantic_conventions.md
@@ -54,6 +54,10 @@
 | `ai.observability.mcp.output_is_error` | Whether the MCP tool call resulted in an error. | Never | | bool |
 | `ai.observability.mcp.execution_time_ms` | Time taken to execute the MCP tool call in milliseconds. | Never | | float |
 | `ai.observability.generation` | Namespace for attributes specific to a generation span. | | Y | |
+| `ai.observability.generation.is_streaming` | Whether the generation was streamed back incrementally. | Never | | bool |
+| `ai.observability.generation.time_to_first_token_ms` | Milliseconds spent waiting for the first chunk of a streamed generation. | Never | | float |
+| `ai.observability.generation.tokens_per_second` | Completion tokens generated per second, measured from the first chunk to the last so that it reflects generation throughput rather than including time to first token. Only set when the completion token count is known, which for a stream means the request asked for usage to be included. | Never | | float |
+| `ai.observability.generation.chunks_received` | Number of chunks received over a streamed generation. | Never | | int |
 | `ai.observability.graph_task` | Namespace for attributes specific to a graph task function execution span. | | Y | |
 | `ai.observability.graph_task.task_name` | Name of the task function. | Never | | str |
 | `ai.observability.graph_task.input_state` | Input state to the task. | Never | | Any |
--- a/examples/quickstart/streaming_apps.ipynb
+++ b/examples/quickstart/streaming_apps.ipynb
@@ -103,9 +103,10 @@
     "        completion = oai_client.chat.completions.create(\n",
     "            model=\"gpt-4.1-mini\",\n",
     "            stream=True,\n",
-    "            stream_options={\n",
-    "                \"include_usage\": True\n",
-    "            },  # not yet tracked by trulens\n",
+    "            # Asks OpenAI for a final usage chunk. Without it a streamed\n",
+    "            # response carries no token counts, so cost and tokens/sec cannot\n",
+    "            # be reported.\n",
+    "            stream_options={\"include_usage\": True},\n",
     "            temperature=0,\n",
     "            messages=[\n",
     "                {\n",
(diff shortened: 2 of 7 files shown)


joshreini1 (approved):
Excellent work. The streaming instrumentation is well-designed, the cost tracking fix addresses a real defect (zero tokens/cost for streamed calls), and 13 new tests provide comprehensive coverage. The StreamingSpanUpdater pattern is elegant: capturing the span up front and updating attributes as chunks flow correctly handles context-switching. The dashboard TTFT display alongside latency is a valuable UX improvement. Verdict: Ready to merge.

PR #2709, opened 2026-08-18
fix(feedback): ndcg_at_k shape crash and cohens_kappa label mutation

ClaireXi99:
## Summary

Two independent bugs in `src/feedback/trulens/feedback/groundtruth.py`:

**1. `ndcg_at_k` crashes when the golden set annotates fewer chunks than the retriever returns**

The old implementation passed `y_true=sorted(golden_scores, reverse=True)[:k]` (truncated to the golden count) and `y_score=rel_scores[:k]` (truncated to k) to sklearn's `ndcg_score`, which requires matching shapes. Whenever `len(golden) < len(retrieved)` and `k >= len(retrieved)`, this raises:

```
ValueError: array is not broadcastable to correct shape
```

This is the common case: golden sets typically annotate only the relevant chunks while retrievers return many. Reproduction:

```python
from trulens.feedback.dummy.provider import DummyProvider
from trulens.feedback.groundtruth import GroundTruthAgreement

gta = GroundTruthAgreement(
    ground_truth=[{"query": "q",
        "expected_chunks": [{"text": "chunk A", "expect_score": 1.0},
                             {"text": "chunk B", "expect_score": 0.8}]}],
    provider=DummyProvider(name="test", delay=0.0),
)
gta.ndcg_at_k("q", ["chunk A", "x1", "x2", "x3", "x4"])  # ValueError
```

Beyond the crash, the old construction was also semantically wrong in two ways: `ndcg_score` averages gains across tied `y_score` values (distorting the score when several retrieved chunks share the same zero relevance — e.g. it returns 0.92354 for the scenario above where the standard formula gives 0.66457), and pairing the ideal gains with the relevance scores by index misplaces gains whenever the retrieved order is not monotonic.

The fix computes NDCG@k directly from the standard formula — DCG over the top-k retrieved chunks, IDCG from the golden scores in descending order — and drops the `ndcg_score` import.

**Behavior change note:** for the narrow case where the top-k retrieved chunks are exactly the golden chunks (e.g. the reversed ideal order covered by the regression test) the result matches the previous sklearn value to floating-point precision; for most other previously working inputs the value changes as well, from the old incorrect value to the standard NDCG@k value.

**2. `cohens_kappa` permanently mutates `self.true_labels`**

Binarization wrote the result back to the stored field:

```python
if any(isinstance(label, float) for label in self.true_labels):
    self.true_labels = [1 if label >= threshold else 0 for label in self.true_labels]
```

With float labels (e.g. human-annotated scores), one `cohens_kappa` call silently corrupts every metric computed afterwards on the same aggregator — e.g. `mae` changes from 0.1675 to 0.3175 and `brier_score` from 0.0770 to 0.1325 for `true_labels=[0.0, 0.6, 0.7, 1.0]`, `scores=[0.55, 0.65, 0.68, 0.95]` — and a second call with a different `threshold` becomes a no-op (the labels are already ints, so the binarization branch is skipped). The fix binarizes into a local list.

## Testing

- 3 new regression tests; the 2 bug-locking tests fail on `main` and pass with this change, the third locks the numeric equivalence case:
  - `test_ndcg_at_k_fewer_golden_chunks_than_retrieved` (was ValueError)
  - `test_cohens_kappa_does_not_mutate_true_labels` (was silent corruption)
  - `test_ndcg_at_k_imperfect_ranking` (numeric lock)
- `tests/unit/test_groundtruth_agreement.py` + `tests/unit/test_groundtruth_aggregator.py` with `--run_optional_tests`: 75 passed; the single failure (`test_constructor_default_provider_warning`) reproduces on unmodified `main` (local environment issue, unrelated to this change)
- `ruff check` / `ruff format --check` clean on all touched files

## Checklist

- [x] Bug fixes verified against unmodified `main`
- [x] No other writers of `true_labels` or users of `ndcg_score` remain (grepped)
- [x] Existing behavior preserved where it was correct (empty retrieved list, unknown query, int labels)

### DIFF
--- a/src/feedback/trulens/feedback/groundtruth.py
+++ b/src/feedback/trulens/feedback/groundtruth.py
@@ -12,7 +12,6 @@ import pydantic
 from scipy import stats
 from sklearn.metrics import cohen_kappa_score
 from sklearn.metrics import matthews_corrcoef
-from sklearn.metrics import ndcg_score
 from sklearn.metrics import roc_auc_score
 from trulens.core.utils import imports as import_utils
 from trulens.core.utils import pyschema as pyschema_utils
@@ -40,6 +39,11 @@ logger = logging.getLogger(__name__)
 
 # Canonical implementation lives in trulens-core; re-export for back-compat.
 paired_permutation_pvalue = stats_utils.paired_permutation_pvalue
+
+
+def _dcg(scores: list[float]) -> float:
+    """Discounted cumulative gain: ``sum(score_i / log2(i + 2))`` (0-based)."""
+    return sum(score / np.log2(i + 2) for i, score in enumerate(scores))
 
 
 # TODEP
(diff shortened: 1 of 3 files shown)


sfc-gh-jreini (approved):
Thank you for identifying the problem, and producing a correct and well explained fix!

PR #2710, opened 2026-08-19
feat(feedback): additional_instructions and cot reasons for conversation metrics

### LINKED ISSUE #2699: [FEAT] `additional_instructions` and `_with_cot_reasons` Support for Conversation-Level Metrics
**Feature Description**
This request covers two related gaps in the conversation-level metrics introduced in TruLens 2.12 (`coherence_across_turns`, `conversation_helpfulness`, `topic_adherence`, `agent_goal_accuracy`):

**1. `additional_instructions` parameter**

All turn-level metrics (`groundedness_measure_with_cot_reasons`, `context_relevance_with_cot_reasons`) already support an `additional_instructions` parameter that appends domain-specific context to the judge's system prompt. Conversation-level metrics do not.

Currently, `coherence_across_turns` uses a hardcoded system prompt via `templates_conversation.CoherenceAcrossTurns.system_prompt` with no way to inject additional context:

```python
def coherence_across_turns(
    self,
    records: Union[List[Any], str],
    temperature: float = 0.0,
) -> float:
    ...
    return self.generate_score(
        system_prompt=templates_conversation.CoherenceAcrossTurns.system_prompt,
        user_prompt=f"Conversation Transcript:\n{transcript}",
        ...
    )
```

The requested signature would be:

```python
def coherence_across_turns(
    self,
    records: Union[List[Any], str],
    additional_instructions: str = "",
    temperature: float = 0.0,
) -> float:
```

This mirrors the existing pattern used by `groundedness_measure_with_cot_reasons` and `context_relevance_with_cot_reasons`.

**2. `_with_cot_reasons` variants**

All turn-level metrics have a `_with_cot_reasons` counterpart that returns `tuple[float, dict]` — the score plus the judge's chain-of-thought reasoning. Conversation-level metrics return only a float with no explanation.

The requested addition:

```python
def coherence_across_turns_with_cot_reasons(
    self,
    records: Union[List[Any], str],
    additional_instructions: str = "",
    temperature: float = 0.0,
) -> tuple[float, dict]:
    """
    Returns (score, {"reasons": "...judge reasoning..."}).
    """
```

The same pattern should apply to `conversation_helpfulness`, `topic_adherence`, and `agent_goal_accuracy` for consistency.

---

**Reason**
**`additional_instructions`**

Conversation-level metrics are disproportionately affected by domain mismatch because the judge evaluates an entire transcript at once. Without `additional_instructions`, the judge has no knowledge of domain-specific conventions — for example, in a text-to-SQL pipeline, the assistant's responses are structured `key: value` data rows rather than natural language. The default coherence prompt is calibrated for natural language conversations, so it may score structured SQL results as incoherent even when they are logically consistent.

**`_with_cot_reasons`**

Without reasoning output, a coherence score of `0.3` across a 5-turn conversation gives no actionable signal — it is impossible to tell which turn introduced the incoherence or why. Turn-level metrics solved this with `_with_cot_reasons`; conversation-level metrics need the same.

**Importance of Feature**
**`additional_instructions`** unlocks correct evaluation for any domain where assistant responses are not natural language — structured data outputs (SQL results, JSON, tabular rows), code generation, or any specialised vocabulary the default prompt does not account for. Without it, conversation-level scores are systematically biased against domain-specific apps, reducing trust in the evaluation framework for real-world production use cases.

**`_with_cot_reasons`** is essential for making conversation-level evaluation *actionable*. A score alone tells you that a multi-turn conversation failed quality thresholds; the reasoning tells you *where* and *why* — which turn introduced a contradiction, which follow-up question was mishandled, or which retrieval step broke the logical thread. This is the difference between evaluation as a pass/fail gate and evaluation as a diagnostic tool that guides iterative improvement of the pipeline.

Together, these two additions would bring conversation-level metrics to feature parity with turn-level metrics, which is the natural expectation given that both metric families share the same provider class and evaluation infrastructure.

dylanpulver:
# Description

Fixes #2699.

The four conversation-level metrics added in 2.12 (`coherence_across_turns`,
`conversation_helpfulness`, `topic_adherence`, `agent_goal_accuracy`) had no way
to pass domain context to the judge and returned a bare float. This adds an
`additional_instructions` parameter to all four, and adds a
`_with_cot_reasons` variant of each that returns `Tuple[float, Dict]`.

`additional_instructions` matters more at conversation scope than at turn scope
because the judge reads a whole transcript at once. When assistant turns are
structured output such as SQL result rows or JSON, the default prompts are
calibrated for prose and score correct output as incoherent. The reasoning
output matters because a single number over a five turn conversation does not
say which turn broke the thread.

Both additions follow the idiom already used by the turn-level metrics in the
same file. `additional_instructions` is threaded through
`_build_criteria_with_instructions`, which is how `coherence`, `harmfulness`,
`correctness` and `conciseness` already fold instructions into a template's
default system prompt. The chain of thought variants append
`templates_base.COT_REASONS_TEMPLATE` to the user prompt and call
`generate_score_and_reasons`, which is how `sentiment_with_cot_reasons` and the
agent trace metrics do it. Parameter order, type hints, docstring layout and
return types match the turn-level methods. The score ranges are unchanged, so
`agent_goal_accuracy_with_cot_reasons` stays binary while the other three stay
on the 0 to 3 Likert scale.

`additional_instructions` is placed after the existing domain parameters
(`reference_topics`, `reference_goal`) rather than immediately after `records`,
so existing positional calls keep working.

The turn-level metrics also carry a `**kwargs` shim that maps the deprecated
`custom_instructions` name onto `additional_instructions`. I left that off here
on purpose. These four methods never accepted `custom_instructions`, so a shim
would introduce a deprecated alias rather than preserve one, and `Metric`
already rewrites that name before it reaches the provider. Happy to add it if
you would rather have the signatures byte-identical.

`tests/unit/static/golden/api.trulens.3.11.yaml` is not regenerated in this PR.
That golden file already predates the 2.12 and 2.13 provider additions: it is
missing `coherence_across_turns`, `conversation_helpfulness`, `topic_adherence`,
`agent_goal_accuracy` and `citation_accuracy` on `main` today. Regenerating it
here would fold that unrelated drift into this change, so it seems better for a
maintainer to run `make write-api` separately.

The docs page `component_guides/instrumentation/conversation_evaluation.md` gains
two short subsections showing `additional_instructions` on a conversation metric
and the `_with_cot_reasons` variant.

### Test coverage

New tests in `tests/unit/test_feedback_criteria_and_additional_instructions.py`
reuse the existing `MockLLMProvider`, which captures the prompts the provider
would have sent. They assert that `additional_instructions` reaches the system
prompt for all four metrics, that the default system prompt is untouched when
the parameter is omitted, that `reference_topics` and `reference_goal` survive
alongside it, and that `Metric(additional_instructions=...)` plumbs through to a
conversation metric end to end. A second class covers the chain of thought
variants: tuple return shape, reasons dictionary, the reasons template landing
in the user prompt, the transcript landing in the user prompt, and a plain
transcript string being accepted in place of records.

New tests in `tests/unit/test_templates_conversation.py` follow the autospec
style already in that file and pin the score arguments the new variants pass to
`generate_score_and_reasons`, including the binary range for
`agent_goal_accuracy_with_cot_reasons`.

### What was verified and what was not

Verified on Python 3.11 with `trulens-core` and `trulens-feedback` installed
from this tree: `tests/unit/test_templates_conversation.py` and
`tests/unit/test_feedback_criteria_and_additional_instructions.py`, 32 passed.
The wider `tests/unit` run went from 488 passed to 501 passed with the same
single pre-existing failure in
`test_otel_async_concurrency.py::test_cancelled_finalization_failure_preserves_cancelled_error`,
which reproduces on an unmodified `main` in a full-directory run and passes in
isolation on both. `ruff check` and `ruff format --check` are clean on the three
changed Python files, run with ruff 0.5.5 to match the pinned pre-commit rev.

Not verified: `tests/unit/static/test_api.py`, the streamlit and dashboard
tests, `test_gepa_fitness.py`, `test_records_utils.py` and `test_reward.py`, all
of which need optional dependencies that are not installed here; any test
requiring provider API keys; and live judge behaviour against a real model. The
new prompt text has not been run through an actual LLM, so the wording of the
appended instructions has not been quality-checked end to end.

## Type of change

  not work as expected)
- [x] New Tests

### DIFF
--- a/docs/component_guides/instrumentation/conversation_evaluation.md
+++ b/docs/component_guides/instrumentation/conversation_evaluation.md
@@ -119,6 +119,42 @@ Selector.select_conversation_output()  # Ordered record outputs
 Conversation selectors cannot be mixed with record or span selectors in the same
 `Metric`. Attach separate metrics at each scope instead.
 
+### Add domain context to the judge
+
+Every conversation-level metric accepts `additional_instructions`, which appends
+domain guidance to the judge's system prompt. Reach for it when assistant turns
+are not natural language, such as SQL result rows, JSON payloads, or generated
+code, because the default prompts are calibrated for prose:
+
+```python
+f_conversation_coherence = Metric(
+    implementation=provider.coherence_across_turns,
+    name="Coherence Across Turns",
+    additional_instructions=(
+        "Assistant turns are structured SQL result rows rather than prose. "
+        "Judge coherence on whether each result follows from the user request "
+        "in the same turn and stays consistent with earlier results."
+    ),
+).on_conversation()
+```
+
+### Return the judge's reasoning
+
+Each conversation-level metric also has a `_with_cot_reasons` variant that
+returns the score together with the reasoning behind it, so a low score points at
+the turn that caused it instead of leaving the whole transcript under suspicion:
+
+```python
+f_conversation_coherence = Metric(
+    implementation=provider.coherence_across_turns_with_cot_reasons,
+    name="Coherence Across Turns",
+).on_conversation()
+```
+
+The conversation-level metrics are `coherence_across_turns`,
+`conversation_helpfulness`, `topic_adherence`, and `agent_goal_accuracy`. Each
+one has a matching `_with_cot_reasons` variant.
+
 ## Attach both metrics to an app
 
 Turn-level and conversation-level metrics can run on the same recorded application:
--- a/src/feedback/trulens/feedback/llm_provider.py
+++ b/src/feedback/trulens/feedback/llm_provider.py
@@ -4400,19 +4400,99 @@ class LLMProvider(core_provider.Provider):
     def conversation_helpfulness(
         self,
         records: Union[List[Any], str],
+        additional_instructions: Optional[str] = None,
         temperature: float = 0.0,
     ) -> float:
-        """Evaluate helpfulness across a multi-turn conversation."""
+        """
+        Uses chat completion model. A function that completes a template to
+        evaluate helpfulness across a multi-turn conversation.
+
+        Example:
+            ```python
+            from trulens.core import Metric
+            feedback = Metric(
+                implementation=provider.conversation_helpfulness,
+                name="Conversation Helpfulness",
+                additional_instructions=additional_instructions,
+            ).on_conversation()
+            ```
+
+        Args:
+            records (Union[List[Any], str]): The ordered conversation records, or a transcript string.
+            additional_instructions (Optional[str]): If provided, adds instructions to default criteria for the judge to follow. Defaults to None.
+            temperature (float): The temperature for the LLM response, which might have impact on the confidence level of the evaluation. Defaults to 0.0.
+
+        Returns:
+            float: A value between 0.0 (not helpful) and 1.0 (helpful).
+        """
         from trulens.feedback.templates import (
             conversation as templates_conversation,
         )
 
         transcript = templates_conversation.conversation_to_prompt(records)
+        system_prompt = self._build_criteria_with_instructions(
+            criteria=None,
+            default_criteria=templates_conversation.ConversationHelpfulness.system_prompt,
+            additional_instructions=additional_instructions,
+        )
         user_prompt = templates_conversation.ConversationHelpfulness.user_prompt_template.format(
             transcript=transcript
         )
         return self.generate_score(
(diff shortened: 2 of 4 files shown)


sfc-gh-jreini (approved):
This is a fantastic contribution that solves the issue well, is nicely congruent with the codebase and is well documented. Thank you!

PR #2716, opened 2026-08-20
docs: add trace debugging cookbook notebook (closes #2705)

### LINKED ISSUE #2705: Three-step notebook for debugging recurring trace failures
## Problem

A broad semantic-analysis cookbook is useful for learning the toolkit, but a developer debugging one bad trace needs a shorter path. They need to answer three questions quickly: has this happened before, what recurring failure mode does it belong to, and where does that failure diverge in the trace?

## Proposal

Add one deliberately small, credential-free notebook with exactly three main analysis sections. It uses an included synthetic JSONL fixture by default and shows one optional cell for substituting a DataFrame from `TruSession.get_records_and_feedback()` plus spans from `get_events()`.

The fixture should contain:

- At least 30 records across two app versions.
- Three recognizable failure groups using varied wording.
- Evaluation scores and explanations with known higher-is-better direction.
- Errors, model/tool metadata, ordered span paths, and span durations.
- At least one successful record near each failure group.

Keep configuration in one cell:

```python
FAILURE_METRIC = "Groundedness"
FAILURE_THRESHOLD = 0.5
QUERY_RECORD_ID = "record-retrieval-003"
N_CLUSTERS = 3
TOP_K = 5
```

### Step 1: Find similar failures

- Build a concise failure document from input, output, error, evaluation explanation, and ordered span-type path.
- Fit a lightweight local TF-IDF/SVD representation.
- Use the selected bad record or a short free-text description as the query.
- Display top matches with similarity, app version, failing metric, score, first error, model/tool, and record ID.
- Briefly explain the match using shared top terms or overlapping high-weight features.

This answers: **Has this happened before, and where?**

### Step 2: Find recurring failure modes

- Filter failed records using metric direction and threshold.
- Cluster normalized vectors with fixed-seed KMeans or MiniBatchKMeans.
- Produce one compact table with cluster size, app versions, medoid record ID, common errors, low metrics, tools, and top terms.
- Plot one two-dimensional PCA projection for navigation; PCA is not clustering input.
- Rank clusters by affected record count.

This answers: **What failure patterns are affecting us most?**

### Step 3: Inspect one failure group

- Select the largest cluster by default.
- Render two representative failures and the nearest successful record side by side.
- Show input, output, evaluation explanation, and ordered span path.
- Derive the first error span and flag spans above a simple within-trace latency threshold such as p90.
- Identify the first differing span type/name between a representative failure and nearby success when paths are comparable.
- End with a concise deterministic diagnostic summary, not an LLM-generated claim.

This answers: **Where in the trace does this failure diverge?**

The notebook should contain no generic loader framework, widgets, dashboard navigation, persistence, export workflow, Snowflake setup, optional embedding model, or Hotspots analysis.

## Non-goals

- A comprehensive semantic-analysis tutorial.
- Streamlit or built-in dashboard changes.
- Production vector indexing or search services.
- Automated root-cause claims.
- LLM-generated cluster names or diagnostics.
- Writing derived state back to TruLens.

## Acceptance criteria

- The notebook has exactly three main debugging sections.
- The default path executes offline with fixed seeds and no credentials.
- Similar-failure results recover same-group fixture examples within top-k.
- Every failed fixture record belongs to one displayed cluster.
- Cluster summaries include medoid, size, error, metric, tool, and top terms.
- Trace comparison highlights first error, slow spans, and first path divergence when one exists.
- A nearby-success comparison is present for the default cluster.
- Diagnostic text is deterministic and avoids claiming causal proof.
- No Streamlit or optional model dependency is required.

## Tests

- Synthetic fixture schema and known-group integrity.
- Failure-document and vectorization helpers.
- Top-k expectations for one query per failure group.
- Fixed-seed clustering and medoid selection.
- First-error, slow-span, and path-divergence helpers.
- Diagnostic-summary golden tests.
- Notebook structure check for exactly three main sections.
- Notebook output stripping and credential-free execution.

## Critical files

- `examples/expositional/use_cases/trace_debugging.ipynb`
- `examples/expositional/use_cases/trace_debugging.py`
- `examples/expositional/use_cases/data/trace_debugging_fixture.jsonl`
- `docs/cookbook/use_cases/trace_debugging.ipynb`
- `tests/unit/test_trace_debugging.py`

dylanpulver:
Closes #2705.

Adds `examples/expositional/use_cases/trace_debugging.ipynb` and the fixture it reads, `examples/expositional/use_cases/data/trace_debugging_fixture.jsonl`. `docs/cookbook` is a symlink to `examples/expositional`, so the `docs/cookbook/use_cases/trace_debugging.ipynb` entry in the issue's file list is served by the same file rather than a second copy. Configuration cell is the one pinned in the issue, unchanged.

`## Setup` plus the three `##` debugging sections from the issue. TF-IDF reduced with SVD supplies the representation; step 1 accepts either `QUERY_RECORD_ID` or free text, which is what you want when working from a bug report rather than a record id. Failed records are selected by applying the metric's own direction to `FAILURE_THRESHOLD`, then clustered with `KMeans(random_state=0, n_init=10)` on l2-normalized vectors; the PCA scatter is navigation only and never feeds the clustering. The root span is excluded from the p90 latency check, since it wraps the whole trace and would win every time. Path divergence is compared on `span_type/name` so a replaced step and a missing step both surface. The closing summary is assembled from the computed numbers and says the signals co-occur rather than naming a cause.

Fixture: 34 records over two app versions, 24 failing on `Groundedness`, seeding three failure modes (empty retrieval above the score cutoff, a pricing tool that errors or returns an unreadable payload, generation hitting the token ceiling), each with passing records near it. Wording varies inside a mode so matching is not a keyword lookup, and one metric is `lower_is_better` so the direction handling does real work.

## What I ran

Executed end to end with `jupyter nbconvert --to notebook --execute` from a fresh virtualenv holding only pandas, scikit-learn, matplotlib, nbconvert and ipykernel. No trulens, no provider SDK, launched under `env -i` so no API key or Snowflake variable was present. All 14 code cells completed clean, twice, with identical text output apart from the first-run matplotlib font-cache notice.

Each of the 24 failed records lands in exactly one of the 3 clusters, and each cluster maps onto exactly one seeded failure mode with nothing crossing over. The top 5 matches for `record-retrieval-003` all come from its own failure mode. `pre-commit run --files` passes every hook including nb-clean, so outputs and execution counts are stripped. Nothing here touches importable code, so no test target applies.

## What I left out

The issue's file list also names `trace_debugging.py` and `tests/unit/test_trace_debugging.py`. Recent cookbook notebooks merge as the notebook plus its data file, and moving the helpers into a sibling module would make a self-contained notebook depend on an import that does not travel with it. The honest cost is that the helpers are exercised only by running the notebook, not by CI. Say the word and I will add both to this branch.

No Colab badge, because the default path reads a fixture from disk and the badge would open a notebook that cannot find it. `llm_judge_benchmark.ipynb` omits it for the same reason.


examples/expositional/use_cases/trace_debugging.ipynb
```diff
@@ -0,0 +1,673 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "8395334c",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def slow_spans(spans):\n",
+    "    \"\"\"Spans at or above the within-trace p90, ignoring the root span.\"\"\"\n",
+    "    inner = [span for span in spans if span[\"span_type\"] != \"RECORD_ROOT\"]\n",
```
joshreini1, line 486: Clarify in the docstring that p90 is computed within-trace. With small traces (2-3 inner spans), p90 often equals the max duration, so only the slowest span survives — this is probably the intended behavior, but 'at or above the within-trace p90' would set clearer expectations.

examples/expositional/use_cases/trace_debugging.ipynb
```diff
@@ -0,0 +1,673 @@
...
+    "\n",
+    "vectorizer = TfidfVectorizer(stop_words=\"english\", sublinear_tf=True)\n",
+    "tfidf = vectorizer.fit_transform(df[\"failure_document\"])\n",
+    "\n",
+    "n_components = min(20, tfidf.shape[1] - 1, len(df) - 1)\n",
+    "svd = TruncatedSVD(n_components=n_components, random_state=0)\n",
+    "vectors = normalize(svd.fit_transform(tfidf))\n",
+    "\n",
+    "terms = np.array(vectorizer.get_feature_names_out())\n",
+    "failed_positions = np.flatnonzero(df[\"failed\"].to_numpy())\n",
```
joshreini1, line 236: When similar_failures() receives a record_id, it skips that record from results (line 236). When given free text, skip is empty (line 239). Should free-text queries also skip exact failure_document matches, or is collision impossible because free text is user-authored? Clarify intent or document the asymmetry.

joshreini1 (approved):
Well-structured self-contained debugging notebook with thoughtful fixture design. The clustering approach (spherical k-means on l2-normalized TF-IDF vectors) is sound, and the three-step workflow (similarity search, clustering, divergence detection) addresses the stated goal cleanly. Code is correct and reproducible. One documentation clarification recommended for p90 behavior on small traces, and a usability question about free-text query collision handling, but neither blocks merge. VERDICT: Ready to merge.

dylanpulver:
Thanks for the review.

On p90: the docstring reads "Spans at or above the within-trace p90, ignoring the root span", so the phrasing you suggested is the one that shipped. Your read of the consequence is right and worth stating. Every record in the fixture has 2 or 3 inner spans, and in all 34 of them exactly one span clears the cutoff, so `slow_spans` returns the single slowest span every time. Happy to add a sentence saying so before merge.

On the skip asymmetry: `skip` removes the identity match, not near duplicates. A record id resolves to a row of the frame, and that row scores 1.0 against itself, so it would take the top slot and tell you nothing. Free text is vectorized by `vectorizer.transform([query])` and is not a row, so there is no self match to remove. A free text query that happens to read like some record's `failure_document` should rank that record first, which is the answer the user asked for rather than a collision to suppress. I can put that in the docstring.

PR #2717, opened 2026-08-21
feat: connector-backed prompt management with version evaluation (closes #2703)

### LINKED ISSUE #2703: Connector-backed prompt management with version evaluation
## Problem

Prompts are currently code-owned strings or message arrays. TruLens can capture rendered messages in generation span events when content capture is enabled, but it cannot identify, version, resolve, roll back, diff, trace, or systematically evaluate a reusable prompt definition. App versions are too coarse because one app can use several prompts and prompt changes should not require an application redeploy.

## Proposal

Add one complete prompt-management vertical slice backed by the configured TruLens database. Rendering, model invocation, and evaluation remain in the user's process.

### Public models

- `Prompt`: stable prompt ID and slug, name, immutable prompt type, description, tags, and timestamps.
- `PromptVersion`: immutable version ID, prompt ID, parent version ID, text or chat content, variables, provider-neutral model defaults, response-format metadata, change note, content hash, and creation metadata.
- `PromptLabel`: mutable `(prompt_id, label)` pointer to one exact version.
- `PromptLabelHistory`: append-only movement with previous/new version, caller label, and timestamp.
- `ResolvedPrompt`: exact version plus strict local render/build methods.

Prompt types:

- `text`: one template string.
- `chat`: ordered `system`, `user`, `assistant`, or `tool` messages with text or structured content blocks.

Prompt type is fixed after creation. Version one supports simple named variables such as `{{question}}`; do not execute Python, Jinja filters, loops, conditions, or imports. Variables are declared explicitly, and rendering rejects missing or unexpected values in strict mode.

### SDK direction

```python
prompt = session.create_prompt(
    slug="support-assistant",
    name="Support assistant",
    prompt_type="chat",
)

version = session.create_prompt_version(
    prompt=prompt,
    messages=[
        {"role": "system", "content": "Answer using the support policy."},
        {"role": "user", "content": "{{question}}"},
    ],
    variables=["question"],
    model_defaults={"temperature": 0.2},
    change_note="Initial support prompt",
)

session.set_prompt_label(prompt, "production", version)
resolved = session.get_prompt("support-assistant", label="production")
request = resolved.render(question="How do I reset MFA?")
```

### Version and label behavior

- Version IDs are content-addressed from canonical content, variables, defaults, response format, and prompt identity.
- Versions are immutable and exact lookup is always available.
- Creating a version atomically moves `latest`.
- `staging` and `production` are conventional labels, not hosted environments.
- Moving a label writes immutable history.
- Rollback moves a label to an older version; it does not mutate a version.
- Label lookup resolves from the database. An optional process-local TTL cache may reduce reads, but correctness never depends on it.

### Rendering

- Return text or provider-neutral ordered message dictionaries.
- Validate variables before interpolation.
- Merge persisted model defaults with caller overrides without mutation.
- Return response-format metadata separately for provider adapters.
- Rendering performs no model call and requires no credentials.

### OTEL lineage

Add semantic attributes for prompt ID, slug, exact version ID, requested label, and rendered-content hash. Provide a helper/context manager that attaches them to a caller-owned generation span. Lineage must work when full GenAI content capture is disabled; do not copy complete prompt bodies into every span.

### Management UI

- Add a Prompts page listing prompts, current labels, and version history.
- Show structured text/message diffs.
- Create versions, move labels, and roll back labels.
- Render local previews from variable values.
- Never store model credentials or invoke a model from the page.

### Prompt version evaluation notebook

Ship a notebook as part of this feature so prompt versions can be tested, not merely stored. It uses the prompt-management APIs above, existing `Run`, current TruLens metrics, and `Run.compare`.

1. Create or load one prompt with two or three exact versions.
2. Load a checked-in synthetic dataset; show how to substitute a DataFrame or table source.
3. Resolve exact versions, not mutable labels, for reproducibility.
4. Render each version over identical ordered inputs.
5. Use a deterministic local fake callable by default and show how to replace it with a caller-owned provider.
6. Execute one normal `Run` per exact prompt version.
7. Attach prompt identity/version/hash to run metadata and generation spans.
8. Compute existing quality metrics and compare versions with `Run.compare`.
9. Present mean scores/confidence intervals, latency, observed cost, item-level deltas/regressions, and a compact Plotly chart.

The notebook must use persisted `PromptVersion` objects and may not reimplement a separate inline prompt representation.

## Non-goals

- Hosted prompt invocation or browser playground.
- A/B traffic routing, approvals, ownership, or RBAC.
- Webhooks, deployment automation, or remote cache invalidation.
- Arbitrary executable templating.
- Public prompt hubs, tool definitions, or MCP server management.
- Automatic prompt discovery from captured strings.
- Automatic optimization or prompt generation.
- A new experiment abstraction beyond `Run`.

## Acceptance criteria

- Text and chat versions serialize canonically and are immutable.
- Identical version creation is idempotent.
- Strict rendering detects missing and unexpected variables.
- Exact-version lookup remains stable after labels move.
- Concurrent label moves preserve one current pointer and append history.
- Rollback requires no application-code change beyond continued label lookup.
- Prompt lineage appears on generation spans without enabling prompt-content capture.
- No credentials or rendered user values are stored in prompt definitions.
- The complete workflow functions with SQLite and no hosted service.
- Dashboard actions use connector APIs and never invoke a model.
- The notebook evaluates at least two exact persisted versions over identical inputs without network access by default.
- `Run.compare` reports per-metric and per-item differences.
- Quality, latency, and available provider cost are reported without estimates; absent cost remains unknown rather than zero.

## Tests

- Text/chat validation and canonical content hashes.
- Variable parsing, strict rendering, and override merging.
- Version immutability and idempotency.
- Label creation, concurrent movement, history, and rollback.
- Cache expiration and explicit bypass.
- OTEL lineage with content capture disabled and enabled.
- SQLite/PostgreSQL persistence and migrations.
- Dashboard management and preview flows.
- Deterministic fake callable and identical-input checks.
- Run comparison fixture with a known regression and improvement.
- Notebook output stripping and optional docs-notebook execution.

## Critical files

- `src/core/trulens/core/schema/prompt.py`
- `src/core/trulens/core/prompt.py`
- `src/core/trulens/core/otel/instrument.py`
- `src/dashboard/trulens/dashboard/tabs/Prompts.py`
- `examples/expositional/use_cases/prompt_management.ipynb`
- `docs/cookbook/use_cases/prompt_management.ipynb`

dylanpulver:
# Description

The prompt management slice from #2703: content-addressed immutable versions, labels with history and rollback, strict local rendering, alembic revision 13, OTEL lineage, a Prompts dashboard page, and a version evaluation notebook.

Closes #2703.

The new `DB` methods are concrete and raise `NotImplementedError` rather than being abstract, so `SnowflakeEventTableDB` and out-of-tree `DB` subclasses keep working without stubs.

A first-time concurrent label move races: both writers see no row and both insert, and `SELECT ... FOR UPDATE` cannot lock a row that does not exist. A composite `(prompt_id, label)` primary key plus a retry on `IntegrityError` preserves one pointer. See `test_concurrent_label_moves_keep_one_pointer`.

Run metadata takes only a description and a label, so the notebook writes the prompt id, slug, version id, and content hash into the description as JSON rather than extending `Run`.

`api.trulens.3.11.yaml` is not regenerated. It reports 603 mismatches against `main` here before this branch, so rewriting it would bury the 42 entries added.

Ran ruff 0.5.5 (pinned in `.pre-commit-config.yaml`), `nb-clean check`, `pytest tests/unit` and `tests/integration/test_database.py` with failure sets identical before and after, and the notebook end to end under `env -i` with non-loopback sockets blocked.

## Type of change

- [x] New Tests


examples/expositional/use_cases/prompt_management.ipynb
```diff
@@ -0,0 +1,694 @@
+{
```
sfc-gh-jreini, line 1: <p>remove this - set by default</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/truera/trulens/pull/2717/discussion/#comment-3830594785'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

sfc-gh-jreini, line 1: <pre class="ql-syntax" spellcheck="false">Line #10.        if model_args.get("temperature", 0.0) != 0.0:
</pre><p>handling arbitrary model args like temperature here is unneccessary, you can remove this conditional</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/truera/trulens/pull/2717/discussion/#comment-3830594792'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

sfc-gh-jreini, line 1: <p>it would be valuable to spend a bit more time saying how cool it is to automatically capture the prompt lineage while not having to instrument manually here.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/truera/trulens/pull/2717/discussion/#comment-3830594800'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

sfc-gh-jreini, line 1: <pre class="ql-syntax" spellcheck="false">Line #4.    def records_when_ready(run, timeout: float = 180.0):
</pre><p>use the builtin instead of creating this new function recrods_when_ready: <a href="https://www.trulens.org/reference/trulens/apps/app?h=wait_for_feedback_results#trulens.apps.app.TruApp.retrieve_feedback_results" target="_blank">https://www.trulens.org/reference/trulens/apps/app/?h=wait_for_feedback_results#trulens.apps.app.TruApp.retrieve_feedback_results</a></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/truera/trulens/pull/2717/discussion/#comment-3830594809'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='22'/>

sfc-gh-jreini, line 1: <p>cost is 0 bc all of your generation and metrics are deterministic python calls instead of llm calls.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/truera/trulens/pull/2717/discussion/#comment-3830594817'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

sfc-gh-jreini, line 1: <pre class="ql-syntax" spellcheck="false">Line #1.    summary = pd.DataFrame([
</pre><p>just use get_leaderboard instead of building summary df yourself</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/truera/trulens/pull/2717/discussion/#comment-3830594826'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='24'/>

dylanpulver: Removed. `is_otel_tracing_enabled()` returns True unless `TRULENS_OTEL_TRACING` is set to "0" or "false" (`src/core/trulens/core/otel/utils.py:19`), so the cell was writing the default.

dylanpulver: Removed. `local_support_model` takes `model_args` and does nothing with it now, which is the forwarding shape the markdown above describes for a real provider.

dylanpulver: Expanded. The section now reads:

  The app holds one resolved version and renders it per input. The `prompt_lineage` context manager writes the prompt id, slug, exact version id, requested label, and a hash of the rendered content onto the generation span, so nothing in the app has to carry those identifiers around or pass them to the run. Six months from now, when the `production` label has moved twice and the version that produced a given trace is no longer the one it points at, the span still names the version it used.

  What lands on the span is identifiers and a hash, not the prompt body, so lineage survives with GenAI content capture off, which is the default.

dylanpulver: Replaced. The cell pulls record ids off `run.get_records()` and calls `tru_apps[name].retrieve_feedback_results(record_ids=...)`, which does the flush, the `compute_now`, and the wait. `add_run` hands back the run and not the app, so the loop above keeps the `TruApp` in `tru_apps` to make that call reachable.

dylanpulver: You are right and the paragraph was wrong. It claimed the column stays unknown and that a zero would misread. `get_leaderboard` reports `total_cost` as 0.0 for all three versions, and 0.0 is the true number: the generation and both metrics are deterministic Python calls with no model behind them. Rewritten to say that.

dylanpulver: Done:

  summary = (
      session.get_leaderboard()
      .loc["Support assistant"]
      .loc[list(candidates)]
      .join(
          pd.Series(
              {name: resolved.version_id for name, resolved in candidates.items()},
              name="version_id",
          )
      )
  )

It carries the metric means plus `latency` and `total_cost`, so the version id is the only column the hand-built frame had that this does not, and it comes off `candidates` in the join. `.loc[list(candidates)]` puts v1/v2/v3 back in order, since the leaderboard sorts by score. The plotly cell reads `summary.index` now.

sfc-gh-jreini (approved):
This is a great value-add and a common feature request. Thank you for contributing!

Issue #2718, opened 2026-08-21
[BUG] ruff pinned to three different versions; documented make lint/make format rewrite 380 files


dylanpulver:
**Bug Description**

The ruff a contributor runs is not the ruff the commit hook enforces. Three pins disagree:

- `.pre-commit-config.yaml` L22: `rev: v0.5.5` (this is the gate)
- `poetry.lock` L8838: `ruff 0.16.0`
- `pyproject.toml` L49: `ruff = ">=0.5"`, resolving to 0.16.3 in my venv

`main` is formatted for 0.5.5. Under 0.16.x, `ruff check --show-settings` reports 415 enabled rules instead of 61, and target-version resolves to 3.10 instead of py38. UP006 then fires 1483 times, UP045 1318, UP035 435.

**To Reproduce**

Clean checkout of `main` @ 97f531a:

```bash
pip install ruff==0.16.3
ruff check --fix   # make lint    -> 310 files
ruff format        # make format  -> 120 files
git status --porcelain | wc -l    # 380 files, +4062 -4152
```

With ruff 0.5.5 both commands change 0 files.

**Expected Behavior**

The documented commands leave unrelated files untouched.

**Why it matters**

DEVELOPMENT.md documents `make format` and `make lint` under "Helpful commands". A contributor who runs them either sends a first PR with a 4,000-line unrelated diff, or reverts and fights a hook that disagrees with the linter they were told to run.

**Suggested direction**

Align the pins so the hook and the lock name the same version. Which version is your call; the rule surface and target-version inference differ a lot between them. I can put up the PR if you pick one.

**Environment:** macOS, Python 3.11, trulens `main` @ 97f531a

Surfaced while preparing #2717.

sfc-gh-jreini:
Thanks @dylanpulver — can you make a PR updating the pre-commit Ruff version to v0.16.3 and aligning the relevant Ruff dependency/lockfile entries?

dylanpulver:
0.16.3 it is. One consequence to name before I open it.

Moving `rev` alone reformats the tree. The hook runs `ruff --fix` and `ruff-format` over `python, pyi, jupyter`; `[tool.ruff.lint] preview = true` turns on 415 rules under 0.16 against 61 under 0.5.5; and with no `target-version` set, 0.16 infers py310 from `tool.poetry.python = "^3.10"` where 0.5.5 fell back to py38. That combination is the 380-file diff in the report, UP006 1483 times, UP045 1318, UP035 435.

[...]

PR #2720, opened 2026-08-21
chore: align the three ruff pins on 0.16.3 (#2718)

dylanpulver:
# Description

Closes the ruff version split in #2718: `.pre-commit-config.yaml` pinned v0.5.5, `poetry.lock` held 0.16.0, `pyproject.toml` had `ruff = ">=0.5"`. All three now name 0.16.3, per @sfc-gh-jreini.

**Read commit 1.** It is the whole change: the three pins plus one config block. Commit 2 is machine output, no hand edits, reproducible from commit 1:

```
pip install ruff==0.16.3
ruff check --fix   # make lint
ruff format        # make format
```

120 files, +934 -645, formatter-only, byte-identical across three clean runs.

On the config block: `[tool.ruff.lint] select` was never set here, so the rule surface was ruff's default, and 0.16 changed that default from 60 rules to 413 (`--isolated`, no config at all). Commit 1 states the old four groups explicitly; `--show-settings` reports the same 61 rules under 0.16.3 as under 0.5.5.

Rejected alternative: inherit 0.16's new defaults. That reformats 380 files (UP006 1483, UP045 1318, UP035 435; 0.16 infers py310 from `python = "^3.10"`, 0.5.5 fell back to py38) and leaves the ruff hook failing on 1084 findings with no safe autofix. Adopting those rules is a separate, lint-policy PR.

The pyproject pin is exact, not `^0.16.3`: ruff changes preview formatting between patch releases.

This conflicts with my #2716 and #2717; I will rebase both once it lands.

## Type of change


sfc-gh-jreini (approved):
thanks for making trulens easier to work with and contribute to!

PR #2721, opened 2026-08-21
feat(datasets): immutable DatasetVersion snapshots and Run provenance (#2701)

### LINKED ISSUE #2701: Immutable DatasetVersion snapshots and Run provenance
## Problem

The current `Dataset` model stores a stable ID, name, and metadata while `GroundTruth` rows point directly to the dataset. Adding or replacing examples changes the effective test set without preserving which snapshot a run used. Users cannot reproduce an evaluation or compare test-set membership over time.

## Proposal

Keep `Dataset` as the stable identity and add immutable versions:

- `DatasetVersion`: version ID, dataset ID, parent version ID, description, source metadata, content hash, item count, and creation timestamp.
- `DatasetVersionItem`: version ID, stable item ID, query/input, expected response, expected contexts, metadata, and named splits.

Use `DatasetVersion` terminology consistently in Python, database columns, documentation, and UI. Do not expose revision aliases.

### SDK direction

```python
version = session.create_dataset_version(
    dataset_name="support-quality",
    dataframe=examples,
    column_spec={
        "input": "question",
        "ground_truth_output": "expected_answer",
        "metadata": "metadata",
    },
    splits={"regression": ["case-1", "case-2"]},
    description="Reviewed August failures",
)

loaded = session.get_dataset_version(
    dataset_name="support-quality",
    dataset_version_id=version.dataset_version_id,
)
```

### Version behavior

- Publishing creates a complete immutable snapshot.
- An item ID is content-addressed from normalized example content and an optional caller-supplied input ID.
- A dataset version ID is content-addressed from ordered item IDs plus stable version metadata.
- Creating identical content is idempotent and returns the existing version.
- Parent version is optional provenance, not a requirement to compute a delta.
- Named splits are properties of the version item and do not mutate later.
- Public APIs can compare two versions and report added, removed, and unchanged item IDs.

### Backward compatibility

- Existing `Dataset` and `GroundTruth` rows remain readable.
- Existing data is exposed as version zero through a migration or compatibility loader without rewriting original ground-truth payloads.
- Existing calls that specify only `dataset_name` resolve to the latest version.
- Reproducible workflows can pin `dataset_version_id` explicitly.
- `GroundTruth` remains supported as a compatibility input and output shape.

### Run integration

- Add optional `dataset_version_id` to `RunConfig`, `Run.SourceInfo`, and persisted run metadata.
- When a version is selected, `Run` loads that exact immutable snapshot.
- Run descriptions and comparison output expose the dataset version used.
- Existing DataFrame and table-backed runs remain unchanged.

## Non-goals

- Human-review or trace-curation workflows.
- Mutable drafts, approvals, or collaborative editing.
- Semantic deduplication.
- Object-store datasets or Snowflake ML Dataset integration.
- A new experiment abstraction.

## Acceptance criteria

- Published versions cannot be updated through public APIs.
- Earlier versions remain loadable after later versions are created.
- Identical content creation is idempotent.
- Named splits and metadata round-trip without changing item identity unexpectedly.
- Existing datasets appear as version zero and current APIs remain functional.
- Latest-version resolution is deterministic.
- Membership comparison correctly identifies added, removed, and unchanged examples.
- New runs can pin and report one exact `dataset_version_id`.
- DataFrame and table runs that do not use persisted datasets are unaffected.

## Tests

- Version creation from DataFrames and `GroundTruth` sequences.
- Immutability, idempotency, content hashing, and parent provenance.
- Named split and metadata round trips.
- Version-zero migration/compatibility fixture from current tables.
- Latest and exact-version lookup.
- Membership comparison.
- Run source provenance and `Run.compare` compatibility.
- SQLite baseline and PostgreSQL persistence contract.

## Critical files

- `src/core/trulens/core/schema/dataset.py`
- `src/core/trulens/core/schema/groundtruth.py`
- `src/core/trulens/core/database/sqlalchemy.py`
- `src/core/trulens/core/run.py`

Payal2000:
Closes #2701

## Problem

`Dataset` holds a stable identity while `GroundTruth` rows point straight at it, so adding or replacing examples changes the effective test set in place. An evaluation cannot be reproduced, and two runs cannot be compared with any confidence that they saw the same examples.

## What this adds

Immutable, content-addressed snapshots alongside the existing mutable tables.

- **`DatasetVersion` / `DatasetVersionItem`** (`schema/dataset.py`). A version id hashes the dataset, the ordered contents and the source metadata, so publishing identical content is idempotent and returns the existing version rather than duplicating or updating it. `description` and `parent_dataset_version_id` are provenance annotations and stay out of the hash.
- **Storage**: `dataset_version` and `dataset_version_item` tables, alembic revision `13`. `(dataset_id, version_index)` is unique so version history cannot fork, and `version_index` is the deterministic ordering key for latest-version resolution.
- **SDK**: `TruSession.create_dataset_version`, `get_dataset_version`, `list_dataset_versions`, `compare_dataset_versions`. Publishing accepts a DataFrame with a `column_spec`, or a `GroundTruth` sequence.
- **Run provenance**: optional `dataset_version_id` on `RunConfig` and `Run.SourceInfo`, persisted into run source info by both the OSS and Snowflake DAOs, read back by `run.start()`, and surfaced by `RunDiff.provenance()` — which warns when the two runs read different versions, since the deltas then mix app changes with test-set changes.
- **Docs**: a Dataset Versions component guide, plus updated schema and migration tables in `contributing/database.md`. API reference is generated from docstrings.

## Backward compatibility

Existing `Dataset` and `GroundTruth` rows are untouched. A dataset that predates versioning is exposed as **version zero**, reconstructed from its ground truth rows by a compatibility loader that reads but never rewrites the original payloads. Version zero is materialized as a real row on the first publish for that dataset, so it stays loadable afterwards and becomes the parent of the first published version.

The Snowflake `source_info` payload only carries `dataset_version_id` when a version is actually pinned, so unversioned runs send an unchanged payload.

## Three judgment calls worth reviewing

The issue left these underdetermined. Each is documented in the code; happy to flip any of them.

1. **Item identity excludes `meta` and `splits`.** The issue asks for ids from "normalized example content" but also that metadata "round-trip without changing item identity unexpectedly". If metadata were part of identity, re-tagging an example would show up in a diff as removed *and* added. So identity is `input` + `expected_response` + `expected_contexts` + `input_id`. To stop that from silently dropping annotations, the **version** hash covers the full per-item payload rather than only the ordered item ids as the issue literally specifies.

2. **`get_ground_truth(dataset_name=...)` is deliberately unchanged.** The issue says name-only calls should resolve to the latest version, but the acceptance criteria also require that current APIs keep working. Resolving to the latest version would mean that publishing a version of a *subset* silently narrows what existing callers read. Legacy semantics are kept and `dataset_version_id` is an opt-in parameter; latest-resolution applies to the new version-aware APIs.

3. **A dataset name can map to several rows**, because `dataset_id` hashes name *and* metadata. Versions hang off the primary row (lowest id) so that reading version zero and materializing it cannot disagree.

## Tests

58 unit tests in `tests/unit/test_dataset_versions.py` covering version creation from DataFrames and `GroundTruth` sequences, content hashing, idempotency, immutability, parent provenance, split and metadata round trips, version-zero compatibility, latest and exact lookup, membership comparison, and run source provenance.

A persistence-contract helper is wired into `tests/integration/test_database.py` for sqlite, postgres and mysql, asserting that a version round-trips to the same content hash it was stored under.

Verified locally: the new unit suite plus the existing run/DAO/dashboard tests pass, as does the sqlite persistence contract. Postgres and mysql need the `docker/test-database.yaml` services and will run in CI. `ruff format` and `ruff check` are clean.

## Unrelated bug noticed

`SQLAlchemyDB.get_datasets()` raises `AttributeError: 'Dataset' object has no attribute 'name'` — it reads `ds.name`/`ds.meta` off the ORM row, but those live inside `dataset_json`. This is broken on `main` today and is left alone here; this PR uses its own resolution path. Worth a separate fix.

### DIFF
--- a/src/core/trulens/core/database/sqlalchemy.py
+++ b/src/core/trulens/core/database/sqlalchemy.py
@@ -2165,6 +2165,23 @@ class SQLAlchemyDB(core_db.DB):
 
             return _dataset.dataset_id
 
+    def get_dataset(
+        self, dataset_name: str
+    ) -> Optional[dataset_schema.Dataset]:
+        """See [DB.get_dataset][trulens.core.database.base.DB.get_dataset]."""
+
+        with self.session.begin() as session:
+            rows = self._dataset_rows_by_name(session, dataset_name)
+            if not rows:
+                return None
+
+            payload = json.loads(rows[0].dataset_json)
+            return dataset_schema.Dataset(
+                name=payload.get("name", dataset_name),
+                dataset_id=rows[0].dataset_id,
+                meta=payload.get("meta"),
+            )
+
     def get_datasets(self) -> pd.DataFrame:
         """See [DB.get_datasets][trulens.core.database.base.DB.get_datasets]."""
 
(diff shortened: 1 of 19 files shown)


src/core/trulens/core/database/sqlalchemy.py
```diff
@@ -2146,6 +2146,23 @@ def insert_dataset(
...
+    ) -> Optional[dataset_schema.Dataset]:
+        """See [DB.get_dataset][trulens.core.database.base.DB.get_dataset]."""
+
+        with self.session.begin() as session:
+            rows = self._dataset_rows_by_name(session, dataset_name)
+            if not rows:
+                return None
+
+            payload = json.loads(rows[0].dataset_json)
+            return dataset_schema.Dataset(
```
joshreini1, line 2188: The TODO acknowledges the race but doesn't fix it. Two threads can read the same latest.version_index, increment to the same new index, then both try to insert—the unique constraint rejects one as a duplicate instead of serializing the history. Wrap the read-increment-write in a database-level lock or retry on constraint violation.

src/core/trulens/core/schema/dataset.py
```diff
@@ -2,15 +2,165 @@
...
+    ignored = []
+
+    for key, value in (column_spec or {}).items():
+        lowered = str(key).lower()
+        field = COLUMN_SPEC_ALIASES.get(lowered, lowered)
+        if field in COLUMN_SPEC_FIELDS:
+            normalized[field] = value
+        else:
+            ignored.append(key)
+
```
joshreini1, line 147: Change logger.debug to logger.warning. A user passing {"retrieval.query_text": "question"} expecting it to populate expected_contexts has misconfigured their spec; silent DEBUG drop means they'll only notice when items are missing data. Warn so the fix is obvious.

src/core/trulens/core/run.py
```diff
@@ -1936,8 +2029,26 @@ def compare_runs(
...
+    dataset_version_id_a = _dataset_version_id_of(run_a)
+    dataset_version_id_b = _dataset_version_id_of(run_b)
+
+    if (
+        dataset_version_id_a is not None
+        and dataset_version_id_b is not None
+        and dataset_version_id_a != dataset_version_id_b
+    ):
+        logger.warning(
+            "The two runs were pinned to different dataset versions (%s vs "
```
joshreini1: The warning format reads awkwardly—"pinned to different dataset versions (A vs B), so metric deltas mix..." Consider: "Runs pinned to different versions (A vs B); metric deltas include both app changes and test-set changes."

joshreini1 (changes requested):
Clean implementation of immutable dataset versioning with strong content-addressing guarantees and backward compatibility. The diff achieves exactly what #2701 specified: idempotent publishing, version-zero reconstruction, split/metadata round-trips, and run provenance. 58 unit tests anchor the contract. One BLOCKER: the alembic revision defines a unique constraint on (dataset_id, version_index) but the insert path does not handle the race where two concurrent publishes assign the same index—the TODO at sqlalchemy.py:2160 flags this but leaves the window open. One SHOULD-FIX: normalize_column_spec logs ignored keys at DEBUG level when the user may have genuinely misconfigured the mapping and expected those columns to be read. Verdict: Request changes—fix the concurrency window before merge.

src/core/trulens/core/database/sqlalchemy.py
Payal2000: Fixed in 3a3e9b06b — you were right that the TODO left the window open.

The publish now retries on the constraint violation: `insert_dataset_version` wraps a single-attempt `_insert_dataset_version_once` in a bounded loop that catches `IntegrityError`, re-reads the now-committed index, and tries again. A lost race serializes into the history instead of reaching the caller as an error.

I went with retry rather than a row lock because SQLite has no `SELECT ... FOR UPDATE`, so locking would only serialize the databases that support it and leave the default backend exactly as racy as before.

One subtlety worth flagging: each attempt restores the caller's original `parent_dataset_version_id` and `version_index` before retrying. Both are derived from whichever version turns out to be latest, so without that reset the second attempt would have kept the parent it computed from the stale read.

Four new tests in `TestConcurrentPublishing` race real publishes across threads, each on its own engine against the same file-backed database: distinct publishes all succeed, the resulting history is gapless with no index assigned twice, identical concurrent content stays idempotent, and the retry budget is bounded. I checked they actually exercise the race by setting the attempt budget back to 1 — two of the four fail, and pass again with the retry restored.

src/core/trulens/core/schema/dataset.py
Payal2000: Agreed, changed to `logger.warning` in 3a3e9b06b. It now also names the supported fields, since the whole problem is that the caller does not know which spelling was expected. Added a comment on the line explaining why it warns, so nobody quietly downgrades it later.

src/core/trulens/core/run.py
Payal2000: Taken as written in 3a3e9b06b — it now reads "Runs pinned to different dataset versions (%s vs %s); metric deltas include both app changes and test-set changes." Clearer than mine.

PR #2722, opened 2026-08-21
feat(datasets): curate production traces into persisted evaluation datasets (#2702)

### LINKED ISSUE #2702: Curate production traces into persisted evaluation datasets
## Problem

Production traces contain the examples users need for regression tests, but `add_ground_truth_to_dataset` expects an already prepared DataFrame with `query`, `expected_response`, and optional expected contexts. Users must manually extract record content, join review corrections, normalize fields, deduplicate, and handle malformed examples before creating an offline dataset.

## Proposal

Add a bounded synchronous curation utility that targets the `Dataset` and `GroundTruth` APIs available today. This issue is independently useful and does not require DatasetVersion or human-review queues.

### Mapping

```python
class TraceDatasetMapping(BaseModel):
    query: str = "input"
    query_id: str | None = "record_id"
    expected_response: str | None = None
    expected_chunks: str | None = None
    metadata: dict[str, str] = Field(default_factory=dict)
```

Mapping values identify input DataFrame columns, not expressions or arbitrary Python code.

### SDK direction

```python
result = session.curate_records_to_dataset(
    dataset_name="production-failures",
    records=records_df,
    mapping=TraceDatasetMapping(
        query="input",
        expected_response="corrected_output",
        metadata={
            "source_record_id": "record_id",
            "app_version": "app_version",
            "groundedness": "Groundedness",
        },
    ),
    expected_response_fn=optional_correction_callback,
    on_error="collect",
)
```

Accepted inputs:

- A DataFrame returned by `TruSession.get_records_and_feedback()`.
- A DataFrame containing explicit record IDs, resolved through current session APIs before mapping.
- A caller-provided CSV or JSON review export loaded into a DataFrame.

Behavior:

- Validate required source columns before processing.
- Resolve JSON/string inputs and outputs into stable text without private database methods.
- Use a mapped corrected output when present.
- Optionally call a user-provided synchronous callback to supply expected outputs; TruLens does not invoke an LLM automatically.
- Preserve source record ID, app name/version, metric scores, and selected fields in ground-truth metadata.
- Normalize expected contexts to `GroundTruth.expected_chunks`.
- Use current content-addressed ground-truth IDs for idempotency.
- Process and insert in bounded batches through existing database methods.
- Collect per-row validation errors or fail fast according to `on_error`.
- Return accepted, duplicate, and rejected counts plus error rows.

### Cookbook notebook

Add a credential-free notebook that:

1. Loads a small synthetic app dataset and retrieves records with `get_records_and_feedback()`.
2. Selects low-scoring or error records in pandas.
3. Adds corrected expected responses in the notebook.
4. Curates them into a current persisted dataset.
5. Loads it through `get_ground_truth()` and uses it in a ground-truth metric.

## Non-goals

- Dataset versioning.
- Annotation or review persistence.
- Background curation workers.
- Automatic expected-output generation.
- Query languages or arbitrary field expressions.
- Mutating source traces.

## Acceptance criteria

- Every accepted item maps to a valid current `GroundTruth` row.
- Curating identical rows twice does not duplicate ground truths.
- Required-column failures are reported before database writes.
- Corrected outputs and expected contexts round-trip through `get_ground_truth()`.
- Source provenance remains available in ground-truth metadata.
- Large DataFrames process with memory bounded by the batch size beyond the input DataFrame itself.
- Collect-errors mode does not publish malformed rows.
- The cookbook runs with synthetic data and current APIs only.

## Tests

- Mapping validation and JSON/text normalization.
- Direct DataFrame, record-ID DataFrame, CSV, and JSON inputs.
- Callback precedence and callback failures.
- Context normalization and metadata selection.
- Duplicate/idempotency behavior.
- Fail-fast and collect-errors modes.
- Batch boundaries.
- Credential-free cookbook execution or helper unit coverage.

## Critical files

- `src/core/trulens/core/dataset.py`
- `src/core/trulens/core/session.py`
- `src/core/trulens/core/schema/groundtruth.py`
- `examples/expositional/use_cases/trace_to_dataset.ipynb`
- `docs/cookbook/use_cases/trace_to_dataset.ipynb`

Payal2000:
Closes #2702

## Problem

Production traces already hold the examples a regression test needs, but `add_ground_truth_to_dataset` only accepts an already-prepared DataFrame. Getting from one to the other means manually extracting record content, joining in review corrections, normalizing fields, deduplicating, and deciding what to do with malformed rows.

## What this adds

`TruSession.curate_records_to_dataset()`, built entirely on the `Dataset` and `GroundTruth` APIs that exist today — no dataset versioning, no review queues.

- **`TraceDatasetMapping`** names which dataframe column supplies which ground truth field. Values are column names, never expressions, so the mapping is validated against the dataframe before anything is written.
- **Normalization**: recorded inputs and outputs resolve to stable text whether they arrive as strings (OTEL) or as already-parsed JSON (legacy schema), with dicts re-serialized under sorted keys so key ordering cannot change a content-addressed id. Expected contexts normalize into the `expected_chunks` shape from lists, strings, dicts, or a JSON encoding of any of those.
- **Accepted inputs**: a `get_records_and_feedback()` frame; a frame carrying only record ids, resolved through the session before mapping (columns the caller supplies always win over the stored ones); or a CSV/JSON review export loaded with pandas.
- **Expected outputs**: a mapped correction always wins; an optional synchronous callback fills in only what is missing. TruLens never calls an LLM on your behalf.
- **Provenance**: source record id and app name/version are preserved in ground truth metadata, alongside any metric scores you map explicitly.
- **Bounded work**: rows are curated lazily and written in batches through the existing `batch_insert_ground_truth`, so peak memory is one batch beyond the input frame.
- **Error handling**: `on_error="raise"` fails on the first unusable row, `on_error="collect"` skips and reports it. Malformed rows are never written in either mode. The result carries accepted / duplicate / rejected counts, the ids written, and an `errors_df()` for inspection.

## Cookbook

`examples/expositional/use_cases/trace_to_dataset.ipynb` walks the whole loop on synthetic data with **no API keys**: record a deliberately imperfect support bot, score it with a deterministic metric, select the low-scoring records in pandas, write corrections, curate them, load the dataset back through `get_ground_truth()`, and use it as a ground truth metric against a fixed version of the bot.

Note that `docs/cookbook` is a symlink to `examples/expositional`, so the two notebook paths named in the issue are the same file.

## One addition beyond the issue

`include_provenance=False`. A `GroundTruth` id hashes its metadata, so automatically attaching `source_record_id` means two records with identical content never deduplicate — the issue asks for both provenance in metadata and deduplication, and these pull against each other. Provenance stays on by default (it is the explicit acceptance criterion); this is the escape hatch for callers who want to deduplicate on example content alone. It is documented on the `duplicates` field.

## Tests

65 unit tests in `tests/unit/test_trace_curation.py`, covering every bullet on the issue's test list: mapping validation, JSON/text normalization, all four input shapes, callback precedence and callback failure, context normalization and metadata selection, duplicate/idempotency behavior, both error modes, batch boundaries, and an end-to-end test mirroring the cookbook so the notebook's narrative cannot silently drift from the API.

The notebook itself was executed end to end with nbclient and runs clean with no credentials.

Verified locally: the full non-OTEL unit suite goes from 317 to 382 passing with no new failures — the remaining failures are byte-identical on a clean tree. Formatting and lint are clean under the ruff version pinned in `.pre-commit-config.yaml`.

## Notes for review

- `src/core/trulens/core/schema/groundtruth.py` is listed among the issue's critical files but is unchanged here: curation constructs `GroundTruth` through its existing constructor and nothing in the schema needed to move. Happy to relocate the conversion helpers there if that is the preferred home.
- Duplicate counting is scoped to a single call. Cross-call idempotency is guaranteed by content-addressed ids plus the existing upsert (and tested), but reporting it would have needed a new batched-existence query on the database, which the issue rules out by asking that this go "through existing database methods".

### DIFF
--- a/src/core/trulens/core/dataset.py
+++ b/src/core/trulens/core/dataset.py
@@ -185,10 +185,10 @@ class CurationRowError(ValueError):
     [CurationError][trulens.core.dataset.CurationError] in `collect` mode.
     """
 
-    def __init__(self, reason: str, message: str, row_index: Any = None):
+    def __init__(self, row_index: Any, reason: str, message: str):
+        self.row_index = row_index
         self.reason = reason
         self.message = message
-        self.row_index = row_index
         super().__init__(f"row {row_index}: {message}")
 
 
@@ -208,7 +208,7 @@ def _is_missing(value: Any) -> bool:
     return False
 
 
-def _normalize_text(value: Any) -> Optional[str]:
+def _normalize_text(value: Any) -> str | None:
     """Resolve a recorded input or output into stable text.
 
     Records carry main inputs and outputs as strings in OTEL mode and as
@@ -240,7 +240,7 @@ def _normalize_text(value: Any) -> Optional[str]:
     return text
 
 
-def _normalize_contexts(value: Any) -> Optional[List[Dict[str, Any]]]:
+def _normalize_contexts(value: Any) -> list[dict[str, Any]] | None:
     """Normalize expected contexts into `GroundTruth.expected_chunks` shape.
 
     Accepts a list of dicts, a list of strings, a single string or dict, or a
@@ -335,7 +335,7 @@ def _ground_truth_of_row(
     dataset_id: types_schema.DatasetID,
     mapping: TraceDatasetMapping,
     available: Set[str],
-    expected_response_fn: Optional[Callable[[pd.Series], Optional[str]]],
+    expected_response_fn: Callable[[pd.Series], str | None] | None,
     include_provenance: bool = True,
 ) -> groundtruth_schema.GroundTruth:
     """Turn one dataframe row into a ground truth.
@@ -411,7 +411,7 @@ def _batched(items: Iterable[Any], batch_size: int) -> Iterable[List[Any]]:
 def _resolve_records(
     records: pd.DataFrame,
     mapping: TraceDatasetMapping,
-    record_resolver: Optional[Callable[[List[str]], pd.DataFrame]],
+    record_resolver: Callable[[list[str]], pd.DataFrame] | None,
 ) -> pd.DataFrame:
     """Fill in record content for a dataframe that only carries record ids.
 
@@ -481,13 +481,13 @@ def curate_records_to_dataset(
     dataset_name: str,
     records: pd.DataFrame,
     db: Any,
-    mapping: Optional[TraceDatasetMapping] = None,
-    expected_response_fn: Optional[Callable[[pd.Series], Optional[str]]] = None,
-    dataset_metadata: Optional[Dict[str, Any]] = None,
+    mapping: TraceDatasetMapping | None = None,
+    expected_response_fn: Callable[[pd.Series], str | None] | None = None,
+    dataset_metadata: dict[str, Any] | None = None,
     on_error: str = ON_ERROR_RAISE,
     batch_size: int = DEFAULT_BATCH_SIZE,
     include_provenance: bool = True,
-    record_resolver: Optional[Callable[[List[str]], pd.DataFrame]] = None,
+    record_resolver: Callable[[list[str]], pd.DataFrame] | None = None,
 ) -> CurationResult:
     """Curate a records dataframe into a persisted dataset.
 
--- a/src/core/trulens/core/dataset.py
+++ b/src/core/trulens/core/dataset.py
@@ -1,0 +1,617 @@
+"""Curation of recorded traces into persisted evaluation datasets.
+
+Production traces already contain the examples a regression test needs, but
+[add_ground_truth_to_dataset][trulens.core.session.TruSession.add_ground_truth_to_dataset]
+expects a dataframe that has already been extracted, joined, normalized and
+deduplicated. This module does that preparation, mapping a records dataframe
+onto [GroundTruth][trulens.core.schema.groundtruth.GroundTruth] rows through
+the existing dataset APIs.
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+import math
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    Iterable,
+    List,
+    Optional,
+    Set,
+)
+
+import pandas as pd
+import pydantic
+from pydantic import BaseModel
+from pydantic import Field
+from trulens.core.schema import dataset as dataset_schema
+from trulens.core.schema import groundtruth as groundtruth_schema
+from trulens.core.schema import types as types_schema
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_BATCH_SIZE = 100
+"""Number of ground truths written per database round trip."""
+
+ON_ERROR_RAISE = "raise"
+"""Fail on the first row that cannot be curated."""
+
+ON_ERROR_COLLECT = "collect"
+"""Skip rows that cannot be curated and report them in the result."""
+
+ON_ERROR_MODES = (ON_ERROR_RAISE, ON_ERROR_COLLECT)
+
+PROVENANCE_COLUMNS = {
+    "source_record_id": "record_id",
+    "source_app_name": "app_name",
+    "source_app_version": "app_version",
+}
+"""Metadata keys automatically preserved from a records dataframe.
+
+Each is copied only when the source column is present and the caller has not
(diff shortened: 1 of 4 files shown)


src/core/trulens/core/dataset.py
```diff
@@ -0,0 +1,617 @@
...
+            ],
+            columns=["row_index", "query_id", "reason", "message"],
+        )
+
+
+class CurationRowError(ValueError):
+    """A single row could not be curated.
+
+    Raised out of `curate_records_to_dataset` in `raise` mode; converted into a
+    [CurationError][trulens.core.dataset.CurationError] in `collect` mode.
```
joshreini1, line 185: The CurationRowError.__init__ accepts row_index as a kwarg-only parameter but assigns it as an instance attribute. Consider making it a required positional arg for consistency with how it's constructed throughout the module (always with row_index in the raise statements).

src/core/trulens/core/dataset.py
```diff
@@ -0,0 +1,617 @@
...
+    ]
+    if not record_ids:
+        return records
+
+    logger.info(
+        "Resolving %d record(s) to fill in missing column(s): %s",
+        len(record_ids),
+        ", ".join(missing),
+    )
+    try:
```
joshreini1, line 451: The expected_response_fn callback receives a pandas Series with signature Callable[[pd.Series], Optional[str]]. Is there a reason this isn't typed as Callable[[pd.Series], str | None] using the modern union syntax given from __future__ import annotations is present at the top?

tests/unit/test_trace_curation.py
```diff
@@ -0,0 +1,858 @@
...
+        ])
+
+    def curate_many(self, count: int, batch_size: int, **kwargs):
+        db = _RecordingDB()
+        result = core_dataset.curate_records_to_dataset(
+            dataset_name="batched",
+            records=self.make_records(count),
+            db=db,
+            mapping=core_dataset.TraceDatasetMapping(
+                query="input", query_id="record_id"
```
joshreini1, line 555: The _RecordingDB stub doesn't implement the full db interface (e.g., no get_ground_truth method). While sufficient for these tests, consider adding a docstring noting it's a minimal stub covering only batch insertion operations to clarify scope for future maintainers.

joshreini1 (approved):
Solid implementation of trace-to-dataset curation with comprehensive test coverage (65 unit tests), clear API design, and good documentation. The PR delivers exactly what the issue spec asks for: mapping validation, normalization, multiple input shapes, batch writing, error handling modes, and provenance tracking. The notebook demonstrates the full workflow with no credentials required. Ready to merge — no blocking issues found.

src/core/trulens/core/dataset.py
Payal2000: Done in 213937b7f — `row_index` is now the first, required positional argument, and the assignments follow the same order. Every raise site already passed it, so nothing else changed.

src/core/trulens/core/dataset.py
Payal2000: Good catch, switched in 213937b7f — that signature is now `Callable[[pd.Series], str | None] | None`, along with the other function signatures in the module.

One deliberate exception: the pydantic field annotations on `TraceDatasetMapping` and `CurationError` keep `Optional[...]`. `from __future__ import annotations` makes function annotations lazy strings that never get evaluated, but pydantic *does* resolve field annotations at runtime, and `str | None` raises `TypeError: unsupported operand type(s)` on the Python 3.9 that `src/core/pyproject.toml` still declares (`python = "^3.9"`). That matches what the rest of the codebase does — `sqlalchemy.py` and `otel/instrument.py` use `X | None` in function signatures while everything in `core/schema/` stays on `Optional`. Worth noting CI only runs 3.10+, so a 3.9 break here would not be caught.

tests/unit/test_trace_curation.py
Payal2000: Added in 213937b7f. The docstring now says it is deliberately not a full `DB` — it covers only the two calls curation makes (`insert_dataset` and `batch_insert_ground_truth`), has no read side at all, and anything needing to load a dataset back has to use a real database the way the other tests here do.

PR #2726, opened 2026-08-25
fix(feedback): stop reading a judge's stated scale as its rating

arthi-arumugam-git:
## The bug

`re_configured_rating` collects every number in the judge's answer, keeps the ones inside the
configured scale, and returns `min()` of them. So when a judge states its own scale before
answering, the bounds it quoted are themselves in range and the minimum wins.

On the default 0-3 scale, against released `trulens-feedback 2.13.1`:

| judge's answer | parsed | the judge meant |
|---|---|---|
| `On a scale of 0 to 3, I rate this 3` | **0** | 3 |
| `Score 0-3: the context is fully relevant, so 3` | **0** | 3 |
| `Rating scale 0 to 3. My rating: 2` | **0** | 2 |
| `Given the criteria (0=irrelevant, 3=highly relevant), I give 3` | **0** | 3 |
| `The answer is fully supported. Score: 3` | 3 | 3 |

Nothing raises. `0.0` is a real score on that scale, so from the outside a floored score and a
genuinely bad one look the same. The only signal is a `logger.warning`.

This is already written down in the repo. `tests/unit/test_feedback_score_generation.py` has:

```python
("In the range of 0-10, I give this a 9.", 0),
# Currently does not have ideal handling as it returns the minimum integer found.
```

## How this differs from #2631

#2631 and its fix in #2636 were about the OpenAI provider not recognising `custom_tool_call`, so
the whole response JSON was handed to the parser. That fix stopped the blob arriving. It did not
change the parser, and this needs no tool calls and no gpt-5: plain text is enough.

## The change

A new `_strip_scale_statement` removes an explicit scale statement in the two shapes judges
actually produce, a range (`0 to 3`, `0-3`, `between 0 and 3`, `0 through 3`) and a legend
(`0 = irrelevant`, `3 = highly relevant`). Only the configured bounds are removed. If exactly one
candidate survives, that is the rating.

I kept it deliberately narrow. Anything else falls through to the existing code path untouched, so
`N out of M` phrasing keeps its current behaviour. That matters because
`tests/integration/test_score_parsing_pipeline.py` pins `"I rate this an 8 out of 10." -> 0.8` and
the unit tests pin `"The rating is 1 out of 3." -> 1`. Both still pass.

I did not change the `-> int` return annotation, which is already inaccurate when
`allow_decimal=True`. That felt like a separate change and I did not want to widen this one.

## Tests

11 new cases covering both scales and both shapes. All 11 fail on `main` and pass here. The
existing pinned case above is updated from `0` to `9` and its comment removed, since the handling
it describes is what this fixes.

```
new + existing unit tests                                   25 passed
test_score_parsing_normalization / _pipeline / _range_validation   46 passed
                                                            71 passed total
ruff check .                                                All checks passed
ruff format --check .                                       2 files already formatted
```

### DIFF
--- a/src/feedback/trulens/feedback/generated.py
+++ b/src/feedback/trulens/feedback/generated.py
@@ -39,6 +39,35 @@ PATTERN_INTEGER: re.Pattern = re.compile(r"([+-]?[1-9][0-9]*|0)")
 """Regex that matches integers."""
 
 
+def _ratings_in_range(
+    s: str, min_score_val: int, max_score_val: int, allow_decimal: bool
+) -> set:
+    """Every number in ``s`` that falls inside the rating scale.
+
+    Args:
+        s: String to read numbers from.
+
+        min_score_val: Minimum value of the rating scale.
+
+        max_score_val: Maximum value of the rating scale.
+
+        allow_decimal: Whether to capture decimal numbers rather than truncate.
+
+    Returns:
+        set: The in-range values. Out-of-range numbers are logged and dropped.
+    """
+    found = set()
+    for match in PATTERN_NUMBER.findall(s):
+        rating = float(match) if allow_decimal else int(float(match))
+        if min_score_val <= rating <= max_score_val:
+            found.add(rating)
+        else:
+            logger.warning(
+                "Rating must be in [%s, %s].", min_score_val, max_score_val
+            )
+    return found
+
+
 def _strip_scale_statement(
     s: str, min_score_val: int, max_score_val: int
 ) -> str:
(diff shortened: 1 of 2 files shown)


src/feedback/trulens/feedback/generated.py
```diff
@@ -75,6 +108,26 @@ def re_configured_rating(
     if not matches:
         raise ParseError("int or float number", s, pattern=PATTERN_NUMBER)
 
+    def in_range(text: str) -> set:
```
joshreini1: Extract in_range as a module-level helper or refactor to eliminate duplication—the range-checking logic here (lines 111-117) duplicates the loop at lines 130-136, creating future maintenance burden when one is updated without the other.

src/feedback/trulens/feedback/generated.py
```diff
@@ -75,6 +108,26 @@ def re_configured_rating(
...
+    # as candidate ratings and the min() below returns the floor instead of the
+    # rating. Drop an explicit scale statement first, and if exactly one candidate
+    # survives, that is the rating. Anything else falls through unchanged, so
+    # "N out of M" phrasing keeps its existing behaviour.
+    without_scale = _strip_scale_statement(s, min_score_val, max_score_val)
+    if without_scale != s:
+        narrowed = in_range(without_scale)
+        if len(narrowed) == 1:
+            return narrowed.pop()
+
```
joshreini1, line 147: When without_scale != s, this fallthrough path still uses matches from the original un-stripped string (line 108), meaning cases like 'scale of 0 to 3, I give 2 or 3' will re-inject the stripped bounds as candidates. Consider recomputing matches = PATTERN_NUMBER.findall(without_scale) before this block when a scale was detected, so the stripped version is used consistently throughout.

joshreini1 (approved):
Strong fix for the stated scale parsing bug. The implementation correctly handles the common case (scale statement + one rating) and preserves backward compatibility with 'N out of M' phrasing. Tests are comprehensive. Two maintainability concerns: the new in_range helper duplicates existing range-checking logic, and the fallthrough path uses matches from the original string even when a scale was stripped, which could confuse future readers. Neither blocks merge. Verdict: Approve with 2 SHOULD-FIX comments.

arthi-arumugam-git:
Both fixed, thanks. The second one was a real bug rather than a readability point, and I had not tested that shape.

`scale of 0 to 3, I give 2 or 3` returned 0 on my branch. The early return only fired when stripping left exactly one candidate; with two the code fell back to the whole string, the bounds came back, and `min()` picked the floor again. So the original bug survived in a narrower form.

The stripped text is now used throughout rather than only for the early return, with a fallback to the original string if stripping takes every number with it. The range check is extracted as `_ratings_in_range`, so the loop is not duplicated and the out-of-range warning lives in one place.

Added your two cases as regression tests. Full score-parsing suite is 73 passed, ruff clean.

joshreini1:
Thank you for making TruLens more reliable!

PR #2728, opened 2026-08-25
fix: keep temperature positional on conversation metrics

### LINKED ISSUE #2727: [BUG] Conversation metrics raise TypeError when temperature is passed positionally
**Bug Description**
PR #2710 added an `additional_instructions` parameter to the conversation metrics, but it was inserted before the existing `temperature` parameter. Any existing code that passed temperature positionally, like `provider.conversation_helpfulness(records, 0.7)`, now binds `0.7` to `additional_instructions` and raises a `TypeError` instead of producing a score.

Affected methods at HEAD (026a268), all in `src/feedback/trulens/feedback/llm_provider.py`:

- `conversation_helpfulness` (L4400) and `conversation_helpfulness_with_cot_reasons` (L4449)
- `topic_adherence` (L4502) and `topic_adherence_with_cot_reasons` (L4552)
- `agent_goal_accuracy` (L4604) and `agent_goal_accuracy_with_cot_reasons` (L4655)
- `coherence_across_turns` (L4708) and `coherence_across_turns_with_cot_reasons` (L4754)

Permalink: https://github.com/truera/trulens/blob/026a26851bc3a7343bbea6dbb8480c57e953141b/src/feedback/trulens/feedback/llm_provider.py#L4400-L4405

```python
def conversation_helpfulness(
    self,
    records: Union[List[Any], str],
    additional_instructions: Optional[str] = None,
    temperature: float = 0.0,
) -> float:
```

Before #2710 the signature was `(self, records, temperature: float = 0.0)`, so `temperature` was the second positional parameter.

The crash happens in `_build_criteria_with_instructions` (same file, L1362-1378). The conversation templates have no `{additional_instructions}` placeholder, so a truthy value falls into the string concatenation branch:

```python
elif additional_instructions:
    # Append as a new section for prompts without the placeholder
    final_criteria += (
        "\n\nAdditional instructions:\n" + additional_instructions
    )
```

With a float that raises `TypeError: can only concatenate str (not "float") to str`.

**To Reproduce**

1. On trulens before #2710, use any conversation metric with a positional temperature, for example `provider.conversation_helpfulness(records, 0.7)`. It works.
2. Upgrade to current main (or any build containing #2710).
3. Run the same call.
4. It raises `TypeError` before the provider is invoked. The keyword form `temperature=0.7` still works.

One extra wrinkle: `provider.conversation_helpfulness(records, 0.0)` does not crash, because `0.0` is falsy and skips the concatenation branch. So the break only shows up for non-zero temperatures, which makes it easy to miss in a quick test.

**Expected Behavior**
Existing positional callers keep working, or the change is called out as breaking. Two low-cost options: append `additional_instructions` after `temperature` instead of before it, or make the new parameter keyword-only (`*,` after `records`) so misuse fails loudly and documented calls are unaffected.

**Relevant Logs/Tracebacks**
TypeError: can only concatenate str (not "float") to str
  (raised from _build_criteria_with_instructions, llm_provider.py line 1374, via conversation_helpfulness)

**Environment:**

- OS: any
- TruLens version: current main (026a268); introduced by #2710
- Versions of other relevant installed libraries: n/a

**Additional Context**
The same parameter-ordering pattern was applied to all eight conversation metric methods listed above, so a fix should cover the whole group.

Found while running [Ito](https://ito.ai) (AI code review, free for open source) against recently merged PRs. Full analysis: https://app.ito.ai/share/3bb6096d-ebf1-4d52-9f53-f16723d0923e.

BetterAndBetterII:
## Summary

PR #2710 inserted `additional_instructions` before `temperature` on the eight conversation metrics. Existing callers that passed temperature positionally (e.g. `provider.conversation_helpfulness(records, 0.7)`) then bound the float to `additional_instructions` and raised `TypeError` inside `_build_criteria_with_instructions`.

Restore `temperature` to its original positional slot and make `additional_instructions` keyword-only on:

- `conversation_helpfulness` / `_with_cot_reasons`
- `topic_adherence` / `_with_cot_reasons`
- `agent_goal_accuracy` / `_with_cot_reasons`
- `coherence_across_turns` / `_with_cot_reasons`

## Test plan

- [x] `pytest tests/unit/test_templates_conversation.py tests/unit/test_feedback_criteria_and_additional_instructions.py` (42 passed)
- [x] Positional temperature is forwarded to `_create_chat_completion_score`
- [x] Keyword `additional_instructions` still reaches `_build_criteria_with_instructions`

Fixes #2727

### DIFF
--- a/src/feedback/trulens/feedback/llm_provider.py
+++ b/src/feedback/trulens/feedback/llm_provider.py
@@ -4400,8 +4400,9 @@ class LLMProvider(core_provider.Provider):
     def conversation_helpfulness(
         self,
         records: Union[List[Any], str],
-        additional_instructions: Optional[str] = None,
         temperature: float = 0.0,
+        *,
+        additional_instructions: Optional[str] = None,
     ) -> float:
         """
         Uses chat completion model. A function that completes a template to
@@ -4419,8 +4420,8 @@ class LLMProvider(core_provider.Provider):
 
         Args:
             records (Union[List[Any], str]): The ordered conversation records, or a transcript string.
-            additional_instructions (Optional[str]): If provided, adds instructions to default criteria for the judge to follow. Defaults to None.
             temperature (float): The temperature for the LLM response, which might have impact on the confidence level of the evaluation. Defaults to 0.0.
+            additional_instructions (Optional[str]): If provided, adds instructions to default criteria for the judge to follow. Defaults to None.
 
         Returns:
             float: A value between 0.0 (not helpful) and 1.0 (helpful).
@@ -4449,8 +4450,9 @@ class LLMProvider(core_provider.Provider):
     def conversation_helpfulness_with_cot_reasons(
         self,
         records: Union[List[Any], str],
-        additional_instructions: Optional[str] = None,
         temperature: float = 0.0,
+        *,
+        additional_instructions: Optional[str] = None,
     ) -> Tuple[float, Dict]:
         """
         Uses chat completion model. A function that completes a template to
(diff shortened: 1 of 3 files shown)


joshreini1 (approved):
Clean fix for the positional temperature regression introduced in PR #2710. The parameter reordering restores backward compatibility for existing callers passing temperature positionally, while making additional_instructions keyword-only prevents future collisions. Test coverage is comprehensive with 42 passing tests including parametrized cases for all eight conversation metrics. Ready to merge.

PR #2730, opened 2026-08-26
fix(database): replace SQL JSON subscript with Python-side aggregation in pre-OTEL leaderboard

### LINKED ISSUE #2729: [BUG] get_leaderboard_aggregates always crashes when TRULENS_OTEL_TRACING is disabled
**Bug Description**
When OTEL tracing is disabled (`TRULENS_OTEL_TRACING=0` or `false`), `get_leaderboard_aggregates` routes to a fallback query path that can never run. It raises `NotImplementedError` while building the SQL statement, before touching the database. So anyone running in the supported disabled-tracing mode loses the leaderboard aggregates entirely.

The routing is in `src/core/trulens/core/database/sqlalchemy.py` at HEAD (026a268):

https://github.com/truera/trulens/blob/026a26851bc3a7343bbea6dbb8480c57e953141b/src/core/trulens/core/database/sqlalchemy.py#L1834-L1848

```python
def get_leaderboard_aggregates(self, ...):
    if is_otel_tracing_enabled():
        return self._get_leaderboard_aggregates_otel(...)
    return self._get_leaderboard_aggregates_pre_otel(...)
```

`_get_leaderboard_aggregates_pre_otel` (L1712) builds these expressions (L1725-1731):

```python
sa.func.avg(
    self.orm.Record.cost_json["n_tokens"].as_float()
).label("Total Tokens"),
...
sa.func.sum(self.orm.Record.cost_json["cost"].as_float()).label(
    "Total Cost (USD)"
),
```

But `cost_json` is not a JSON column. It's declared as `Text` in `src/core/trulens/core/database/orm.py` (L35 `TYPE_JSON = Text`, L238 `cost_json = Column(TYPE_JSON, nullable=False)`):

https://github.com/truera/trulens/blob/026a26851bc3a7343bbea6dbb8480c57e953141b/src/core/trulens/core/database/orm.py#L235-L239

SQLAlchemy does not support `[]` indexing on `Text` columns, so `cost_json["n_tokens"]` raises `NotImplementedError: Operator 'getitem' is not supported on this expression` at statement construction time. The failure is unconditional: this code path cannot succeed against any database backend.

**To Reproduce**

1. Set `TRULENS_OTEL_TRACING=0`.
2. Create a `TruSession` with the default SQLite connector and record at least one app run (or use any existing database).
3. Call `session.connector.db.get_leaderboard_aggregates()`, or open the dashboard leaderboard, which calls it.
4. Observe `NotImplementedError` from SQLAlchemy instead of a leaderboard DataFrame.

**Expected Behavior**
With tracing disabled, the leaderboard should still aggregate from the pre-OTEL `Record` table. A fix could extract the JSON fields with a dialect-appropriate expression (for example `sa.func.json_extract` on the text column, similar in spirit to `_json_extract_otel` at L869 which the OTEL path uses on real JSON columns), or fall back to loading `cost_json` and aggregating in Python.

**Relevant Logs/Tracebacks**
NotImplementedError: Operator 'getitem' is not supported on this expression
  (raised while constructing the record_stmt select in _get_leaderboard_aggregates_pre_otel, sqlalchemy.py line 1726)

**Environment:**

- OS: any
- TruLens version: current main (026a268)
- Versions of other relevant installed libraries: SQLAlchemy (any supported version)

**Additional Context**
The bug predates #2724, but #2724 formalized the disabled-tracing mode with an explicit warning, and this fallback is the path that mode routes to. The OTEL-enabled path is unaffected.

Found while running [Ito](https://ito.ai) (AI code review, free for open source) against recently merged PRs. Full analysis: https://app.ito.ai/share/0ffde7ec-a3c9-4da6-ae41-7eab316aed80.

Aftabbs:
## Problem

`get_leaderboard_aggregates()` crashes with `NotImplementedError` whenever `TRULENS_OTEL_TRACING` is not enabled (the default for most users).

The root cause is in `_get_leaderboard_aggregates_pre_otel`: the SQL query uses SQLAlchemy's `[]` subscript operator to extract values from `cost_json` and references a non-existent `Record.latency` column:

```python
sa.func.avg(self.orm.Record.cost_json["n_tokens"].as_float())  # raises NotImplementedError
sa.func.avg(self.orm.Record.latency)                            # AttributeError – column doesn't exist
```

`cost_json` and `perf_json` are declared as `TYPE_JSON = Text` (`orm.py:35`), so they are plain TEXT columns. SQLAlchemy raises `NotImplementedError` for `[]` subscript access on `Text` columns at _statement construction time_, before any query ever hits the database.

Fixes #2729.

## Fix

Replace the broken SQL-level aggregation with Python-side aggregation using the existing module-level helpers `_extract_tokens_and_cost` and `_extract_latency` (already used by the `get_records_and_feedback` code path):

1. `record_stmt` now selects individual rows with raw `cost_json` and `perf_json` text columns (no subscript, no missing `latency` column).
2. After fetching, parse cost/latency with `_extract_tokens_and_cost` / `_extract_latency`.
3. Use `pandas.groupby(...).agg(...)` for the per-app aggregation that was previously done in SQL.

The output schema (`Records`, `Total Tokens`, `Average Latency (s)`, `Total Cost (USD)`) is unchanged.

## Tests

Added `tests/unit/test_leaderboard_pre_otel.py` with three test cases:

- **empty database** – returns empty DataFrame without raising.
- **records present** – aggregated `Records`, `Total Tokens`, `Total Cost (USD)`, and `Average Latency (s)` are computed correctly.
- **`app_name` filter** – only the matching app's rows are returned.

All tests run against an in-memory SQLite database, matching the existing test pattern used in `test_dashboard_utils.py`.

## Checklist

- [x] Bug reproduced and root cause confirmed in `orm.py` (`TYPE_JSON = Text`)
- [x] Fix avoids dialect-specific JSON extraction (works with SQLite, PostgreSQL, Snowflake)
- [x] Output schema preserved
- [x] Unit tests added


joshreini1 (approved):
Retracted. This review relied on the incorrect premise that most users have OTEL disabled. OTEL has been the default for more than a year, and the relevant path is OTEL usage. SQL aggregation is also substantially more scalable than moving aggregation into frontend Python.

joshreini1 (changes requested):
OTEL has been enabled by default for more than a year and its aggregation path already works. This PR should keep its scope on the legacy non-OTEL path, but implement that aggregation in SQL rather than moving it into frontend Python. SQL aggregation is materially more scalable and avoids transferring raw records to the frontend for processing.

Aftabbs:

tonydzi (commented):
Hi, Mycroft here, Tony's synthetic AI co-founder. No body, no coffee, so I spend my free hours re-running other people's leaderboards. I wrote this review myself; Tony answers for it.

The SQL version was pushed on 08-29 and hasn't been re-reviewed since the change request, so I checked the current head (`e92c412`) against current `main` (`d1fb563`). Short version: it fixes #2729, but the legacy path still adds USD and Snowflake credits together. #2766 fixed exactly that on the OTEL path.

**Setup:** trulens-core 2.14.0 (editable from `main`), SQLAlchemy 2.0.52, Python 3.11.15, SQLite. `TRULENS_OTEL_TRACING=0` is set before import. I write two records through `TruBasicApp` and then call `TruSession().connector.db.get_leaderboard_aggregates()`, the same call the dashboard Leaderboard/Trends tabs make.

| case | `main` | `main` + #2730 | #2730 + split below |
|---|---|---|---|
| #2729 repro: 100 tok/$0.01 + 300 tok/$0.05 | `NotImplementedError: Operator 'getitem' is not supported on this expression` (3/3) | passes: tokens 400.0, USD 0.06 (3/3) | passes (3/3) |
| control: `get_records_and_feedback()`, OTEL off | pass | pass | pass |
| control: leaderboard with OTEL **on** | pass | pass | pass |
| **mixed currency:** USD 10.0 + `Snowflake credits` 1000.0 | same crash | **USD = 1010.0, Snowflake Credits = 0.0** | USD = 10.0, Snowflake Credits = 1000.0 |

The last row comes from two things in the PR:

- `sa.func.sum(cost_expr).label("Total Cost (USD)")` sums every record, whatever its `cost_json.cost_currency` says.
- Further down, `base_df["Total Cost (Snowflake Credits)"] = 0.0` hardcodes the credits column.

This isn't a regression, because `main` crashes before it ever gets there. But once this merges, a legacy-mode Snowflake user sees credits reported as dollars, and nothing errors. The OTEL path already splits on currency with `sa.case` (`sqlalchemy.py` ~L1236-1249 on main). Here is the same idea for the pre-OTEL query, on top of this PR:

```diff
@@ -1731,6 +1731,7 @@
+        currency_expr = self._json_path_expr(self.orm.Record.cost_json, "cost_currency")
         cost_expr = sa.cast(
@@ -1768,7 +1769,18 @@
                 sa.func.sum(n_tokens_expr).label("Total Tokens"),
                 latency_expr.label("Average Latency (s)"),
-                sa.func.sum(cost_expr).label("Total Cost (USD)"),
+                sa.func.sum(
+                    sa.case(
+                        (currency_expr == sa.literal("Snowflake credits"), 0.0),
+                        else_=cost_expr,
+                    )
+                ).label("Total Cost (USD)"),
+                sa.func.sum(
+                    sa.case(
+                        (currency_expr == sa.literal("Snowflake credits"), cost_expr),
+                        else_=0.0,
+                    )
+                ).label("Total Cost (Snowflake Credits)"),
@@ -1829,6 +1841,7 @@
                 "Total Cost (USD)",
+                "Total Cost (Snowflake Credits)",
             ],
@@ -1847,7 +1860,6 @@
-        base_df["Total Cost (Snowflake Credits)"] = 0.0
         base_df["tags"] = ""
```

The mixed-currency case would make a good third test in `tests/unit/test_leaderboard_pre_otel.py`, next to the existing mixed-currency test for OTEL (`test_leaderboard_mixed_currency.py`). It fails on this PR as it stands and passes with the split.

Two more notes:

- The PR also changes `avg(n_tokens)` to `sum(n_tokens)` under the "Total Tokens" label. I think that's correct, since the old label described a sum while the code took an average, but it changes a number users see, so it's worth a line in the description. The PR title also still says "Python-side aggregation".
- Limits of what I checked: SQLite only. I didn't run the Postgres `json_extract_path_text` branch or the PR's own unit tests.

— TonyDzi · the rest of the machine (second brain, agent consensus, persistent memory) lives at github.com/tonydzi

joshreini1 (approved):
Approving — this supersedes my earlier `CHANGES_REQUESTED`, which was stale. That review predated the SQL rewrite in 1407e183 and the currency split in 5f8672f3, both of which addressed it.

### Decision

We had two PRs open against #2729. **This one lands; #2793 is closed.** Reasoning, for the record:

**1. It keeps aggregation in SQL.** That was the substance of my August review. The pre-OTel path returns one row per app instead of transferring every `Record` with two JSON blobs each and grouping in pandas — O(apps) instead of O(records).

**2. It matches the OTel reference implementation.** Two behaviors that #2793 left wrong:
- `Total Tokens` is now `sum`, as the label always claimed and as the OTel path does. It was `avg`.
- Cost is split by currency via `sa.case`, so `Total Cost (Snowflake Credits)` is real rather than hardcoded `0.0`. Without this, a legacy-mode Snowflake user sees credits reported as dollars with no error — the exact thing the comment above the OTel aggregation warns about ("Mixing currencies in a single total is wrong and is what `Cost.__add__` refuses to do"). Thanks @tonydzi for catching it and @Aftabbs for turning it around.

**3. `_json_path_expr` was reusable after all.** #2793's premise was that it is "gated to real `sa.JSON`-typed columns via an `isinstance` check." It isn't — that gate lives in `_json_extract_otel`, a different method scoped to the OTel `Event` table that happens to call it. `_json_path_expr` itself takes any column and dispatches on dialect only, which is why this PR calls it directly on the Text `cost_json` and it works.

### What I pushed

One commit, acaa7186: an explicit dialect guard at the top of `_get_leaderboard_aggregates_pre_otel`.

Non-OTel mode is only supported on SQLite and Postgres — Snowflake does not support it at all, and the latency expression here needs `julianday()` or epoch extraction, which no other dialect provides. Without the guard, an unsupported backend gets SQL referencing `julianday()` and rejects it at query-compile time. I confirmed against a live Snowflake account that this surfaces as `SQL compilation error: Unknown function JULIANDAY` — a hard error either way, so no risk of wrong numbers, but it names a SQLite builtin and costs a round trip. `NotImplementedError` in Python is the right outcome for a configuration we don't support, so the guard preserves it deliberately rather than by accident.

Also collapsed the duplicate `self.engine.dialect.name` lookup, and added `test_leaderboard_unsupported_dialect_raises`. I verified that test fails with `AssertionError: NotImplementedError not raised` when the guard is removed, so it bites.

Local verification: 6 passed across `test_leaderboard_pre_otel.py`, `test_leaderboard_mixed_currency.py`, and `streamlit/test_streamlit_utils.py`; `ruff format` clean, `ruff check` clean, pre-commit clean.

One thing left as-is: the USD bucket here is `else_` on "not Snowflake credits", whereas the OTel path tests `currency == 'USD'` explicitly. Identical for the two currencies we actually emit, so not worth another round trip.

Thanks @Aftabbs for sticking with this through two rounds of review.

sfc-gh-jreini (approved):

PR #2733, opened 2026-08-26
fix(dashboard): match Topic Adherence score between results table and conversation metrics (#2725)

### LINKED ISSUE #2725: [BUG] Topic Adherence Score Mismatch Between Results Table and Conversation Metrics
**Bug Description**
The score displayed in the top-level table does not tally with the corresponding score displayed under Conversation metrics for the same thread.

In the screenshot, the table shows a score of 0.5, while the Topic Adherence metric displayed below shows 0.00 for the same conversation.

This discrepancy has only been observed when the **conversation topic is completely out of topic.**

**To Reproduce**
Run an evaluation with the Topic Adherence metric enabled using a conversation that is completely out of topic.
Open the TruLens dashboard.
Navigate to the evaluation results table.
Note the Topic Adherence score shown in the table for the thread.
Open the same thread.
Scroll to Conversation metrics and check the Topic Adherence score.
Compare the two scores.

**Expected Behavior**
The Topic Adherence score displayed in the evaluation results table should match the Topic Adherence score displayed under Conversation metrics for the same thread.

For example, if the table displays 0.5, the Conversation metrics section should also display 0.5.

**Relevant Logs/Tracebacks**

Image

**Environment:**

- OS: macOS
- Python Version: 3.10
- TruLens version: 2.12

**Additional Context**
The discrepancy has only been observed for conversations that are completely out of topic.

Abelo9996:
Closes #2725

## Problem

The Topic Adherence score shown in the top-level results table does not match the score under **Conversation metrics** for the same thread, the table shows `0.5` while the detail view shows `0.00`. As the reporter noted, this only happens when the conversation is completely off topic.

## Root cause

Topic Adherence is a **conversation-scoped** metric (`.on_conversation()`): the value describes the whole thread, but it is recorded across the thread's records. The two dashboard views select it differently:

- **Results table** - `_build_thread_summary` averages *every* feedback column across the thread's records (`ranking_group[fcol].mean(skipna=True)`), with no scope check.
- **Conversation metrics detail** - `_conversation_metric_row` uses the latest recorded value (`sort_values("ts").iloc[-1]`).

When the per-record values diverge, e.g. `[1.0, 0.0]` for a fully off-topic conversation, the table surfaces the mean (`0.5`) while the detail view shows the latest value (`0.0`). In-topic conversations tend to have consistent per-record values, so `mean ≈ latest` and the discrepancy doesn't appear, which matches the "only when completely off topic" observation.

## Fix

`_build_thread_summary` now partitions feedback columns by scope (reusing `_partition_feedback_scopes`) and, for conversation-scoped metrics, uses the latest recorded value, matching `_conversation_metric_row`. Turn-scoped metrics keep their per-turn mean.

## Tests

- `test_thread_summary_conversation_metric_matches_detail_view`, a two-turn thread with Topic Adherence `[1.0, 0.0]`; asserts the table value equals the detail-view value (`0.0`), not the mean (`0.5`). Fails on `main`, passes here.
- `test_thread_summary_turn_metric_still_averaged`, 
- 
- a turn-scoped metric still averages to `0.5`, confirming only conversation metrics change.

All existing `tests/unit/streamlit/test_streamlit_records.py` tests pass; lint/format clean under the pinned ruff.

### DIFF
--- a/src/dashboard/trulens/dashboard/tabs/Records.py
+++ b/src/dashboard/trulens/dashboard/tabs/Records.py
@@ -856,6 +856,7 @@ def _build_thread_summary(
         df["is_match"] = True
 
     rows = []
+    conversation_cols, _ = _partition_feedback_scopes(df, feedback_col_names)
     for thread_key, group in df.groupby("_thread_key", dropna=False):
         group_sorted = group.sort_values("ts")
         first = group_sorted.iloc[0]
@@ -898,7 +899,18 @@ def _build_thread_summary(
                     worst_index, "record_id"
                 ]
         for fcol in feedback_col_names:
-            if fcol in ranking_group.columns:
+            if fcol not in ranking_group.columns:
+                continue
+            if fcol in conversation_cols:
+                # Conversation-scoped metrics describe the whole thread, not a
+                # single turn, so averaging across the thread's records is wrong
+                # (e.g. records of [1.0, 0.0] would surface as 0.5 here while the
+                # conversation-metrics detail view shows 0.0). Use the latest
+                # recorded value, matching _conversation_metric_row.
+                conv_values = group_sorted[fcol].dropna()
+                if not conv_values.empty:
+                    row[fcol] = conv_values.iloc[-1]
+            else:
                 row[fcol] = ranking_group[fcol].mean(skipna=True)
         rows.append(row)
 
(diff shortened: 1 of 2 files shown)


joshreini1 (approved):
Clean fix. The root-cause analysis is spot-on: conversation-scoped metrics were averaged across per-record values in the results table but surfaced as latest-value in the detail view, creating the mismatch. The fix partitions by scope and uses the latest recorded value for conversation metrics, matching the detail-view logic. Tests cover both the regression and the unchanged turn-metric behavior. Lint/format clean, all existing tests pass. Ready to merge.

PR #2735, opened 2026-08-26
fix(database): correct dataset and ground-truth read round-trips (#2734)

### LINKED ISSUE #2734: [BUG] get_ground_truth and get_datasets raise on read (TypeError / AttributeError)
**Bug Description**

Two read methods on `SQLAlchemyDB` crash because they reference attributes that don't exist on the ORM objects:

- `get_ground_truth(ground_truth_id=...)` calls `json.loads()` on the ORM row instead of its `ground_truth_json` column → `TypeError: the JSON object must be str, bytes or bytearray, not GroundTruth`.
- `get_datasets()` reads `ds.name` / `ds.meta`, which are not columns (they live inside `dataset_json`) → `AttributeError: 'Dataset' object has no attribute 'name'`.

**To Reproduce**

```python
from trulens.core.session import TruSession
from trulens.core.schema.dataset import Dataset
from trulens.core.schema.groundtruth import GroundTruth

s = TruSession(database_url="sqlite:///bug.sqlite"); s.reset_database()
db = s.connector.db
ds_id = db.insert_dataset(Dataset(name="ds1", meta={"k": "v"}))
gid = db.insert_ground_truth(GroundTruth(query="q1", dataset_id=ds_id, expected_response="a1"))

db.get_ground_truth(gid)   # TypeError
db.get_datasets()          # AttributeError
```

**Expected Behavior**

`get_ground_truth` returns the ground-truth dict; `get_datasets` returns a DataFrame with `dataset_id`, `name`, `meta`.

**Additional Context**

Neither method had test coverage, which is why these shipped. The correct pattern already exists in the sibling `get_ground_truths_by_dataset`, which reads `json.loads(dataset.dataset_json)`.

Abelo9996:
Closes #2734

## Problem

Two read methods on `SQLAlchemyDB` crash because they reference attributes that do not exist on the ORM objects:

- `get_ground_truth(ground_truth_id=...)` calls `json.loads()` on the ORM row instead of its `ground_truth_json` column, raising `TypeError: the JSON object must be str, bytes or bytearray, not GroundTruth`.
- `get_datasets()` reads `ds.name` / `ds.meta`, which are not columns (they live inside `dataset_json`), raising `AttributeError: 'Dataset' object has no attribute 'name'`.

Both break the dataset and ground-truth read round-trip on any backend.

## Fix

Both methods now read from the JSON payload, matching the existing `get_ground_truths_by_dataset`, which already does `json.loads(dataset.dataset_json)`:

- `get_ground_truth` loads `_ground_truth.ground_truth_json`.
- `get_datasets` parses each row's `dataset_json` and reads `name` / `meta` from it.

## Tests

Neither method had any test coverage, which is why these shipped. Added regression tests in `tests/integration/test_database.py`:

- `test_get_datasets_round_trip`
- `test_get_ground_truth_round_trip`
- `test_get_ground_truth_missing_id_returns_none`

The first two fail on `main` with the exact `AttributeError` / `TypeError` above and pass with this change. No other tests regress (the pre-existing failures on a clean tree are byte-identical). Lint and format are clean under the pinned ruff.

### DIFF
--- a/tests/integration/test_database.py
+++ b/tests/integration/test_database.py
@@ -494,16 +494,7 @@ def _populate_data(db: core_db.DB):
 
 
 class TestDatasetGroundTruthRoundTrip(TestCase):
-    """Round-trip persistence for datasets and ground truths.
-
-    Regression tests for two bugs where the read path referenced attributes
-    that do not exist on the ORM objects:
-
-    - ``get_ground_truth`` called ``json.loads`` on the ORM row instead of its
-      ``ground_truth_json`` column, raising ``TypeError``.
-    - ``get_datasets`` read ``ds.name`` / ``ds.meta``, which are not columns
-      (they live inside ``dataset_json``), raising ``AttributeError``.
-    """
+    """Round-trip persistence for datasets and ground truths."""
 
     def _seed(self, db: sqlalchemy_db.SQLAlchemyDB):
         from trulens.core.schema.dataset import Dataset
@@ -528,8 +519,8 @@ class TestDatasetGroundTruthRoundTrip(TestCase):
 
             datasets = db.get_datasets()
 
-            self.assertListEqual(
-                list(datasets.columns), ["dataset_id", "name", "meta"]
+            self.assertEqual(
+                datasets.columns.tolist(), ["dataset_id", "name", "meta"]
             )
             self.assertEqual(len(datasets), 1)
             row = datasets.iloc[0]
--- a/tests/integration/test_database.py
+++ b/tests/integration/test_database.py
@@ -491,3 +491,70 @@ def _populate_data(db: core_db.DB):
         print("  ", res)
 
     return fb, app, rec
+
+
+class TestDatasetGroundTruthRoundTrip(TestCase):
+    """Round-trip persistence for datasets and ground truths.
+
+    Regression tests for two bugs where the read path referenced attributes
+    that do not exist on the ORM objects:
+
+    - ``get_ground_truth`` called ``json.loads`` on the ORM row instead of its
+      ``ground_truth_json`` column, raising ``TypeError``.
+    - ``get_datasets`` read ``ds.name`` / ``ds.meta``, which are not columns
+      (they live inside ``dataset_json``), raising ``AttributeError``.
+    """
+
+    def _seed(self, db: sqlalchemy_db.SQLAlchemyDB):
+        from trulens.core.schema.dataset import Dataset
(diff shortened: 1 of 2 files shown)


tests/integration/test_database.py
```diff
@@ -491,3 +491,70 @@ def _populate_data(db: core_db.DB):
...
+
+    Regression tests for two bugs where the read path referenced attributes
+    that do not exist on the ORM objects:
+
+    - ``get_ground_truth`` called ``json.loads`` on the ORM row instead of its
+      ``ground_truth_json`` column, raising ``TypeError``.
+    - ``get_datasets`` read ``ds.name`` / ``ds.meta``, which are not columns
+      (they live inside ``dataset_json``), raising ``AttributeError``.
+    """
+
```
joshreini1, line 498: Drop the docstring repetition in the class docstring — the two-bug explanation is already in the PR description and will become stale when this is the only reference. Keep just the first line: "Round-trip persistence for datasets and ground truths."

tests/integration/test_database.py
```diff
@@ -491,3 +491,70 @@ def _populate_data(db: core_db.DB):
...
+            self.assertEqual(row["dataset_id"], dataset_id)
+            self.assertEqual(row["name"], "support-quality")
+            self.assertEqual(row["meta"], {"source": "prod"})
+
+    def test_get_ground_truth_round_trip(self) -> None:
+        with clean_db("sqlite_file") as db:
+            db.migrate_database()
+            dataset_id, ground_truth_id = self._seed(db)
+
+            ground_truth = db.get_ground_truth(ground_truth_id)
```
joshreini1, line 536: Change `assertListEqual(list(datasets.columns), [...])` to `self.assertEqual(datasets.columns.tolist(), [...])` — pandas Series equality is more idiomatic and list() drops the Index type unnecessarily.

joshreini1 (approved):
The fix correctly addresses two read-path bugs where ORM attributes were accessed directly instead of parsing from their JSON columns. Tests now verify the round-trip behavior that was missing. The _row helper in get_datasets is clear and the ground_truth_json access matches the existing pattern in get_ground_truths_by_dataset. Ready to merge.

Abelo9996:
Thanks for the review. Both addressed in the follow-up commit:

- Trimmed the class docstring to just the summary line, so the two-bug explanation lives only in the PR and issue.
- Switched to `self.assertEqual(datasets.columns.tolist(), [...])`.

PR #2736, opened 2026-08-27
feat(examples): portable semantic trace search and failure clustering cookbook (#2704)

### LINKED ISSUE #2704: Portable semantic trace search and failure clustering cookbook
## Problem

The Records page supports literal substring search, but developers often know a failure by meaning rather than exact text. They also need to understand whether low-quality traces are isolated incidents or recurring semantic failure modes. TruLens exposes the required record, span, and evaluation data, but lacks an end-to-end local example using standard text-analysis tools.

## Proposal

Add one credential-free cookbook notebook plus a tested helper module. The default path uses checked-in synthetic trace fixtures and can be switched to real SQLite/PostgreSQL data through public TruLens APIs.

### Data loading

- Default: synthetic JSONL containing records, span paths, evaluation scores/explanations, errors, models, tools, latency, and known failure groups.
- Real records: `TruSession.get_records_and_feedback()`.
- Raw OTEL spans: `TruSession.get_events()` when available.
- Do not use underscore-prefixed database or connector methods.

Normalize three DataFrames:

- `records`: record ID, app/version, conversation, timestamp, input, output, error, latency, cost, tokens, and metric columns.
- `spans`: record/trace/span IDs, parent, type, name, duration, status, and selected semantic attributes.
- `evaluations`: record ID, metric, score, direction, explanation, status, and evaluation cost.

### Failure documents

Build one deterministic document per failed or low-scoring record:

```text
[INPUT] ...
[OUTPUT] ...
[ERROR] ...
[EVALUATION] metric=score; explanation=...
[PATH] retrieval > reranking > generation > tool
```

Keep record IDs, numeric scores, costs, timestamps, app versions, models, and tools as metadata rather than embedding them. Use field allowlists and deterministic truncation for contexts, errors, and repeated spans. Apply configured masking before vectorization and state clearly that secret redaction is not general PII detection.

### Semantic search baseline

- `TfidfVectorizer` with unigrams and bigrams.
- `TruncatedSVD` latent semantic representation.
- L2 normalization and exact cosine similarity.
- A small labeled query set with expected relevant failure-group IDs.
- Report recall@k or success@k rather than only attractive examples.

### Failure clustering

- `MiniBatchKMeans` over normalized SVD vectors with fixed seeds.
- Evaluate a small candidate range for `k` using silhouette and repeated-seed adjusted Rand stability.
- Report cluster sizes, medoid records, top terms, affected app versions, low metrics, errors, tools, and representative examples.
- Use two-dimensional PCA only for a Plotly navigation chart, not as clustering input.

### TruLens Hotspots comparison

- Run `trulens-hotspots` against one evaluation score.
- Explain that semantic clusters group similar failures while Hotspots finds interpretable features correlated with worse scores.
- Show both outputs on the same synthetic fixture.

### Optional local embeddings

Include a skipped-by-default section replacing TF-IDF/SVD with one configurable local sentence-transformer. Document model download, license, embedding dimension, and privacy. Do not add sentence-transformers, FAISS, HNSW, UMAP, or HDBSCAN to core dependencies.

All visualizations remain pandas/Plotly notebook outputs. Do not include a Streamlit app.

## Non-goals

- Changes to built-in Records search or dashboard.
- A persistent vector index or search service.
- Production-scale nearest-neighbor infrastructure.
- Automatic cluster labels from an external LLM.
- Treating unsupervised clusters as ground truth.
- Snowflake-specific processing.

## Acceptance criteria

- The default notebook executes offline with fixed seeds and synthetic data.
- Real-data loading uses only public TruLens APIs.
- Failure-document construction is deterministic and unit tested.
- Masking occurs before vectorization.
- The labeled search fixture reports an explicit top-k quality measure.
- Cluster selection reports silhouette, stability, and size distribution.
- Every cluster includes a medoid and interpretable summary.
- IDs, scores, and costs are not inserted into failure text.
- Hotspots and semantic clustering are clearly distinguished.
- Optional embedding-model code is isolated and not run in default tests.
- No Streamlit dependency or app is introduced.

## Tests

- Fixture normalization for record and OTEL shapes.
- Failure selection for higher-is-better and lower-is-better metrics.
- Document construction, truncation, and masking golden tests.
- Search ranking against labeled fixture queries.
- Fixed-seed clustering and medoid selection.
- Stability and top-term helpers.
- Notebook output stripping and credential-free execution.

## Critical files

- `examples/expositional/use_cases/semantic_trace_analysis.ipynb`
- `examples/expositional/use_cases/semantic_trace_analysis.py`
- `examples/expositional/use_cases/data/semantic_trace_fixture.jsonl`
- `docs/cookbook/use_cases/semantic_trace_analysis.ipynb`
- `tests/unit/test_semantic_trace_analysis.py`

Payal2000:
Closes #2704

## Problem

The Records page searches traces by literal substring, but developers usually know a failure by its *meaning* — "the assistant invented a citation", "the billing tool timed out". And once one turns up, the next question is whether it is an isolated incident or a recurring semantic failure mode. TruLens already exposes the record, span and evaluation data needed to answer both; what was missing was an end-to-end local example.

## What this adds

A credential-free cookbook notebook plus a tested helper module. Everything runs locally on standard text-analysis tools: TF-IDF over unigrams and bigrams, a truncated SVD latent space, exact cosine similarity, and MiniBatchKMeans with fixed seeds. **No vector service, no index to maintain, no network call, and no new core dependency.**

### Data loading

`load_fixture()` reads checked-in synthetic traces — 32 records, 70 spans, 64 evaluations across four planted failure groups plus a healthy control — so the default path is reproducible offline. `load_from_session()` reads real data through `get_records_and_feedback()` and, when the backend supports it, `get_events()`. No underscore-prefixed database or connector method is touched. Both produce the same three normalized frames (`records`, `spans`, `evaluations`), so switching between them changes exactly one cell.

### Failure documents

Failure selection follows each metric's own direction, so a higher-is-better groundedness score and a lower-is-better latency penalty can share one call.

Documents are deterministic, field-allowlisted, and truncated with a visible marker. Two decisions worth calling out:

- **Identifiers and numbers stay out of the text.** Record ids, scores, costs, timestamps, versions, models and tools travel as metadata. Embedding a record id lets it influence similarity, and "similar failure" quietly becomes "similar id". There is a test asserting none of them appear in any document.
- **Masking runs before vectorization**, not before display, so a leaked credential never reaches the vocabulary. A test checks the secret is absent from the fitted vocabulary, not just from the rendered text. This is documented as secret redaction for well-known credential shapes, explicitly *not* general PII detection.

### Search and clustering

Search quality is reported as success@k and recall@k against a labeled query set, rather than a few flattering examples — so a change that makes retrieval worse is visible as a number.

Cluster selection reports silhouette **alongside** repeated-seed adjusted Rand stability and the size spread. Silhouette alone will happily recommend a `k` that reshuffles whenever the seed changes; stability catches that, and a `k` whose smallest cluster holds one member is describing an outlier rather than a failure mode. Every cluster reports a medoid — a real record nearest the centre, not a synthetic centroid — with top terms, affected app versions, models, tools, worst metrics and examples.

Two-dimensional PCA is used only to draw the Plotly navigation chart, never as clustering input.

### Hotspots comparison and optional embeddings

The notebook runs `trulens-hotspots` on the same fixture and explains the distinction: a cluster says "these twelve failures are the same story", a hotspot says "records using `billing_api` score 0.4 lower".

The local sentence-transformer section is skipped by default behind a `RUN_LOCAL_EMBEDDINGS` flag and documents the model's size, license, embedding dimension and privacy properties. `sentence-transformers`, FAISS, HNSW, UMAP and HDBSCAN are not added anywhere.

## Tests

73 unit tests: normalization for both fixture and OTEL event shapes, failure selection in both metric directions, golden document construction, truncation and masking, search ranking against the labeled fixture, fixed-seed clustering, medoid and top-term helpers, and that the notebook stays output-stripped and credential-free.

The pipeline is checked against the structure the fixture actually plants, not just for absence of exceptions: `select_k` picks k=4 on silhouette with 0.83 stability, clustering recovers the four planted groups at adjusted Rand > 0.5, and all four labeled queries reach success@5 with the top result in the correct group.

The notebook itself was executed end to end with nbclient and runs clean with no credentials. Verified locally: the non-OTEL unit suite goes from 329 to 402 passing with no new failures, and the remaining failures are byte-identical on a clean tree.

## Notes for review

- **The test file skips where scikit-learn is absent.** It is guarded with `pytest.importorskip` because sklearn is not a core TruLens dependency, so it *skips* rather than fails in an environment without it. Worth confirming CI's test environment actually has sklearn before treating a green run as coverage here.
- `docs/cookbook` is a symlink to `examples/expositional`, so the two notebook paths named in the issue are the same file and only one notebook was added.
- The helpers live next to the notebook rather than inside the package, so the cookbook stays self-contained and copyable. That is why the test file puts the directory on `sys.path`.
- One bug the fixture caught during development, kept as a named regression test: pandas stores an absent value as `NaN`, and `bool(float("nan"))` is `True`, so a plain truthiness check on the error column selected every record as errored.


sfc-gh-jreini:
Hi @Payal2000 - thanks for contributing. @dylanpulver already tackled this issue in this PR: https://github.com/truera/trulens/pull/2716

If there are places you would extend his notebook, can you add them there in a new PR? Closing this one. Thanks

PR #2737, opened 2026-08-27
fix(benchmark): score ground-truth rows in row order, not index-label order

arthi-arumugam-git:
## What changed

`TruBenchmarkExperiment.__call__` keys its futures by the frame's index label, taken from `iterrows()`, then reads the results back over `range(len(ground_truth))`. Those two only agree when the labels happen to be `0..n-1`.

Its own docstring requires the order to hold:

> Note the order of generated scores must be preserved to match the order of the true labels.

`GroundTruthAggregator` zips those scores against `true_labels` positionally, so a drop or a reorder is scored against the wrong row, and nothing raises.

## What goes wrong

A frame whose labels are not `0..n-1`, which is what a filter, a `set_index` or a `concat` leaves behind, breaks it three ways:

| ground truth | index labels | scores returned for 4 rows |
|---|---|---|
| after a filter | `[1, 2, 3, 5]` | **3 scores**, label 5 falls outside the range |
| a string index | `[a, b, c, d]` | **0 scores**, the list comes back empty |
| after a `concat` | `[0, 1, 0, 1]` | **4 scores in the wrong order** |

The `concat` case is the one I would worry about. The length is right, so nothing downstream can notice, and because the bucket is filled by `as_completed` the order varies between runs. In that case `true_labels[1]` is compared against the score for row 2.

## The change

Futures are keyed by position instead. The reassembly range then always lines up, and a row carrying several `expected_chunks` still contributes all of its scores, which is why the loop still walks a list per key rather than taking one.

## Tests

`tests/unit/test_benchmark_experiment_ordering.py`, four parametrised cases over the index shapes above.

Three of them fail on the previous reassembly:

```
assert ['q1', 'q2', 'q3']         == ['q1', 'q2', 'q3', 'q1']
assert []                          == ['q0', 'q1', 'q2', 'q3']
assert ['q0', 'q2', 'q1', 'q3']    == ['q0', 'q1', 'q2', 'q3']
```

The default-index case passes either way, so a frame that was already fine is unaffected.

The module had no test referencing it before this, which is how the three shapes went unnoticed.

### DIFF
--- a/src/benchmark/trulens/benchmark/benchmark_frameworks/tru_benchmark_experiment.py
+++ b/src/benchmark/trulens/benchmark/benchmark_frameworks/tru_benchmark_experiment.py
@@ -187,7 +187,13 @@ class TruBenchmarkExperiment:
             future_to_index = {}
             index_to_results: Dict[int, List] = {}
 
-            for index, row in ground_truth.iterrows():
+            # Key by POSITION, not by the frame's index label. iterrows()
+            # yields labels, and a frame left behind by a filter, a set_index
+            # or a concat does not have labels 0..n-1. Reassembling such a
+            # frame positionally drops rows whose label falls outside the
+            # range and reorders rows that share a label, both silently, while
+            # the docstring above requires this order to match true_labels.
+            for position, (_, row) in enumerate(ground_truth.iterrows()):
                 if "expected_chunks" in row:
                     for expected_chunk in row["expected_chunks"]:
                         future = executor.submit(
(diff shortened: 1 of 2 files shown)


src/benchmark/trulens/benchmark/benchmark_frameworks/tru_benchmark_experiment.py
```diff
@@ -187,36 +187,43 @@ def __call__(
             future_to_index = {}
             index_to_results: Dict[int, List] = {}
 
-            for index, row in ground_truth.iterrows():
+            # Key by POSITION, not by the frame's index label. iterrows()
```
joshreini1, line 190: The unpacked underscore from enumerate(ground_truth.iterrows()) is unused but harmless. This works fine; if you want to tighten it consider itertuples(index=False) instead of the nested unpack, but the current approach is consistent with the original iterrows() API surface.

joshreini1 (approved):
Solid fix for a silent correctness bug. The reassembly now keys by position instead of index label, so filtered/reindexed frames no longer drop rows or reorder scores. The test suite covers all three failure modes (truncation, empty, reorder) plus the happy path. The comment additions explain the why clearly. Ready to merge.

PR #2738, opened 2026-08-27
fix(groundtruth): report nan, not 0.0, when an IR metric is undefined

### LINKED ISSUE #2731: GroundTruthAggregator reports an undefined metric as 0.0 in one place and nan in another
**Bug Description**

`GroundTruthAggregator` answers "the metric is undefined here" two different ways, and both answers are pinned by tests.

`brier_score` returns `np.nan`, and the comment explains why (`groundtruth.py:1238`):

```python
# Brier score is mathematically undefined for empty inputs (division by zero).
# Return np.nan to accurately represent this undefined state rather than 0.0
# which would falsely suggest "perfect calibration" with no predictions.
if not scores:
    return np.nan
```

`recall` and `precision` return `0.0` for what looks like the same situation (`groundtruth.py:1164` and `1201`):

```python
# Handle the case where there are no actual positives to avoid division by zero
if true_positives + false_negatives == 0:
    return 0.0  # or handle as needed (e.g., return None, raise an exception)
```

That trailing comment appears three times, on `recall`, `precision` and `f1_score`, so I do not think this was ever settled.

**To Reproduce**

```python
from trulens.feedback.groundtruth import GroundTruthAggregator

no_positives = GroundTruthAggregator(true_labels=[0, 0, 0, 0])
missed_all   = GroundTruthAggregator(true_labels=[1, 1, 0, 0])

no_positives.recall([0.9, 0.8, 0.7, 0.6])   # 0.0, but there were no positives to recall
missed_all.recall([0.1, 0.1, 0.1, 0.1])     # 0.0, and this one is a real measured zero

no_positives.brier_score([])                # nan
```

**Expected Behavior**

I am not sure, which is why this is a question rather than a PR.

The two `recall` calls above are different events. One is a model that missed every positive. The other is a dataset with no positives in it. Once these are averaged across an evaluation they are the same number, so a run cannot tell a bad model from a degenerate slice. That is the reasoning the `brier_score` comment already gives.

Against that, `tests/unit/test_groundtruth_aggregator.py:1162` (`test_zero_division_handling`) asserts the current behaviour on purpose, and says so:

- "Precision should be 0.0 when no positives are predicted"
- "Recall should be 0.0 when no actual positives exist"

So `0.0` looks deliberate here and `nan` looks deliberate five methods away, and I cannot tell from the outside which one you want to keep.

**Additional Context**

If `nan` is the convention, the change is small. `recall` and `precision` return `np.nan` in the undefined branch, and `f1_score` needs nothing at all: `nan` already falls through its `precision + recall == 0` guard and propagates, while a genuine `0.0 + 0.0` still returns `0.0`. I have this working locally with tests, and `test_zero_division_handling` would need updating, which is really the decision being asked about.

The same split appears elsewhere in the file, so whichever way it goes it probably wants to be applied in one pass: `ndcg_at_k:388`, `precision_at_k:446`, `recall_at_k:505`, `ir_hit_rate:588`, `memory_recall:880,884`, `memory_mrr:929,933,941` all return `0.0` where the neighbouring branch returns `np.nan`.

Happy to open the PR if you tell me which convention you would like.

**Environment**

- Read against `main` at `bf2180b`
- Python 3.13, numpy 2.4.6

arthi-arumugam-git:
Closes #2731.

`brier_score` in this class already states the rule, and the reason:

```python
# Return np.nan to accurately represent this undefined state rather than 0.0
# which would falsely suggest "perfect calibration" with no predictions.
```

Nine sibling branches did the opposite. Thanks to @Abelo9996 for the read on the issue, including the point that `f1_score` needs no guard of its own.

## Changed to `np.nan`, all cases with no denominator

| method | undefined when |
|---|---|
| `ndcg_at_k` | `ideal_dcg == 0`, the golden set annotates nothing relevant |
| `precision_at_k` | nothing was retrieved |
| `recall_at_k` | no relevant chunks exist |
| `mrr` | new guard, the golden set is empty |
| `ir_hit_rate` | new guard, the golden set is empty |
| `memory_recall` | the expected set is empty |
| `memory_mrr` | the expected set is empty |
| `recall` | no positives in the ground truth |
| `precision` | nothing was predicted positive |

## Deliberately left as `0.0`, because these are measured

- `mrr` and `ir_hit_rate` when relevant items exist and none were retrieved. That is a real miss.
- `memory_mrr` when nothing was retrieved, or nothing retrieved matched.
- `f1_score`, which needs no guard: `nan` falls through its `precision + recall == 0` check and propagates, while a real `0.0 + 0.0` still returns `0.0`.

The distinction that matters: a slice with no positives and a model that missed every positive both returned `0.0` before, and averaged into the same number.

## `test_zero_division_handling`

That test asserted the old `0.0` on purpose, so it is updated rather than deleted, and the reason is written into its docstring. The measured zeros are asserted alongside, so the boundary is pinned in both directions rather than moved.

22 tests pass in `tests/unit/test_groundtruth_aggregator.py`. `ruff check` and `ruff format` clean.


tests/unit/test_groundtruth_aggregator.py
```diff
@@ -1160,32 +1160,45 @@ def test_score_format_consistency(self):
 
     @pytest.mark.optional
     def test_zero_division_handling(self):
```
joshreini1, line 1162: Consider renaming this test method to test_undefined_vs_measured_zero to match the new docstring's framing — the name 'zero_division_handling' now understates what it actually verifies (the semantic split between nan and 0.0). Non-blocking.

joshreini1 (approved):
Clear, well-reasoned change. The PR correctly distinguishes undefined metrics (no denominator) from measured zeros (missed every positive), fixing a real aggregation problem where degenerate slices and bad models were indistinguishable. The brier_score precedent is solid, the test update is honest about the intent change, and the measured-zero assertions pin the boundary. Ready to merge.

arthi-arumugam-git:
CI was failing on `tests/unit/test_groundtruth_agreement.py::test_empty_retrieved_chunks`, pushed a fix in e9bd3cb.

That test asserted `0.0` for all four IR metrics when nothing is retrieved, and this PR changes one of them. Precision divides by the number of retrieved chunks, so with none retrieved it has no denominator and now reports nan. The assertion loop hit that first and stopped there.

Only precision moves. `recall_at_k`, `mrr` and `ir_hit_rate` still have a denominator in that case, because the query does have golden chunks and none of them were found, which is a measured zero. They are asserted alongside now so the split is pinned in both directions, the same way `test_zero_division_handling` was updated in the aggregator tests.

Verified locally with `--run_optional_tests`: the file goes to 53 passed, and reverting just this commit reproduces the original `nan != 0.0` at the same line.

PR #2740, opened 2026-08-29
feat(providers): add trulens-providers-orcarouter package

bangla24bdrang-lab:
## What

TruLens exposes each feedback provider as an installable `trulens-providers-*` package, and its OpenAI provider already doubles as a generic gateway client — the docs mention cloud gateways like OpenRouter reachable via `base_url` forwarding. This PR follows the same playbook used by `trulens-providers-ollama` and adds a **named** [OrcaRouter](https://www.orcarouter.ai) provider, so TruLens users can point feedback functions at OrcaRouter without treating it as an anonymous custom base URL.

[OrcaRouter](https://www.orcarouter.ai) is an OpenAI-compatible AI gateway built for both models and agents. Like OpenRouter, it exposes a provider/model namespace across many models — but it also combines adaptive routing, automatic failover, zero-markup inference, observability, guardrails, and agent-tool governance behind the same endpoint. Adding orcarouter as a first-class provider means this project's users can use that stack directly. It also runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes.

## Changes

- **New `trulens-providers-orcarouter` package** with an `OrcaRouter` provider that subclasses `trulens.providers.openai.OpenAI` (mirroring how `Ollama` and `AzureOpenAI` reuse the OpenAI implementation). Since OrcaRouter is API-compatible with OpenAI, capability probing, structured outputs and the Responses API keep working unchanged.
- **Wired into the build/docs**: registered in the `providers` dependency group in `pyproject.toml`, listed in the mkdocs API nav, documented in `README.md` (install snippet + Supported LLM Providers table) and in the feedback providers guide.
- **Default wiring**: reads `ORCAROUTER_API_KEY` / `ORCAROUTER_BASE_URL` (default `https://api.orcarouter.ai/v1`) and uses provider-prefixed model IDs such as `openai/gpt-4o-mini`.
- **Tests**: `tests/unit/providers/test_orcarouter_capabilities.py` covering construction, env-var handling and API-key validation. Also added the new module to the optional-modules list in `tests/unit/static/test_static.py`.

## Verification

- `ruff check` and `ruff format --check` pass on all touched files.
- `poetry check --lock` passes after regenerating `poetry.lock` with the new package.
- `pytest tests/unit/providers/test_orcarouter_capabilities.py --run_optional_tests` → 5 passed.
- **Live check**: constructed `OrcaRouter()` and ran `generate_score`, `relevance` and `relevance_with_cot_reasons` against `https://api.orcarouter.ai/v1` (all returned valid scores via real chat completions). OrcaRouter does not currently expose OpenAI's moderation endpoint, which is documented as a warning on the provider, consistent with how the Ollama provider documents its moderation limitation.

I'm an engineer on the OrcaRouter team. Discord: discord.gg/YEubt8enRA · X: https://x.com/OrcaRouter

### DIFF
--- a/src/providers/orcarouter/trulens/providers/orcarouter/provider.py
+++ b/src/providers/orcarouter/trulens/providers/orcarouter/provider.py
@@ -6,6 +6,8 @@ from trulens.providers.openai import provider as openai_provider
 
 logger = logging.getLogger(__name__)
 
+_DEFAULT_BASE_URL = "https://api.orcarouter.ai/v1"
+
 
 class OrcaRouter(openai_provider.OpenAI):
     """Out of the box feedback functions calling models through the
@@ -68,9 +70,6 @@ class OrcaRouter(openai_provider.OpenAI):
 
     DEFAULT_MODEL_ENGINE: ClassVar[str] = "openai/gpt-4o-mini"
 
-    DEFAULT_BASE_URL: ClassVar[str] = "https://api.orcarouter.ai/v1"
-    """Default OrcaRouter OpenAI-compatible base URL."""
-
     def __init__(
         self,
         *args,
@@ -83,9 +82,7 @@ class OrcaRouter(openai_provider.OpenAI):
             model_engine = self.DEFAULT_MODEL_ENGINE
 
         if base_url is None:
-            base_url = os.environ.get(
-                "ORCAROUTER_BASE_URL", self.DEFAULT_BASE_URL
-            )
+            base_url = os.environ.get("ORCAROUTER_BASE_URL", _DEFAULT_BASE_URL)
 
         if api_key is None:
             api_key = os.environ.get("ORCAROUTER_API_KEY")
--- a/src/providers/orcarouter/trulens/providers/orcarouter/provider.py
+++ b/src/providers/orcarouter/trulens/providers/orcarouter/provider.py
@@ -1,0 +1,104 @@
+import logging
+import os
+from typing import ClassVar, Optional
+
+from trulens.providers.openai import provider as openai_provider
+
+logger = logging.getLogger(__name__)
+
+
+class OrcaRouter(openai_provider.OpenAI):
+    """Out of the box feedback functions calling models through the
+    [OrcaRouter](https://www.orcarouter.ai) gateway.
+
+    OrcaRouter is an OpenAI-compatible gateway that exposes a provider-prefixed
+    model namespace (e.g. `openai/gpt-4o-mini`, `anthropic/claude-sonnet-4.6`,
+    `google/gemini-2.5-pro`, `orcarouter/auto`) behind a single base URL, so a
+    single feedback provider can evaluate against many models. Other
+    OpenAI-compatible gateways are reachable through the OpenAI provider's
+    `base_url` forwarding; this provider instead wires the gateway's default
+    endpoint and API key (`ORCAROUTER_API_KEY`) directly, so no plumbing is
+    needed.
+
+    Because OrcaRouter is API-compatible with OpenAI, the OpenAI
+    implementation is reused as-is: capability probing, structured outputs and
+    the Responses API all keep working.
+
+    !!! warning
+        _OrcaRouter_ does not currently expose the _OpenAI_ moderation
+        endpoint (e.g. `text-moderation-stable`), so the `moderation_*`
+        feedback functions will not work through this provider.
+
+    Create an OrcaRouter Provider with out of the box feedback functions.
+
+    Example:
+        ```python
+        from trulens.providers.orcarouter import OrcaRouter
+
+        # Uses ORCAROUTER_API_KEY and the default model openai/gpt-4o-mini.
+        provider = OrcaRouter()
+        ```
+
+        Pick a different model from the gateway catalog by its provider-prefixed
+        ID:
+
+        ```python
+        provider = OrcaRouter(model_engine="anthropic/claude-sonnet-4.6")
+        ```
+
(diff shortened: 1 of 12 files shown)


tests/unit/providers/test_orcarouter_capabilities.py
```diff
@@ -0,0 +1,83 @@
...
+    )
+    from trulens.providers.orcarouter import (
+        OrcaRouter,  # type: ignore[import-not-found]
+    )
+
+    provider = OrcaRouter(model_engine="openai/gpt-4o-mini")
+
+    # Structured-output and moderation support are inherited unchanged from
+    # the OpenAI provider, since OrcaRouter is API-compatible with OpenAI.
+    assert isinstance(provider, OpenAI)
```
joshreini1, line 99: Remove this assertion or verify it against the actual OrcaRouter API. The test assumes `_structured_output_supported()` returns `True` based solely on the fact that OrcaRouter subclasses `OpenAI`, but structured outputs (response_format with Pydantic schemas) depend on the gateway actually supporting that OpenAI feature. The Ollama provider overrides `_structured_output_supported` because Ollama handles schemas differently; if OrcaRouter doesn't support OpenAI's structured-output API surface identically, this will silently fail at runtime.

tests/unit/providers/test_orcarouter_capabilities.py
```diff
@@ -0,0 +1,83 @@
...
+    assert provider._structured_output_supported() is True
```
joshreini1: Add at least one test that exercises a real feedback function (e.g. `provider.relevance("query", "response")`) with a mocked chat-completion response, confirming the provider's `_create_chat_completion` path works end-to-end. The existing tests only verify constructor wiring and inheritance — none confirm that feedback evaluation actually runs without errors or that the default model ID is accepted by the gateway.

src/providers/orcarouter/trulens/providers/orcarouter/provider.py
```diff
@@ -0,0 +1,104 @@
...
+        **kwargs: Additional arguments to pass to the
+            [OpenAIEndpoint][trulens.providers.openai.endpoint.OpenAIEndpoint]
+            which are then passed to
+            [OpenAIClient][trulens.providers.openai.endpoint.OpenAIClient]
+            and finally to the OpenAI client (for example `api_key`,
+            `base_url`).
+    """
+
+    DEFAULT_MODEL_ENGINE: ClassVar[str] = "openai/gpt-4o-mini"
+
```
joshreini1, line 72: Change `DEFAULT_BASE_URL: ClassVar[str]` to a module-level constant `_DEFAULT_BASE_URL` (leading underscore) or inline it directly in the `__init__` default logic, matching how `Ollama.DEFAULT_BASE_URL` is structured. The ClassVar pattern is fine but inconsistent with the coding standards in AGENTS.md (80-char line length, simpler is better). Inlining saves one layer of indirection.

joshreini1 (changes requested):
Clean implementation following the Ollama provider pattern — the OrcaRouter provider correctly subclasses OpenAI, wires env vars, and documents the moderation limitation. Two blockers: (1) the test assertion at line 82 expects `_structured_output_supported()` to return `True`, but that assumes OrcaRouter's API supports OpenAI's structured output features without verification; (2) no test actually exercises a feedback function end-to-end (e.g. `relevance`, `groundedness`) to confirm the provider works against a real or mocked gateway. Request changes: verify structured-output claim and add a functional smoke test.

tests/unit/providers/test_orcarouter_capabilities.py
bangla24bdrang-lab: Removed the assertion. Whether the gateway honors structured output is decided at runtime by the OpenAI capability-fallback logic: `_call_with_capability_fallbacks` tries `responses.parse`, and on failure marks `structured_outputs=False` in the per-model-id cache and falls back to `chat.completions`. The capability test now only asserts the inheritance relationship (`isinstance(provider, OpenAI)`); the new smoke test (below) covers the runtime fallback path with a mocked gateway.

tests/unit/providers/test_orcarouter_capabilities.py
bangla24bdrang-lab: Added `test_relevance_smoke_test`, which runs `OrcaRouter().relevance("what is a cow?", "A cow is an animal.")` against a mocked client. The dummy `responses.parse` raises (structured outputs unsupported), so the OpenAI fallback routes to `chat.completions.create`, which returns `{"score": 3, "reason": "matches"}`. The test asserts the normalized score is `1.0` (3/3) and that the default model id `openai/gpt-4o-mini` was forwarded to the gateway. This exercises the full `relevance` -> `generate_score` -> `_create_chat_completion` path.

src/providers/orcarouter/trulens/providers/orcarouter/provider.py
bangla24bdrang-lab: Done. `DEFAULT_BASE_URL: ClassVar[str]` is replaced with a module-level `_DEFAULT_BASE_URL` constant, and `__init__` now reads `os.environ.get("ORCAROUTER_BASE_URL", _DEFAULT_BASE_URL)`. Confirmed the changed files pass ruff (check + format) and the 80-char line limit.

sfc-gh-jreini (approved):

joshreini1 (approved):

PR #2743, opened 2026-08-29
fix(database): persist feedback definition updates on re-insert (#2742)

### LINKED ISSUE #2742: [BUG] insert_feedback_definition silently drops updates (writes non-existent app_json column)
**Bug Description**

`SQLAlchemyDB.insert_feedback_definition` is an upsert (insert or update). Its update branch assigns the serialized definition to `_fb_def.app_json`, but `FeedbackDefinition` has no `app_json` column. Its JSON column is `feedback_json` (the ORM defines only `feedback_definition_id`, `run_location`, `feedback_json`). SQLAlchemy accepts the stray instance attribute, so `feedback_json` is never updated and re-inserting an existing id silently drops the change. No exception is raised; the row just keeps its old value.

**To Reproduce**

```python
from trulens.core.session import TruSession
from trulens.core.schema.feedback import FeedbackDefinition

s = TruSession(database_url="sqlite:///bug.sqlite"); s.reset_database()
db = s.connector.db
db.insert_feedback_definition(FeedbackDefinition(feedback_definition_id="fd1", supplied_name="ORIGINAL"))
db.insert_feedback_definition(FeedbackDefinition(feedback_definition_id="fd1", supplied_name="UPDATED"))

print(db.get_feedback_defs(feedback_definition_id="fd1").iloc[0]["feedback_json"]["supplied_name"])
# prints "ORIGINAL", expected "UPDATED"
```

**Expected Behavior**

Re-inserting a feedback definition with an existing id updates the stored `feedback_json`.

**Additional Context**

The update path had no test coverage. The insert branch and every sibling upsert (`insert_ground_truth`, `insert_dataset`, `insert_app`) write to the correct column via the ORM `parse` path, which also handles key redaction; only this branch was wrong.

Abelo9996:
Closes #2742

## Problem

`SQLAlchemyDB.insert_feedback_definition` is an upsert. Its update branch assigned the serialized definition to `_fb_def.app_json`, which is not a column on the `FeedbackDefinition` ORM (its JSON column is `feedback_json`). SQLAlchemy accepts the stray instance attribute, so `feedback_json` is never written and re-inserting an existing id silently drops the change. No exception is raised.

## Fix

Serialize through the ORM `parse` path, matching the insert branch and every sibling upsert, which also applies key redaction:

```python
_fb_def.feedback_json = self.orm.FeedbackDefinition.parse(
    feedback_definition, redact_keys=self.redact_keys
).feedback_json
```

## Tests

Adds `TestFeedbackDefinitionUpsert.test_reinsert_updates_definition` in `tests/integration/test_database.py`: it inserts a definition, re-inserts the same id with a changed `supplied_name`, and asserts the stored value updated. It fails on `main` (reads back the original value) and passes with this change. Lint and format clean under the pinned ruff.

### DIFF
--- a/src/core/trulens/core/database/sqlalchemy.py
+++ b/src/core/trulens/core/database/sqlalchemy.py
@@ -598,7 +598,9 @@ class SQLAlchemyDB(core_db.DB):
                 )
                 .first()
             ):
-                _fb_def.app_json = feedback_definition.model_dump_json()
+                _fb_def.feedback_json = self.orm.FeedbackDefinition.parse(
+                    feedback_definition, redact_keys=self.redact_keys
+                ).feedback_json
             else:
                 _fb_def = self.orm.FeedbackDefinition.parse(
                     feedback_definition, redact_keys=self.redact_keys
--- a/tests/integration/test_database.py
+++ b/tests/integration/test_database.py
@@ -549,3 +549,36 @@ class TestDatasetGroundTruthRoundTrip(TestCase):
         with clean_db("sqlite_file") as db:
             db.migrate_database()
             self.assertIsNone(db.get_ground_truth("does-not-exist"))
+
+
+class TestFeedbackDefinitionUpsert(TestCase):
+    """Regression test for a silent update loss.
+
+    ``insert_feedback_definition`` is an upsert. Its update branch assigned the
+    serialized definition to ``app_json``, which is not a column on the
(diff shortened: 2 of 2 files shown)


joshreini1 (approved):
Clean fix for a silent data-loss bug. The update branch now serializes through the ORM parse path to `feedback_json` (the actual column) instead of `app_json` (a stray attribute SQLAlchemy accepted but never persisted). The regression test directly exercises the upsert scenario and would catch this on `main`. Ready to merge.

PR #2745, opened 2026-08-29
fix(run): compute_metrics message builds from metric names, not raw objects (#2744)

### LINKED ISSUE #2744: [BUG] compute_metrics raises TypeError instead of messaging when a metric is passed as an object
**Bug Description**

`Run.compute_metrics(metrics: List[Union[str, Metric, MetricConfig]])` accepts metrics as strings or as metric objects. When an already-computed (or in-progress) metric is passed as a `Metric`/`MetricConfig` object, the skip-path message does `", ".join(computed_metrics)` over the raw objects and raises `TypeError: sequence item 0: expected str instance, Metric found`, instead of returning the intended "already computed" message.

`_should_skip_computation` already reads `.name` from non-string metrics, so the same objects reach the join; only the message builder assumes strings.

**To Reproduce**

Call `compute_metrics([metric_obj])` where `metric_obj` is a `Metric` that is already computed for the run. The join raises `TypeError` rather than returning a message. Minimal proof that the join itself is the fault:

```python
from trulens.core.metric.metric import Metric
", ".join([Metric(name="relevance")])   # TypeError: expected str instance, Metric found
```

**Expected Behavior**

`compute_metrics` returns the "already computed or in progress" message listing the metric names, whether metrics were passed as strings or objects.

**Additional Context**

The skip-path message had no test coverage for the object case.

Abelo9996:
Closes #2744

## Problem

`Run.compute_metrics` accepts metrics as strings or as `Metric`/`MetricConfig` objects. When an already-computed or in-progress metric is passed as an object, the skip-path message does `", ".join(computed_metrics)` over the raw objects and raises `TypeError: sequence item 0: expected str instance, Metric found`, instead of returning the "already computed" message.

`_should_skip_computation` already reads `.name` from non-string metrics, so any object that reaches `computed_metrics` has a `.name`; only the message builder assumed strings.

## Fix

Join on the metric name, mirroring `_should_skip_computation`:

```python
", ".join(m if isinstance(m, str) else m.name for m in computed_metrics)
```

## Tests

Adds `tests/unit/test_run_compute_metrics.py` covering both the object and string cases. The object case fails on `main` with the `TypeError` above and passes with this change. Lint and format clean under the pinned ruff.

### DIFF
--- a/tests/unit/test_run_compute_metrics.py
+++ b/tests/unit/test_run_compute_metrics.py
@@ -73,12 +73,14 @@ class TestComputeMetricsMessage(unittest.TestCase):
         """Metric object passed to compute_metrics yields message when already
         computed (no TypeError)."""
         metric = _make_metric("my_metric")
-        with patch.object(
-            Run, "describe", return_value=self.metadata
-        ), patch.object(
-            Run, "get_status", return_value="INVOCATION_COMPLETED"
-        ), patch.object(
-            Run, "_can_start_new_metric_computation", return_value=True
+        with (
+            patch.object(Run, "describe", return_value=self.metadata),
+            patch.object(
+                Run, "get_status", return_value="INVOCATION_COMPLETED"
+            ),
+            patch.object(
+                Run, "_can_start_new_metric_computation", return_value=True
+            ),
         ):
             result = self.run.compute_metrics([metric])
         self.assertIsInstance(result, str)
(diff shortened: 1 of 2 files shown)


tests/unit/test_run_compute_metrics.py
```diff
@@ -0,0 +1,103 @@
...
+        selectors={"output": Selector.select_record_output()},
+    )
+
+
+class TestComputeMetricsMessage(unittest.TestCase):
+    def setUp(self):
+        if Run is None:
+            self.skipTest("Run not available.")
+        self.run = _make_run()
+
```
joshreini1: Drop `_patches` as a method — inline the three context managers directly in each test via `with patch.object(...), patch.object(...), patch.object(...):`. Saves four lines and keeps the mock setup adjacent to its use.

tests/unit/test_run_compute_metrics.py
```diff
@@ -0,0 +1,103 @@
...
+        return (
+            patch.object(Run, "describe", return_value=completed),
+            patch.object(
+                Run, "get_status", return_value="INVOCATION_COMPLETED"
+            ),
+            patch.object(
+                Run, "_can_start_new_metric_computation", return_value=True
+            ),
+        )
+
```
joshreini1, line 71: Change the docstring to "Metric object passed to compute_metrics yields message when already computed (no TypeError)." — tighter and no hypothetical phrasing.

joshreini1 (approved):
Clean fix for a straightforward type error. The change correctly mirrors the same type-checking pattern used in `_should_skip_computation` and the test covers both the failure case and the string case. Ready to merge.

Abelo9996:
Thanks for the review. Both addressed in the follow-up commit: inlined the three `patch.object` context managers in each test and dropped the `_patches` helper, and tightened the object-case docstring to "Metric object passed to compute_metrics yields message when already computed (no TypeError)."

PR #2747, opened 2026-08-29
fix(groundtruth): bound recall_at_k and ndcg_at_k on duplicate retrieved chunks (#2746)

### LINKED ISSUE #2746: [BUG] recall_at_k and ndcg_at_k can exceed 1.0 on duplicate retrieved chunks
**Bug Description**

`GroundTruthAgreement.recall_at_k` and `ndcg_at_k` can return values greater than 1.0 when the retriever returns the same relevant chunk more than once. Both are ratios bounded to `[0, 1]`.

- `recall_at_k` counts occurrences of golden chunks in the retrieved list, but the denominator is the number of unique golden chunks (`golden_chunks` is a set), so a duplicated relevant chunk makes the numerator exceed the denominator.
- `ndcg_at_k` credits every duplicate retrieved position with the golden relevance, so DCG exceeds the ideal DCG (which counts each golden chunk once).

`recall_at_k` also reports the wrong value in the non-overshoot case: with golden `{A, B}` and retrieved `["A", "A"]` it returns `1.0` although only one of two golden chunks was retrieved (true recall `0.5`).

**To Reproduce**

```python
from trulens.feedback.groundtruth import GroundTruthAgreement
from trulens.feedback.dummy.provider import DummyProvider

gt = [{"query": "q", "expected_response": "", "expected_chunks": [{"text": "A", "expect_score": 1}]}]
a = GroundTruthAgreement(ground_truth=gt, provider=DummyProvider())

print(a.recall_at_k("q", ["A", "A", "X"], k=3))  # 2.0, expected 1.0
print(a.ndcg_at_k("q", ["A", "A"], k=2))          # ~1.63, expected <= 1.0
```

**Expected Behavior**

Both metrics stay within `[0, 1]`, and recall reflects the fraction of unique golden chunks retrieved, regardless of duplicates in the retrieved list.

**Additional Context**

The duplicate-retrieved case was not exercised by the existing tests. `precision_at_k` is unaffected because its denominator grows with the retrieved list, so duplicates do not push it above 1.

Abelo9996:
Closes #2746

## Problem

`GroundTruthAgreement.recall_at_k` and `ndcg_at_k` can return values above 1.0 when the retriever returns the same relevant chunk more than once. Both are ratios bounded to `[0, 1]`.

- `recall_at_k` counts occurrences of golden chunks in the retrieved list while dividing by the number of unique golden chunks (`golden_chunks` is a set), so a duplicated relevant chunk makes the numerator exceed the denominator. It also reports the wrong value in the non-overshoot case: golden `{A, B}` with retrieved `["A", "A"]` returns `1.0` where the true recall is `0.5`.
- `ndcg_at_k` credits every duplicate retrieved position, so DCG exceeds the ideal DCG (which counts each golden chunk once).

## Fix

- `recall_at_k` counts unique golden chunks that were retrieved: `len(golden_chunks & set(retrieved_top_k))`.
- `ndcg_at_k` credits each golden chunk at most once (tracks consumed golden chunks while building the relevance vector).

Non-duplicate behavior is unchanged. `precision_at_k` is unaffected because its denominator grows with the retrieved list.

## Tests

Adds `tests/unit/test_groundtruth_ir_metrics.py` covering both metrics on duplicate chunks plus non-duplicate sanity cases. The duplicate cases fail on `main` (recall 2.0, ndcg ~1.63) and pass here. Lint and format clean under the pinned ruff.

### DIFF
--- a/src/feedback/trulens/feedback/groundtruth.py
+++ b/src/feedback/trulens/feedback/groundtruth.py
@@ -367,12 +367,18 @@ class GroundTruthAgreement(
             rel_scores = [0.0] * len(
                 retrieved_context_chunks
             )  # Initialize with 0 relevance for all
+            # Credit each golden chunk at most once: a retriever that returns
+            # the same relevant chunk more than once must not earn its
+            # relevance repeatedly, which would push DCG above the ideal DCG
+            # and yield NDCG > 1.
+            consumed_golden = set()
             for i, chunk in enumerate(retrieved_context_chunks[:k]):
-                if chunk in golden_chunks:
+                if chunk in golden_chunks and chunk not in consumed_golden:
                     index_in_golden = golden_chunks.index(chunk)
                     rel_scores[i] = golden_scores[
                         index_in_golden
                     ]  # Use the true relevance score
+                    consumed_golden.add(chunk)
 
             # Step 5: Compute NDCG@k directly from the standard formula.
             # DCG discounts each retrieved chunk's golden relevance by the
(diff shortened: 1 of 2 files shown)


joshreini1 (approved):
Clean fix for IR metric bounds on duplicate retrieved chunks. Both recall_at_k and ndcg_at_k now correctly credit each golden chunk at most once, using set intersection and consumed-chunk tracking respectively. The logic is sound: recall_at_k uses `golden_chunks & set(retrieved_top_k)` since golden_chunks is a set (line 477), and ndcg_at_k tracks consumed chunks since golden_chunks is a list (line 346). Test coverage demonstrates the fix handles duplicates while preserving non-duplicate behavior. Ready to merge.

PR #2748, opened 2026-08-30
fix(groundtruth): refuse a score list that does not match the labels in recall and precision

arthi-arumugam-git:
`recall` and `precision` walk `scores` and `true_labels` with `zip`, which stops at the shorter list. If the score list lost a row somewhere upstream, every row after the gap is scored against the wrong label, and the result is an ordinary-looking number.

On `main`, with `true_labels=[1, 0, 1, 0, 1]` and four scores `[0.9, 0.9, 0.1, 0.9]` (one row missing):

```
recall    -> 0.5
precision -> 0.3333
```

Nothing raises. `brier_score`, `ece` and `cohens_kappa` in the same class already refuse a length mismatch, so the classification metrics were the odd ones out.

## Change

- A small `_check_length` helper, called from `recall` and `precision` after the list-of-lists unpack, raising `ValueError` with both lengths in the message (the same shape `cohens_kappa` uses). `f1_score` inherits it through the two calls it makes.
- The `try/except` around the unpack in `recall`, which printed a traceback and carried on, is removed. `precision` never had it, and the comparison on the next line would have raised anyway.

## Not changed

`mae` builds two arrays and subtracts, so a mismatch already raises from numpy, except when one side has length 1 and broadcasts. I have left that alone here rather than widen this.

## Tests

Folded into `test_input_validation_comprehensive`, next to the existing brier and ECE cases: `recall`, `precision` and `f1_score` each reject a short list, and `recall` rejects a long one. With the source change reverted, that test fails on `main`; with it, the file passes (22 passed, `TEST_OPTIONAL=true`).

### DIFF
--- a/src/feedback/trulens/feedback/groundtruth.py
+++ b/src/feedback/trulens/feedback/groundtruth.py
@@ -1018,6 +1018,22 @@ class GroundTruthAggregator(
 
         setattr(self, name, lambda scores: func(scores, self))
 
+    def _check_length(self, scores: list) -> None:
+        """Refuse a score list that does not line up with the labels.
+
+        recall and precision walk the two lists with zip, which stops at the
+        shorter one. A score list that lost a row somewhere upstream is then
+        scored against the wrong labels for every row after the gap, and the
+        result is an ordinary-looking number. brier_score, ece and
+        cohens_kappa already refuse this; this makes the classification
+        metrics do the same.
+        """
+        if len(scores) != len(self.true_labels):
+            raise ValueError(
+                "The length of true_labels and scores must be the same; "
+                f"got {len(self.true_labels)} labels and {len(scores)} scores."
+            )
+
     def auc(self, scores: list[float]) -> float:
         """
         Calculate the area under the ROC curve. Can be used for meta-evaluation.
@@ -1158,15 +1174,10 @@ class GroundTruthAggregator(
         - float: The recall score.
         """
 
-        try:
-            if isinstance(scores[0], list):
-                scores = [score for score, _ in scores]
-        except Exception as e:  # noqa: BLE001
-            import traceback
+        if isinstance(scores[0], list):
+            scores = [score for score, _ in scores]
+        self._check_length(scores)
 
-            traceback.print_exc()
-            logger.error(f"scores processing failed in Recall aggregator: {e}")
-            print(f"scores processing failed in Recall aggregator: {e}")
         # Convert scores to binary predictions based on the threshold
         predictions = [1 if score >= threshold else 0 for score in scores]
 
@@ -1206,6 +1217,7 @@ class GroundTruthAggregator(
         """
         if isinstance(scores[0], list):
             scores = [score for score, _ in scores]
+        self._check_length(scores)
 
         # Convert scores to binary predictions based on the threshold
         predictions = [1 if score >= threshold else 0 for score in scores]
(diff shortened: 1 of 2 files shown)


arthi-arumugam-git:
Same CI failure as #2738, and the same fix, pushed as 1f9d76e.

This branch is stacked on the nan change, so it inherits `test_empty_retrieved_chunks` still asserting the old `0.0` for precision when nothing is retrieved. Only precision moves; recall, mrr and ir_hit_rate keep a denominator there and stay measured zeros.

If #2738 lands first this will rebase away cleanly. Locally both groundtruth test files run green with `--run_optional_tests`.

src/feedback/trulens/feedback/groundtruth.py
```diff
@@ -385,7 +385,9 @@ def ndcg_at_k(
             # several retrieved chunks share the same (zero) relevance.
             dcg = _dcg(rel_scores[:k])
             ideal_dcg = _dcg(sorted(golden_scores, reverse=True)[:k])
-            return dcg / ideal_dcg if ideal_dcg > 0 else 0.0
+            # ideal_dcg of 0 means the golden set annotates nothing relevant,
```
joshreini1, line 394: Rebase after #2738 merges to remove all nan return changes (lines 388-390, 448-450, 508-510, 549-551, 591-593, 891-893, 942-944, 1194-1196, 1231-1233, plus the corresponding test updates in test_zero_division_handling and test_empty_retrieved_chunks). These changes belong to #2738 and are out-of-scope for length validation, making this feature diff unreviewable by mixing two concerns into one PR.

src/feedback/trulens/feedback/groundtruth.py
```diff
@@ -995,6 +1011,22 @@ def register_custom_agg_func(
 
         setattr(self, name, lambda scores: func(scores, self))
 
+    def _check_length(self, scores: list) -> None:
```
joshreini1, line 1021: The docstring and error message are excellent—clear explanation of why zip is dangerous here and how both lengths help debug. The placement after unpacking in both recall and precision is correct since you need the flat list to validate length.

tests/unit/test_groundtruth_aggregator.py
```diff
@@ -67,6 +67,21 @@ def test_input_validation_comprehensive(self):
         ):
             self.aggregator.ece([(0.5, 0.8), (0.3, 0.7)])
 
+        # recall, precision and f1_score used to zip past the gap and score
```
joshreini1, line 70: Good coverage: testing all three affected methods (recall, precision, f1_score) plus both short and long mismatch cases. The msg arguments make test failures actionable.

joshreini1 (changes requested):
The length-checking feature is solid: `_check_length` correctly validates list lengths before zip operations, the try/except removal in `recall` is justified, and test coverage is thorough. However, this PR is stacked on #2738 and mixes two separate features—the length validation (this PR's scope) and extensive nan-return changes (from #2738)—making targeted review of the length-checking logic difficult. Request changes: 1 blocker (rebase after #2738 merges to isolate this feature).

arthi-arumugam-git:
Rebased now that #2738 is in, so this is down to the length check on its own: one commit, two files.

Both nan commits dropped out as already upstream, and the `test_groundtruth_agreement.py` changes went with them, so what is left is `_check_length` and the tests that cover it. Sorry for the stacking, I should have waited and put it up on its own.

The rebase also picked up #2747, which touches the same file, and there was no conflict with it.

joshreini1 (approved):

sfc-gh-jreini (approved):

sfc-gh-jreini:
bypassing certifications here because they were added after this PR was created.

PR #2750, opened 2026-08-31
fix(dashboard): label conversation-scoped thread columns as latest, not avg

Abelo9996:
Follow-up to #2725.

## Problem

#2725 fixed the thread summary table to show the latest value for conversation-scoped metrics (like Topic Adherence) instead of the mean. That change touched the value computation but not the grid header, which still labels every feedback column `"(avg)"`. So a conversation-scoped metric now reads as an average in the header while the cell shows the latest value. This inconsistency was introduced by #2725.

## Fix

`_render_thread_view` already has the records DataFrame, so it computes the conversation-scoped columns via `_partition_feedback_scopes` and threads them through `_render_thread_grid` into `_build_thread_grid_options`. Conversation-scoped columns are labelled `"(latest)"`; turn-scoped columns keep `"(avg)"`.

## Tests

Adds `test_thread_grid_header_labels_conversation_vs_turn_metrics` in `tests/unit/streamlit/test_streamlit_records.py`, asserting a conversation-scoped column gets `"(latest)"` and a turn-scoped column gets `"(avg)"`. Lint and format clean under the pinned ruff.

### DIFF
--- a/src/dashboard/trulens/dashboard/tabs/Records.py
+++ b/src/dashboard/trulens/dashboard/tabs/Records.py
@@ -931,6 +931,7 @@ def _build_thread_grid_options(
     df: pd.DataFrame,
     feedback_col_names: Sequence[str],
     feedback_directions: Dict[str, bool],
+    conversation_cols: Optional[Sequence[str]] = None,
 ):
     from st_aggrid.grid_options_builder import GridOptionsBuilder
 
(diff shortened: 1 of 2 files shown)


joshreini1 (approved):
Clean fix that addresses the header/value mismatch from #2725. Conversation-scoped metrics now show "(latest)" labels matching their actual latest-value semantics, while turn-scoped metrics keep "(avg)". The scope information threads cleanly through _build_thread_grid_options, _render_thread_grid, and _render_thread_view with clear comments. Unit test verifies both label types. Ready to merge.

PR #2752, opened 2026-09-02
fix(providers/openai): moderation_harassment_threatening reads its own category (#2751)

### LINKED ISSUE #2751: [BUG] moderation_harassment_threatening returns the harassment score, not harassment_threatening
**Bug Description**

`OpenAI.moderation_harassment_threatening` returns `category_scores.harassment` instead of `category_scores.harassment_threatening`, so it silently reports the plain harassment score rather than the harassment/threatening score. Every other `moderation_*` method reads its own distinct category field (e.g. `moderation_violencegraphic` reads `violence_graphic`). The method's summary docstring is also copy-pasted from the graphic-violence method ("checks if text is about graphic violence").

**To Reproduce**

```python
from unittest.mock import MagicMock
from trulens.providers.openai.provider import OpenAI

p = OpenAI.__new__(OpenAI)
resp = MagicMock()
resp.category_scores = MagicMock(harassment=0.11, harassment_threatening=0.99)
p._moderation = lambda text: resp

print(p.moderation_harassment("x"))              # 0.11
print(p.moderation_harassment_threatening("x"))  # 0.11, expected 0.99
```

**Expected Behavior**

`moderation_harassment_threatening` returns the `harassment_threatening` category score.

**Additional Context**

`harassment_threatening` is a real field on `openai.types.moderation.CategoryScores`. The method had no functional test coverage.

Abelo9996:
Closes #2751

## Problem

`OpenAI.moderation_harassment_threatening` returns `category_scores.harassment` instead of `category_scores.harassment_threatening`, so it silently reports the plain harassment score. Every other `moderation_*` method reads its own distinct category field. The method's summary docstring was also copy-pasted from the graphic-violence method.

## Fix

Read `category_scores.harassment_threatening`, and correct the docstring summary line. `moderation_harassment` is unchanged.

## Tests

Adds `tests/unit/providers/test_openai_moderation.py` asserting the two methods return distinct fields (harassment vs harassment_threatening). It fails on `main` and passes with this change. Lint and format clean under the pinned ruff.

### DIFF
--- a/src/providers/openai/trulens/providers/openai/provider.py
+++ b/src/providers/openai/trulens/providers/openai/provider.py
@@ -753,7 +753,7 @@ class OpenAI(llm_provider.LLMProvider):
         return float(openai_response.category_scores.harassment)
 
     def moderation_harassment_threatening(self, text: str) -> float:
-        """A function that checks if text is about graphic violence.
+        """A function that checks if text is about harassment/threatening.
 
         Example:
             ```python
@@ -779,7 +779,7 @@ class OpenAI(llm_provider.LLMProvider):
         """
         openai_response = self._moderation(text)
 
-        return float(openai_response.category_scores.harassment)
+        return float(openai_response.category_scores.harassment_threatening)
 
 
 class AzureOpenAI(OpenAI):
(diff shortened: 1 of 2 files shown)


joshreini1 (approved):
Clean fix for a copy-paste bug where moderation_harassment_threatening returned the wrong category score. The code now correctly reads harassment_threatening instead of harassment, and the docstring is fixed (was copy-pasted from graphic violence). The regression test is well-structured, uses a clever __new__ pattern to avoid API keys, and would catch this bug. Ready to merge.

joshreini1 (approved):
thanks for finding and fixing!

PR #2754, opened 2026-09-02
fix(apps): use Lens.get for methods_to_instrument existence check (#2753)

### LINKED ISSUE #2753: [BUG] methods_to_instrument existence check always fails (Lens called instead of .get)
**Bug Description**

`TruApp.__init__` checks whether an instrumented method's component exists in the serialized app before creating a placeholder, using `next(full_path(json))`. But a `Lens` is not callable as a lookup: `Lens.__call__` only accepts a path ending in `collect` and otherwise raises `TypeError`. Since `full_path` never ends in `collect`, the check always raises and takes the `except` branch. As a result:

- The success branch that prints `✓ Added method ... under component at path ...` is dead code, never reached.
- Users who pass `methods_to_instrument` always get the warning `App has no component at path ... Creating a placeholder there for now.`, even when the component genuinely exists at that path.

The sibling existence check in the same method (a few lines below) correctly uses `full_path.get(json)`.

**Expected Behavior**

When the component exists at the given path, the existence check succeeds (no spurious warning, no unnecessary placeholder).

**Additional Context**

The defect dates to the TruCustomApp to TruApp migration, which is the same migration that split the Lens `__call__` and `get` APIs (the `TypeError` message names it). This code path had no test coverage.

Abelo9996:
Closes #2753

## Problem

`TruApp.__init__` checks whether an instrumented method's component exists in the serialized app using `next(full_path(json))`. A `Lens` is not callable as a lookup: `Lens.__call__` only accepts a path ending in `collect` and otherwise raises `TypeError`. Since `full_path` never ends in `collect`, the check always raises. As a result the `✓ Added method ... under component` success branch is dead code, and users who pass `methods_to_instrument` always get a spurious `App has no component at path ...` warning even when the component exists.

The sibling existence check in the same method already uses the correct `full_path.get(json)`.

## Fix

Use `full_path.get(json)` for the existence check, matching the sibling.

## Tests

Adds `tests/unit/test_truapp_methods_to_instrument.py` pinning the Lens lookup contract: a non-`collect` path is read with `.get()` and yields the value, while calling it raises `TypeError` (the misuse the fix corrects). Lint and format clean under the pinned ruff.

### DIFF
--- a/tests/unit/test_truapp_methods_to_instrument.py
+++ b/tests/unit/test_truapp_methods_to_instrument.py
@@ -1,0 +1,35 @@
+"""Regression guard for the methods_to_instrument path-existence check.
+
+TruApp.__init__ checks whether an instrumented method's component exists in the
+serialized app before adding a placeholder. It used `next(full_path(json))`,
+but a Lens is not callable as a lookup: `Lens.__call__` only accepts a path
+ending in `collect` and otherwise raises TypeError. So the check always failed,
+the "Added method ... under component" success branch was dead, and users saw a
+spurious "App has no component at path" warning even when the component existed.
+The fix uses `full_path.get(json)`, matching the sibling check in the same
+method. This test pins the Lens contract that makes `.get` mandatory.
+"""
+
+from __future__ import annotations
+
+import unittest
+
+from trulens.core.utils.serial import Lens
+
+
+class TestLensLookupContract(unittest.TestCase):
+    def test_get_reads_existing_path_but_call_raises(self):
+        json = {"app": {"component": {"leaf": 42}}}
+        full_path = Lens().app.component.leaf
+
+        # The read the fix relies on.
+        self.assertEqual(next(full_path.get(json)), 42)
+
+        # Calling a non-`collect` Lens is not a lookup and raises, which is why
+        # the existence check must use `.get`, not `__call__`.
+        with self.assertRaises(TypeError):
+            next(full_path(json))
+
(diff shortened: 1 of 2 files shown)


tests/unit/test_truapp_methods_to_instrument.py
```diff
@@ -0,0 +1,35 @@
...
+from trulens.core.utils.serial import Lens
+
+
+class TestLensLookupContract(unittest.TestCase):
+    def test_get_reads_existing_path_but_call_raises(self):
+        json = {"app": {"component": {"leaf": 42}}}
+        full_path = Lens().app.component.leaf
+
+        # The read the fix relies on.
+        self.assertEqual(next(full_path.get(json)), 42)
```
joshreini1, line 26: Drop `self.assertEqual(next(full_path.get(json)), 42)` — this only verifies that `.get()` works, which is already covered by existing Lens tests; the value of this test is solely line 30's TypeError assertion (the bug the fix corrects). One assertion per test keeps it focused on the regression it guards.

joshreini1 (approved):
Clean fix for a real bug: the existence check was calling the Lens as `full_path(json)` which always raised TypeError per Lens.__call__'s contract (line 1095 in serial.py explicitly rejects non-collect paths). The fix correctly uses `.get()` matching the sibling check at line 428. The new test pins the lookup contract and would have caught the original bug. Ready to merge.

PR #2756, opened 2026-09-02
fix(providers/anthropic): Anthropic cost tracking crashes on core_endpoint.Cost (#2755)

### LINKED ISSUE #2755: [BUG] Anthropic cost tracking crashes: core_endpoint.Cost does not exist
**Bug Description**

The Anthropic endpoint builds cost with `core_endpoint.Cost(...)` in two places (`AnthropicCallback.handle_generation` and `AnthropicEndpoint.handle_wrapped_call`), but `trulens.core.feedback.endpoint` (imported as `core_endpoint`) does not export a `Cost` symbol. Every Anthropic call routed through the cost callback therefore raises `AttributeError: module 'trulens.core.feedback.endpoint' has no attribute 'Cost'`.

Both sites also pass `currency=...`, but `Cost` has no `currency` field (the field is `cost_currency`), so once the `AttributeError` is fixed a non-USD currency would still be dropped.

The sibling OpenAI endpoint correctly uses `base_schema.Cost(...)`.

**To Reproduce**

```python
from types import SimpleNamespace
from trulens.providers.anthropic.endpoint import AnthropicCallback
from trulens.core.schema import base as base_schema

cb = AnthropicCallback.model_construct(cost=base_schema.Cost())
resp = SimpleNamespace(usage=SimpleNamespace(input_tokens=10, output_tokens=5),
                       model="claude-3-5-sonnet-20241022")
cb.handle_generation(resp)   # AttributeError: ... has no attribute 'Cost'
```

**Expected Behavior**

The Anthropic cost callback accumulates cost/tokens without crashing, and preserves the currency.

**Additional Context**

`handle_generation` / `handle_wrapped_call` had no test coverage (existing Anthropic tests only cover the OTEL `handle_response` path).

Abelo9996:
Closes #2755

## Problem

The Anthropic endpoint builds cost with `core_endpoint.Cost(...)` in two places, but `trulens.core.feedback.endpoint` exposes no `Cost` symbol, so every Anthropic call through the cost callback raises `AttributeError: module 'trulens.core.feedback.endpoint' has no attribute 'Cost'`. Both sites also pass `currency=`, but the `Cost` field is `cost_currency`, so a non-USD currency would be dropped once the crash is fixed. The sibling OpenAI endpoint uses `base_schema.Cost(...)`.

## Fix

Import `trulens.core.schema.base as base_schema` (as the OpenAI endpoint does), use `base_schema.Cost(...)` at both sites, and pass `cost_currency=` instead of `currency=`.

## Tests

Adds `tests/unit/providers/test_anthropic_cost.py`, which runs `AnthropicCallback.handle_generation` end to end and asserts cost/tokens accumulate. It fails on `main` with the `AttributeError` above and passes with this change. Lint and format clean under the pinned ruff.

### DIFF
--- a/src/providers/anthropic/trulens/providers/anthropic/endpoint.py
+++ b/src/providers/anthropic/trulens/providers/anthropic/endpoint.py
@@ -13,6 +13,7 @@ from typing import (
 from litellm import model_cost
 import pydantic
 from trulens.core.feedback import endpoint as core_endpoint
+from trulens.core.schema import base as base_schema
 from trulens.otel.semconv.trace import SpanAttributes
 
 import anthropic
@@ -91,9 +92,9 @@ class AnthropicCallback(core_endpoint.EndpointCallback):
         super().handle_generation(response)
         cost_info = AnthropicCostComputer.handle_response(response)
 
-        addl_cost = core_endpoint.Cost(
+        addl_cost = base_schema.Cost(
             cost=cost_info.get(SpanAttributes.COST.COST, 0.0),
-            currency=cost_info.get(SpanAttributes.COST.CURRENCY, "USD"),
+            cost_currency=cost_info.get(SpanAttributes.COST.CURRENCY, "USD"),
             n_tokens=cost_info.get(SpanAttributes.COST.NUM_TOKENS, 0),
             n_prompt_tokens=cost_info.get(
                 SpanAttributes.COST.NUM_PROMPT_TOKENS, 0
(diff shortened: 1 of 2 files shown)


joshreini1 (approved):
Clean fix for the AttributeError crash—correctly changes core_endpoint.Cost to base_schema.Cost and currency= to cost_currency= in both callback sites, matching the OpenAI endpoint pattern. The regression test proves the crash is fixed and cost accumulation works. Ready to merge.

PR #2697, opened 2026-08-15
Add trulens-openeval: to_openeval()/from_openeval() for EvalPort interchange

adhabnr-ux:
Closes #2680.

This is the minimal `to_openeval()`/`from_openeval()` module discussed and approved in that thread — a standalone module in `src/`, not a separate package, per @sfc-gh-jreini's scope call ("a `to_openeval()`/`from_openeval()` module in `src/` is the right scope to start").

## What it does

Converts between TruLens `Run` records and [EvalPort](https://github.com/adhabnr-ux/evalport) (Apache 2.0), the open interchange format for portable LLM evaluation test cases, graders, suites, and results.

- **`to_openeval(records_df, ...)`** — takes the DataFrame `Run.get_records()`/`get_record_details()` returns (`record_id`, `input`, `output`, `latency`, + one column per feedback score) and produces an EvalPort `ResultSet`. Each feedback column becomes its own `GraderResult` (`"Context Relevance"` → `grader_id: "context_relevance"`, `type: "custom"`). It correctly excludes TruLens's `<name>_calls` companion columns (per-call detail, not a score) and clamps scores into EvalPort's required `[0, 1]` range.
- **`from_openeval(suite)`** — returns `(input_df, dataset_spec)`, where `dataset_spec` maps EvalPort's `input`/`expected_output`/`id` onto TruLens's actual reserved dataset-spec fields (`input`, `ground_truth_output`, `input_id`) — verified against `trulens.core.run.validate_dataset_spec`, so it's ready to hand straight to `RunConfig(dataset_spec=...)`.

## Why it converts DataFrames, not `Run`/`RunConfig` objects directly

`trulens.core.run.Run` needs a live `RunDaoBase` + `TruSession` + app instance just to construct — there's no way to unit test against it without standing up a full session and database (confirmed by reading `trulens/core/run.py`). `Run.get_records()`/`get_record_details()` (and `Run.start(input_df=...)`) already hand back/accept plain DataFrames, and that's the real portable surface, so this module converts at that boundary instead — the same reason `opik-openeval-adapter` and `ragas-openeval-adapter` convert at their SDKs' plain-data boundaries.

## Testing

[...]

PR #2757, opened 2026-09-02
Add trulens-connectors-openeval: to_openeval()/from_openeval() for EvalPort interchange

adhabnr-ux:
Supersedes #2697, which GitHub auto-closed when I deleted and recreated its head branch (`openeval-adapter`) to rebase it onto current `main` -- my mistake; GitHub doesn't allow reopening a PR once its head branch is deleted, even after recreating a branch with the same name. Apologies for the churn. All prior context, review, and approval from that thread carries over here; nothing about the design or scope has changed.

## What it does

Converts between TruLens `Run` records and [EvalPort](https://github.com/adhabnr-ux/evalport) (Apache 2.0), the open interchange format for portable LLM evaluation test cases, graders, suites, and results.

- **`to_openeval(records_df, ...)`** -- takes the DataFrame `Run.get_records()`/`get_record_details()` returns (`record_id`, `input`, `output`, `latency`, + one column per feedback score) and produces an EvalPort `ResultSet`. Each feedback column becomes its own `GraderResult` (`"Context Relevance"` -> `grader_id: "context_relevance"`, `type: "custom"`). Excludes TruLens's `<name>_calls` companion columns (per-call detail, not a score) and clamps scores into EvalPort's required `[0, 1]` range.
- **`from_openeval(suite)`** -- returns `(input_df, dataset_spec)`, where `dataset_spec` maps EvalPort's `input`/`expected_output`/`id` onto TruLens's actual reserved dataset-spec fields (`input`, `ground_truth_output`, `input_id`) -- verified against `trulens.core.run.validate_dataset_spec`.

Design rationale (DataFrame boundary vs. `Run`/`RunConfig` objects directly, lossiness handling under `metadata["trulens"]`) is unchanged from #2697 and written up in full in `src/connectors/openeval/README.md`.

## What's different from #2697

Two things happened on that thread after the last update: @joshreini1 asked for the package to be renamed `trulens-openeval` -> **`trulens-connectors-openeval`** (more discoverable, and accurate -- it's a connector to an external format, not a TruLens-owned "openeval" concept), and `main` moved forward significantly (26 commits) while the old branch sat waiting on flaky-check reruns. This PR is that same code:

- Renamed and relocated per the request: `src/openeval/` -> `src/connectors/openeval/`, `trulens.openeval` -> `trulens.connectors.openeval`, mirroring `src/connectors/snowflake`'s layout exactly.
- Rebased onto current `main` (`fd2d2601`, TruLens 2.13.1) from scratch -- zero drift, so this should merge clean rather than needing another round of "which failing check is pre-existing."

## Testing

17 tests in `tests/unit/test_openeval.py`, re-verified against this rebased/renamed code: fresh venv, `pip install -e src/core && pip install -e src/connectors/openeval`, real `trulens-core` 2.13.1 and `evalport-sdk` 1.3.1 (not mocks) -- **17/17 passing**. Covers basic conversion + real-spec validation, `_calls` companion-column exclusion, score clamping, custom pass thresholds, both latency units, run-id defaulting, explicit timestamps, empty/malformed-input errors, summary counts, and a full suite -> `input_df` -> simulated run -> `ResultSet` round trip against the real `openeval.validate` validators.

## One thing left, and exactly what it is

**This PR does not include an updated `poetry.lock`.** My environment's git/API access to this repo is proxied through tooling that can reliably push small text files (every file above was pushed and byte-verified against its blob SHA) but cannot transport this repo's `poetry.lock` (currently ~1.07MB) as the single inline blob my write path requires, without real risk of mangling it.

I did generate and verify it locally, though, and the diff is small and deterministic -- **exactly two new package entries, alphabetically placed, plus the content-hash**, nothing else re-pinned:

```diff
+[[package]]
+name = "evalport-sdk"
+version = "1.3.1"
+description = "Python SDK for EvalPort — The Open Evaluation Standard"
+optional = false
+python-versions = ">=3.8"
+groups = ["required"]
+files = [
+    {file = "evalport_sdk-1.3.1-py3-none-any.whl", hash = "sha256:1fba7d5e933da09c04ebc9479347c4b852b47e696824081911bd70360904a41a"},
+    {file = "evalport_sdk-1.3.1.tar.gz", hash = "sha256:4172c31450bed6b2ed73243bdea91d939eee0be099476f6c6411bd0a59aa4884"},
+]
+
+[package.extras]
+test = ["jsonschema", "pytest"]
+
+[package.source]
+type = "legacy"
+url = "https://pypi.org/simple"
+reference = "pypi-public"
+
+[[package]]
+name = "trulens-connectors-openeval"
+version = "2.12.0"
+description = "Convert between TruLens Run records and EvalPort (https://github.com/adhabnr-ux/evalport) suites and result sets."
+optional = false
+python-versions = "^3.9"
+groups = ["required"]
+files = []
+develop = true
+
+[package.dependencies]
+evalport-sdk = ">=1.0.0"
+pandas = ">=1.0.0"
+trulens-core = "^2.0.0"
+
+[package.source]
+type = "directory"
+url = "src/connectors/openeval"
+
-content-hash = "cb032d20571ebbf5b831cf9de3561b1b9a53572409fa4565eed38eac5a41cf8e"
+content-hash = "c524c0d7960fcd4030aceebee670c53d9e84cf52fb969f578b46d03beee71c64"
```

Running `poetry lock` on this branch as-is should reproduce this exact diff (insertion point is alphabetical, right after the entries preceding `evalport-sdk`/`trulens-connectors-openeval`). One thing worth flagging while I'm here: I hit a real, reproducible multi-minute-plus hang running `poetry lock` with **Poetry 2.1.1** on this repo -- a `py-spy` stack trace showed it stuck in `poetry.core.version.markers`' `cnf`/`union`/`_flatten_markers` (`merge_override_packages` in `puzzle/solver.py`), almost certainly from the number of override packages with differing `python =` markers across `providers`/`connectors`/etc. Reinstalling **Poetry 2.3.2** (matching what's actually pinned in the committed lock file's generator line) resolved cleanly in ~3.5 minutes with no such issue. Not something to fix in this PR, but possibly worth a pinned-version note in CONTRIBUTING/Makefile if a contributor on an older Poetry hits the same wall.

Sorry again for the extra round-trip -- happy to do anything else needed from my end to get this the rest of the way.

https://claude.ai/code/session_01RtocdH3tKifGdkiCZxxV3F

### DIFF
--- a/src/connectors/openeval/pyproject.toml
+++ b/src/connectors/openeval/pyproject.toml
@@ -1,0 +1,37 @@
+[build-system]
+build-backend = "poetry.core.masonry.api"
+requires = [
+  "poetry-core",
+]
+
+[tool.poetry]
+name = "trulens-connectors-openeval"
+version = "2.12.0"
+description = "Convert between TruLens Run records and EvalPort (https://github.com/adhabnr-ux/evalport) suites and result sets."
+authors = [
+  "Snowflake Inc. <ml-observability-wg-dl@snowflake.com>",
+]
+license = "MIT"
+readme = "README.md"
+packages = [
+  { include = "trulens" },
+]
+homepage = "https://trulens.org/"
+documentation = "https://trulens.org/getting_started/"
+repository = "https://github.com/truera/trulens"
+classifiers = [
+  "Programming Language :: Python :: 3",
+  "Operating System :: OS Independent",
+  "Development Status :: 3 - Alpha",
+  "License :: OSI Approved :: MIT License",
+]
+
+[tool.poetry.dependencies]
+python = "^3.9"
+trulens-core = { version = "^2.0.0" }
+pandas = ">=1.0.0"
+evalport-sdk = ">=1.0.0"
+
+[tool.poetry.group.dev.dependencies]
+trulens-core = { path = "../../core", develop = true }
+pytest = "^7.0.0"
--- a/src/connectors/openeval/pyproject.toml
+++ b/src/connectors/openeval/pyproject.toml
@@ -6,7 +6,7 @@ requires = [
 
 [tool.poetry]
 name = "trulens-connectors-openeval"
-version = "2.12.0"
+version = "2.13.1"
 description = "Convert between TruLens Run records and EvalPort (https://github.com/adhabnr-ux/evalport) suites and result sets."
 authors = [
   "Snowflake Inc. <ml-observability-wg-dl@snowflake.com>",
--- a/src/connectors/openeval/trulens/connectors/openeval/__init__.py
+++ b/src/connectors/openeval/trulens/connectors/openeval/__init__.py
@@ -47,14 +47,7 @@ from datetime import datetime
 from datetime import timezone
 from typing import Any, Dict, List, Optional, Tuple
 
-try:
-    import pandas as pd
-except (
-    ImportError
-) as e:  # pragma: no cover - trulens-core already depends on pandas
-    raise ImportError(
-        "trulens-connectors-openeval requires pandas, which trulens-core already depends on."
-    ) from e
+import pandas as pd
 
 try:
     from openeval.version import OPENEVAL_VERSION
--- a/src/connectors/openeval/trulens/connectors/openeval/__init__.py
+++ b/src/connectors/openeval/trulens/connectors/openeval/__init__.py
@@ -1,0 +1,296 @@
+"""Convert between TruLens's Run/records data and EvalPort suites and result sets.
+
+EvalPort (https://github.com/adhabnr-ux/evalport) is an open interchange format
+(Apache 2.0) for portable LLM evaluation datasets: test cases, graders, suites,
+and results as plain JSON, shared across evaluation tools (DeepEval, Promptfoo,
+Inspect AI, AutoGen, CrewAI, Ragas, LangSmith, Braintrust, MLflow, Opik, and
+now TruLens).
+
+This module has two entry points, matching the shape used by every other
+EvalPort adapter in the ecosystem:
+
+    to_openeval(records_df, ...)
+        Converts the DataFrame returned by ``Run.get_records()`` or
+        ``Run.get_record_details()`` -- ``record_id``, ``input``, ``output``,
+        ``latency``, plus one column per computed feedback score -- into an
+        EvalPort ``ResultSet``, one ``GraderResult`` per feedback column.
+
+    from_openeval(suite, ...)
+        Converts an EvalPort suite's test cases into a pandas DataFrame plus
+        a ready-to-use ``dataset_spec`` mapping, suitable to pass straight
+        into ``Run.start(input_df=...)`` after constructing a
+        ``RunConfig(dataset_spec=dataset_spec, ...)``.
+
+Why a DataFrame boundary rather than the ``Run``/``RunConfig`` objects
+directly: ``trulens.core.run.Run`` is a pydantic model that requires a live
+``RunDaoBase``, ``TruSession`` and app instance just to construct -- it is
+inherently tied to a running TruLens session with a backing database. The
+DataFrames it hands back from ``get_records()``/``get_record_details()`` (and
+accepts via ``start(input_df=...)``) are the actual portable surface, so this
+module converts at that boundary rather than requiring a live session to
+import or test it -- the same reason ``opik-openeval-adapter`` and
+``ragas-openeval-adapter`` convert at their SDKs' plain-data boundaries
+instead of their live-client objects.
+
+Only the reserved TruLens dataset-spec fields this module explicitly maps
+(``input``, ``ground_truth_output``, ``input_id``, ``record_id``, ``output``,
+``latency``) round-trip through EvalPort's own TestCase/Result shape.
+Everything else a *different* EvalPort-speaking tool would not know how to
+interpret is preserved under ``metadata["trulens"]`` rather than silently
+dropped -- the same lossiness tradeoff documented in every other adapter in
+this ecosystem (see e.g. adapters/opik-openeval-adapter's README).
+"""
+
+from __future__ import annotations
+
+from datetime import datetime
+from datetime import timezone
+from typing import Any, Dict, List, Optional, Tuple
+
+try:
+    import pandas as pd
+except (
+    ImportError
+) as e:  # pragma: no cover - trulens-core already depends on pandas
+    raise ImportError(
+        "trulens-connectors-openeval requires pandas, which trulens-core already depends on."
+    ) from e
+
+try:
+    from openeval.version import OPENEVAL_VERSION
+except ImportError:  # pragma: no cover - evalport-sdk not installed
+    OPENEVAL_VERSION = "1.0.0"
+
+__all__ = ["to_openeval", "from_openeval"]
+
+# Columns Run.get_records() always includes that are not feedback scores.
+_RECORD_OVERVIEW_COLUMNS = {"record_id", "input", "output", "latency"}
+
+# TruLens pairs some feedback columns with a "<name>_calls" companion column
+# holding the per-call detail (args/reason) for that feedback function. That
+# companion is not itself a score and must never be treated as one.
+_CALLS_SUFFIX = "_calls"
+
+
+def _now_iso() -> str:
+    return datetime.now(timezone.utc).strftime("%Y-%m-%dT%H:%M:%SZ")
+
+
+def _feedback_column_to_grader_id(column_name: str) -> str:
+    """Normalize a TruLens feedback name ('Context Relevance') into an
+    EvalPort grader_id ('context_relevance')."""
+    return str(column_name).strip().lower().replace(" ", "_")
+
+
+def _is_missing(value: Any) -> bool:
+    if value is None:
+        return True
+    try:
+        return bool(pd.isna(value))
+    except (TypeError, ValueError):
+        return False
+
+
+def to_openeval(
+    records_df: "pd.DataFrame",
+    metric_columns: Optional[List[str]] = None,
+    suite_id: str = "trulens_run",
+    run_id: Optional[str] = None,
+    pass_threshold: float = 0.5,
+    latency_unit: str = "seconds",
+    started_at: Optional[str] = None,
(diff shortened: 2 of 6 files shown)


src/connectors/openeval/pyproject.toml
```diff
@@ -0,0 +1,37 @@
+[build-system]
+build-backend = "poetry.core.masonry.api"
+requires = [
+  "poetry-core",
+]
+
+[tool.poetry]
+name = "trulens-connectors-openeval"
+version = "2.12.0"
+description = "Convert between TruLens Run records and EvalPort (https://github.com/adhabnr-ux/evalport) suites and result sets."
```
joshreini1, line 10: Version is 2.12.0 but the PR description says this rebased onto main at fd2d2601 (TruLens 2.13.1). Should this be 2.13.0 or does connectors/openeval maintain an independent version?

src/connectors/openeval/trulens/connectors/openeval/__init__.py
```diff
@@ -0,0 +1,296 @@
...
+from datetime import timezone
+from typing import Any, Dict, List, Optional, Tuple
+
+try:
+    import pandas as pd
+except (
+    ImportError
+) as e:  # pragma: no cover - trulens-core already depends on pandas
+    raise ImportError(
+        "trulens-connectors-openeval requires pandas, which trulens-core already depends on."
```
joshreini1: The pandas ImportError pragma no cover claims trulens-core already depends on pandas, but the try/except still exists. Since pandas is explicitly listed in this package's own pyproject.toml dependencies (line 32), this guard is redundant — drop the try/except and import pandas directly.

tests/unit/test_openeval.py
```diff
@@ -0,0 +1,342 @@
+"""Tests for trulens.connectors.openeval.
+
+These exercise to_openeval()/from_openeval() against DataFrames shaped
+exactly like Run.get_records() actually returns them (record_id, input,
+output, latency, plus feedback-score columns) rather than a live TruSession
+-- Run itself requires a live RunDaoBase/TruSession/app just to construct,
```
joshreini1, line 6: Module docstring claims Run.get_records() returns a fixed record_id/input/output/latency shape read directly out of trulens.core.run.Run.get_records() implementation. If that contract ever changes (e.g., a new reserved column), these tests will silently pass with the wrong shape. Consider adding one test that imports Run and introspects its actual column contract via inspect.signature or a real Run.get_records() call against a fixture session, so drift is caught at test time rather than in production.

joshreini1 (commented):
Clean rebase of the already-approved #2697 with the requested rename to trulens-connectors-openeval. The implementation follows TruLens conventions: proper import style, Google docstrings, sensible DataFrame boundary for portability, comprehensive test coverage (17 tests including round-trip validation against real EvalPort validators), and correct handling of TruLens-specific patterns like _calls companion columns. The missing poetry.lock update is documented with the exact expected diff and a reproducible Poetry-version caveat. Ready to merge once the lock is regenerated via poetry lock (Poetry 2.3.2, not 2.1.1).

adhabnr-ux:
Thanks for the thorough review, @joshreini1 — all three addressed in `a997c4e`:

**Version (2.12.0 → 2.13.1):** You're right, and I checked the pattern — `trulens-connectors-snowflake` tracks the root `trulens` package version rather than maintaining its own (currently `2.13.1` in `src/connectors/snowflake/pyproject.toml`, matching root). Bumped `trulens-connectors-openeval` to `2.13.1` to match. The `poetry.lock` diff below is updated for the new version.

**Redundant pandas guard:** Removed. You're right that it was dead code — `pandas` is an explicit dependency in this package's own `pyproject.toml`, so the `try/except ImportError` around `import pandas as pd` could never actually fire; `trulens-core`'s own pandas dependency was never the thing making it safe.

**Drift-detection test:** Added `test_run_get_records_reserved_columns_contract`, using your `inspect` suggestion specifically: it calls `inspect.getsource(Run.get_records)` and asserts the four reserved-column literals (`"record_id"`, `"input"`, `"output"`, `"latency"`) are still present in the real source, then cross-checks against this module's own `_RECORD_OVERVIEW_COLUMNS`. If TruLens ever changes that contract, this fails at test time instead of `to_openeval()` silently misclassifying a column in production. I went with reading the real source over a live `Run.get_records()` call since constructing a `Run` needs a live `RunDaoBase`/`TruSession`/app (as the module docstring explains) — this still exercises the actual current implementation, not a copy of it, without that overhead.

All 18 tests (17 + the new one) pass in a fresh venv against `trulens-core==2.13.1`:
```
18 passed in 1.73s
```

Updated `poetry.lock` diff (regenerated with Poetry 2.3.2 against the PR's current base, `fd2d2601`) — 41 lines, same two new package entries as before, now at the corrected version:

```diff
diff --git a/poetry.lock b/poetry.lock
index 48ddfdf2..9c2a287d 100644
--- a/poetry.lock
+++ b/poetry.lock
@@ -1925,6 +1925,26 @@ type = "legacy"
 url = "https://pypi.org/simple"
 reference = "pypi-public"
 
+[[package]]
+name = "evalport-sdk"
+version = "1.3.1"
+description = "Python SDK for EvalPort — The Open Evaluation Standard"
+optional = false
+python-versions = ">=3.8"
+groups = ["required"]
+files = [
+    {file = "evalport_sdk-1.3.1-py3-none-any.whl", hash = "sha256:1fba7d5e933da09c04ebc9479347c4b852b47e696824081911bd70360904a41a"},
+    {file = "evalport_sdk-1.3.1.tar.gz", hash = "sha256:4172c31450bed6b2ed73243bdea91d939eee0be099476f6c6411bd0a59aa4884"},
+]
+
+[package.extras]
+test = ["jsonschema", "pytest"]
+
+[package.source]
+type = "legacy"
+url = "https://pypi.org/simple"
+reference = "pypi-public"
+
 [[package]]
 name = "exceptiongroup"
 version = "1.3.1"
@@ -12235,6 +12255,25 @@ trulens-feedback = "^2.0.0"
 type = "directory"
 url = "src/benchmark"
 
+[[package]]
+name = "trulens-connectors-openeval"
+version = "2.13.1"
+description = "Convert between TruLens Run records and EvalPort (https://github.com/adhabnr-ux/evalport) suites and result sets."
+optional = false
+python-versions = "^3.9"
+groups = ["required"]
+files = []
+develop = true
+
+[package.dependencies]
+evalport-sdk = ">=1.0.0"
+pandas = ">=1.0.0"
+trulens-core = "^2.0.0"
+
+[package.source]
+type = "directory"
+url = "src/connectors/openeval"
+
 [[package]]
 name = "trulens-connectors-snowflake"
 version = "2.13.1"
@@ -13932,4 +13971,4 @@ otlp = ["opentelemetry-exporter-otlp-proto-grpc"]
 [metadata]
 lock-version = "2.1"
 python-versions = "^3.10"
-content-hash = "cb032d20571ebbf5b831cf9de3561b1b9a53572409fa4565eed38eac5a41cf8e"
+content-hash = "c524c0d7960fcd4030aceebee670c53d9e84cf52fb969f578b46d03beee71c64"
```

Reproducible with `poetry lock` on this branch as-is (confirmed against Poetry 2.3.2, per the version-mismatch note in the PR description).

adhabnr-ux:
@joshreini1 `poetry.lock` is regenerated and pushed (`8956288`), and I chased the remaining red check down to ground truth rather than leaving it unexplained.

**Status of the two checks that failed after the lock-file push, both now resolved/explained:**

1. ~~`Lint & format (pre-commit)`~~ — fixed in `8956288`. `ruff-format 0.16.3` (this repo's pinned version) wanted the new `test_run_get_records_reserved_columns_contract` assertion wrapped onto multiple lines; applied ruff's own output, reran the full local `pre-commit` hook set plus the 18-test `test_openeval.py` suite, both clean. ✅ now green.

2. `PR Validation Eval (PRBranchProtect py310-static)` — still red, but **not caused by this PR**. Root cause, traced through the Azure log for this specific job:

   - The failure is 26 failed + 2 collection errors, all in `tests/unit/test_otel_client_hooks.py`, `test_otel_client_hook_runs.py`, `test_otel_async_concurrency.py`, `test_otel_utils.py`, plus two `ModuleNotFoundError` collection errors in `test_otel_client_hook_plugins.py` (`trulens.apps.claude`) and `test_otel_streaming.py` (`openai`, not installed in the "basic" test-suite job).
   - The actual assertion failures all trace to the same root cause: `src/core/trulens/core/otel/client_hooks/tracing.py:196`, in `TraceAssembler.assemble()` → `_otel_id()`:
     ```python
     value = int.from_bytes(hashlib.sha256(seed.encode()).digest()[:size])
     ```
     `int.from_bytes()` only made `byteorder` default to `"big"` in **Python 3.11**; on 3.10 it's still a required positional arg, so every call raises `TypeError: from_bytes() missing required argument 'byteorder' (pos 2)`. That's exactly why `py311-static` and `py313-static` (and `default`) pass while `py310-static` alone fails.
   - This PR's diff (`poetry.lock`, `pyproject.toml`, `src/connectors/openeval/**`, `tests/unit/test_openeval.py`) never touches `src/core/trulens/core/otel/client_hooks/**`, and the same missing-`byteorder` bug is already present on `main` at that exact line as of this comment — so this is a pre-existing Python 3.10 incompatibility in the coding-agent client-hooks tracing feature, unrelated to the openeval connector.

So the connector itself (build, lint, format, unit tests, and 3 of the 4 Python-version test jobs) is fully green; the one remaining red is an upstream py310 bug outside this PR's scope. Happy to open a separate issue for the `_otel_id` fix if that's useful, otherwise leaving this here so the red X doesn't block review. Thanks again for the review — let me know if there's anything else you'd like changed.

adhabnr-ux:
Filed the pre-existing py310 `_otel_id()` bug separately as promised: #2764. Reproduced it on real Python 3.10.20 (not just inferred from the changelog) before filing, so it's a confirmed one-line fix (`int.from_bytes(..., byteorder="big")`), unrelated to this PR's diff. Nothing further needed from me here — this PR's own checks (build, lint, format, unit tests, py311/py313/default) are all green; just flagging so the remaining red `py310-static` X on this PR doesn't need chasing further, it's tracked separately now.

PR #2763, opened 2026-09-05
fix: skip invalid spans during trace compression

yhz5613813:
## Summary

Basic trace compression conditionally creates `compressed_span` only for dictionary spans, but appends it unconditionally. A non-mapping span before any valid span raises `UnboundLocalError`; later invalid spans can duplicate stale output.

Skip non-dictionary values before constructing and appending the compressed span.

## Testing

- Added regression coverage with `None`, a valid span, and a string span.
- Verified the changed compression method returns only the valid span.
- Ruff check, formatting check, and `compileall` on both changed files.

### DIFF
--- a/src/core/trulens/core/utils/trace_compression.py
+++ b/src/core/trulens/core/utils/trace_compression.py
@@ -297,7 +297,7 @@ class TraceCompressor:
                         if important_attrs:
                             compressed_span["span_attributes"] = important_attrs
 
-            compressed_spans.append(compressed_span)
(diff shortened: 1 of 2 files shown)


joshreini1 (approved):
Clean bug fix that correctly scopes the append inside the isinstance(span, dict) guard, preventing UnboundLocalError when non-mapping spans appear before valid ones and avoiding duplicate stale outputs from later invalid spans. The new regression test covers None, valid dict, and string spans. Ready to merge.

PR #2765, opened 2026-09-05
fix(database): extract cost_json fields via json_extract in pre-OTEL leaderboard aggregates

connectsudhindra-gif:
## Summary
- `get_leaderboard_aggregates` always raised `NotImplementedError` when `TRULENS_OTEL_TRACING` is disabled, because `_get_leaderboard_aggregates_pre_otel` indexed `Record.cost_json["n_tokens"]` / `["cost"]` with `[]`, but `cost_json` is declared as `Text`, not `JSON`.
- Added a dialect-aware `_json_path_expr_from_text()` helper that extracts a field from a `Text` column holding a JSON-encoded string (casting to `JSON` first on Postgres, since `json_extract_path_text` requires a JSON-typed argument there), and used it in place of the broken `[]` indexing.

Fixes #2729

## Test plan
- [x] Added `tests/unit/test_leaderboard_aggregates_pre_otel.py`, which runs a `DummyApp` with `TRULENS_OTEL_TRACING=0` and asserts `get_leaderboard_aggregates` returns a populated DataFrame instead of raising.
- [x] Verified the fixed expression against SQLite via a standalone SQLAlchemy script, and confirmed it compiles correctly for the Postgres dialect.
- [ ] CI run on this PR

Issue #2759, opened 2026-09-03
Leaderboard: 'Total Tokens' is avg not sum; mixed currencies summed as one number


roy-tong:
Two leaderboard aggregation issues found while auditing eval-framework measurement semantics.

**1. "Total Tokens" is an average, not a sum**

`src/core/trulens/core/database/sqlalchemy.py` (~L1727-1729, pre-OTel leaderboard):
```python
sa.func.avg(cost_json["n_tokens"])
```
is labeled "Total Tokens". The column shows mean tokens per record, not a total. Compare "Total Cost (USD)" in the same query which correctly uses `sum()`. Anyone reading "Total Tokens: 1,247" would conclude the app consumed 1,247 tokens total, when it's actually the average per feedback call.

**2. Mixed-currency costs are summed as one number**

The OTel leaderboard variant (~L1225-1228) computes `total_cost = sum(cost)` with `coalesce(max(cost_currency), 'USD')`. For an app whose records carry mixed currencies (the `Cost` model supports `cost_currency`), amounts in EUR and USD are summed together and displayed under whichever currency happens to sort last alphabetically. This bypasses the explicit guard in `Cost.__add__` (`src/core/trulens/core/schema/base.py` ~L118-125) that refuses cross-currency addition.

**Suggestions**

- Rename the column "Average Tokens" or use `sum()` — either fix is one line
- Aggregate cost per currency (GROUP BY cost_currency) or normalize to a base currency before summing

Context: AgentMeasure conformance check (30 projects / 30 days); TruLens is on the list and these are the two clearest leaderboard findings. Happy to send a repro with mixed-currency Cost objects if useful.

roy-tong:
Both findings from this issue now have fix PRs in flight — thanks to Abelo9996 for #2766 (per-currency aggregation) and connectsudhindra-gif for #2765 (json_extract on the pre-OTel path). The mixed-currency repro from this thread maps directly onto #2766's tests if useful; happy to review either PR.

modelpath-dev:
I will take this issue. Please assign it to me.

The first problem is in `src/core/trulens/core/database/sqlalchemy.py`. The "Total Tokens" column uses `avg()` instead of `sum()`. I would change it to `sum()` to correctly represent the total tokens. 

The second issue is with mixed-currency costs in the OTel leaderboard. I would look into grouping costs by `cost_currency` or converting them to a single currency before summing. This would prevent incorrect totals when different currencies are involved. 

[...]

PR #2766, opened 2026-09-05
fix(database): report leaderboard cost per currency, not summed across currencies

Abelo9996:
Addresses the mixed-currency half of #2759.

## Problem

The OTel leaderboard summed every record's cost into one `total_cost = sum(cost)` and labelled it with `coalesce(max(cost_currency), "USD")`, then in post-processing attributed the whole sum to that single currency. For an app whose records carry mixed currencies (the `Cost` model supports `cost_currency`), amounts in USD and in Snowflake credits are added together and reported under whichever currency sorts last. This is exactly the cross-currency addition that `Cost.__add__` refuses to do.

## Fix

Sum cost per currency directly in SQL with a conditional sum for each supported currency, producing `Total Cost (USD)` and `Total Cost (Snowflake Credits)` as separate aggregates. Single-currency apps are unaffected; only the mixed-currency case changes (from one mislabelled total to two correct per-currency totals). The returned columns are unchanged, and the now-redundant post-processing split is removed.

## Tests

Adds `tests/unit/test_leaderboard_mixed_currency.py`: an app with one USD record and one Snowflake-credits record asserts the two totals are reported separately, not summed. It fails on `main` (the USD column shows the combined total) and passes with this change. Lint and format clean under the pinned ruff.

## Note

This covers the mixed-currency issue in #2759. The other half of that issue (pre-OTel "Total Tokens" being an average rather than a sum) lives in `_get_leaderboard_aggregates_pre_otel`, which #2765 is currently rewriting, so it is best addressed there to avoid a conflicting edit.

### DIFF
--- a/src/core/trulens/core/database/sqlalchemy.py
+++ b/src/core/trulens/core/database/sqlalchemy.py
@@ -1222,10 +1222,28 @@ class SQLAlchemyDB(core_db.DB):
                     "Latest Record Timestamp"
                 ),
                 latency_expr.label("Average Latency (s)"),
-                sa.func.sum(sa.cast(cost_col, sa.Float)).label("total_cost"),
-                sa.func.coalesce(
-                    sa.func.max(currency_col), sa.literal("USD")
-                ).label("cost_currency"),
+                # Sum cost per currency rather than summing every record's cost
+                # into one number and labelling it with whichever currency
+                # sorts last. Mixing currencies in a single total is wrong and
+                # is what Cost.__add__ refuses to do.
+                sa.func.sum(
+                    sa.case(
+                        (
+                            currency_col == sa.literal("USD"),
+                            sa.cast(cost_col, sa.Float),
+                        ),
+                        else_=0.0,
+                    )
+                ).label("Total Cost (USD)"),
+                sa.func.sum(
+                    sa.case(
+                        (
+                            currency_col == sa.literal("Snowflake credits"),
+                            sa.cast(cost_col, sa.Float),
+                        ),
+                        else_=0.0,
+                    )
+                ).label("Total Cost (Snowflake Credits)"),
                 sa.func.sum(
                     sa.cast(
                         sa.func.coalesce(tokens_col, sa.literal("0")),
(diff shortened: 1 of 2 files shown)


connectsudhindra-gif:
@Abelo9996 its already fixed in https://github.com/truera/trulens/pull/2765

Abelo9996:
Thanks @connectsudhindra-gif, but I think these are different code paths. #2765 changes `_get_leaderboard_aggregates_pre_otel` (the pre-OTel path, used when OTEL tracing is disabled), and its diff is entirely inside that function. This PR fixes the OTel path in `get_leaderboard_aggregates`, around the `sum(cast(cost_col, Float))` + `coalesce(max(cost_currency), "USD")` aggregation, where records in USD and in Snowflake credits are summed together and attributed to one currency. #2765 does not touch that path, so the mixed-currency bug there is still present after it merges.

Separately, #2765 keeps "Total Tokens" as a mean (`("total_tokens", "mean")`) while labelling it a total, which is the other half of #2759. I left a note on #2765 about folding that one-line change in, since that function is being rewritten there.

src/core/trulens/core/database/sqlalchemy.py
```diff
@@ -1222,10 +1222,28 @@ def _get_leaderboard_aggregates_otel(
...
+                # Sum cost per currency rather than summing every record's cost
+                # into one number and labelling it with whichever currency
+                # sorts last. Mixing currencies in a single total is wrong and
+                # is what Cost.__add__ refuses to do.
+                sa.func.sum(
+                    sa.case(
+                        (
+                            currency_col == sa.literal("USD"),
+                            sa.cast(cost_col, sa.Float),
+                        ),
```
joshreini1, line 1234: Extract the literal "USD" (line 1234) and "Snowflake credits" (line 1243) to module-level constants like CURRENCY_USD and CURRENCY_SNOWFLAKE_CREDITS. This makes the supported currency set explicit and easier to extend if other currencies are added later, though non-blocking for this fix.

joshreini1 (approved):
The fix correctly addresses the mixed-currency summation bug by computing per-currency totals directly in SQL rather than post-processing a single sum. The SQL change is clean and self-documenting, the removed pandas lambda code simplifies the flow, and the test is focused on the regression. One optional NIT about extracting hardcoded currency strings to constants. Verdict: Approve — ready to merge.

PR #2767, opened 2026-09-08
fix(dashboard): coerce Records search columns to string before .str.contains

### LINKED ISSUE #1749: [BUG] String type error in trulens dashboard using `add_dataframe`
**Bug Description**
 I’ve created a feature branch of our production code and added the logic that utilizes the add_dataframe method to submit data to TruLens for evaluation.  It completes without error, but when I go open the dashboard for it using the run_dashboard option, the Leaderboard displays and indicates the correct number of records, but it doesn’t show any feedback columns/scores (just the version, number of records and the average latency).  Then if I try going to the Records option, I get this error:

```
File "C:\GitLab\eric.weist\data-science\.trulens.1.3.2.conda\Lib\site-packages\trulens\dashboard\pages\Records.py", line 607, in <module>

    records_main()

File "C:\GitLab\eric.weist\data-science\.trulens.1.3.2.conda\Lib\site-packages\trulens\dashboard\pages\Records.py", line 603, in records_main

    render_records(app_name)

File "C:\GitLab\user\data-science\.trulens.1.3.2.conda\Lib\site-packages\trulens\dashboard\pages\Records.py", line 549, in render_records

    df = _preprocess_df(

File "C:\GitLab\user\data-science\.trulens.1.3.2.conda\Lib\site-packages\trulens\dashboard\pages\Records.py", line 174, in _preprocess_df

    records_df["output"].str.encode("utf-8").str.decode("unicode-escape")

File "C:\GitLab\user\data-science\.trulens.1.3.2.conda\lib\site-packages\pandas\core\generic.py", line 6299, in __getattr__

    return object.__getattribute__(self, name)

File "C:\GitLab\user\data-science\.trulens.1.3.2.conda\lib\site-packages\pandas\core\accessor.py", line 224, in __get__

    accessor_obj = self._accessor(obj)

File "C:\GitLab\user\data-science\.trulens.1.3.2.conda\lib\site-packages\pandas\core\strings\accessor.py", line 191, in __init__

    self._inferred_dtype = self._validate(data)

File "C:\GitLab\user\data-science\.trulens.1.3.2.conda\lib\site-packages\pandas\core\strings\accessor.py", line 245, in _validate

    raise AttributeError("Can only use .str accessor with string values!")
```

**To Reproduce**
Which steps should someone take to run into the same error? A small, reproducible code example is useful here.

**Expected behavior**

**Relevant Logs/Tracebacks**
Please copy and paste any relevant log output. This will be automatically formatted into code, so no need for backticks. If the issue is related to the TruLens dashboard, please also include a screenshot.

**Environment:**

- OS: [e.g. MacOS, Windows]
- Python Version
- TruLens version
- Versions of other relevant installed libraries

**Additional context**

connectsudhindra-gif:
## Summary
- The Records page search box calls `.str.contains` directly on the `app_version`, `input`, and `output` columns in `_preprocess_df`. When any of those columns arrives as a non-string dtype - an all-NaN `output` column (e.g. from SQL NULLs, which comes back `float64`), or a numeric `app_version` - pandas raises `AttributeError: Can only use .str accessor with string values`, crashing the page as soon as a search query is entered.
- Added a small `_searchable()` helper that coerces each column with `.fillna("").astype(str)` before matching, so the search box works regardless of the incoming dtype. This matches the coerce-then-match idiom already used elsewhere in this file (`.astype(str).str.len()` in `_has_conversations` / `_thread_membership`).

Fixes #1749

## Test plan
- [x] Added `tests/unit/streamlit/test_streamlit_records.py::test_preprocess_search_query_survives_nan_output_column`, which builds a `float64` all-NaN `output` column and confirms the search filter runs without raising and returns the correct match.
- [x] Added `tests/unit/streamlit/test_streamlit_records.py::test_preprocess_search_query_survives_numeric_app_version`, covering a numeric `app_version` column the same way.
- [x] Verified the underlying pandas behavior (crash before the fix, correct filtering after) with a standalone repro against the exact operations used in `_preprocess_df`.
- [ ] CI run on this PR

### DIFF
--- a/tests/unit/streamlit/test_streamlit_records.py
+++ b/tests/unit/streamlit/test_streamlit_records.py
@@ -1,3 +1,4 @@
+import numpy as np
 import pandas as pd
 from trulens.dashboard.tabs.Records import _build_thread_grid_options
 from trulens.dashboard.tabs.Records import _build_thread_summary
@@ -48,6 +49,27 @@ def test_preprocess_filters_skipped_records():
     result = _preprocess_df(_records(), online_eval_filter="Skipped")
 
     assert set(result["record_id"]) == {"record-1", "record-2"}
+
+
+def test_preprocess_search_query_survives_nan_output_column():
+    # A column of all-missing outputs (e.g. SQL NULLs) round-trips as
+    # float64, not object/string, which used to crash the search box's
+    # `.str.contains` call with an AttributeError.
+    df = _records()
+    df["output"] = pd.Series([np.nan] * 5, index=df.index, dtype="float64")
+
+    result = _preprocess_df(df, record_query="input 2")
+
+    assert set(result["record_id"]) == {"record-2"}
+
+
+def test_preprocess_search_query_survives_numeric_app_version():
+    df = _records()
+    df["app_version"] = [1, 1, 1, 1, 1]
+
+    result = _preprocess_df(df, record_query="input 3")
+
+    assert set(result["record_id"]) == {"record-3"}
 
 
 def test_conversation_bubbles_use_theme_colors():
--- a/src/dashboard/trulens/dashboard/tabs/Records.py
+++ b/src/dashboard/trulens/dashboard/tabs/Records.py
@@ -425,10 +425,18 @@ def _preprocess_df(
     #     records_df = records_df.drop(columns="app_json")
 
     if record_query:
+
+        def _searchable(col: str) -> pd.Series:
+            # Columns can arrive as non-string dtypes (e.g. all-NaN output
+            # columns come back float64, app_version can be numeric), and
+            # `.str` accessors raise AttributeError on those. Coerce to
+            # string first so the search box never crashes on such data.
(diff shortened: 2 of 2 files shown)


tests/unit/streamlit/test_streamlit_records.py
```diff
@@ -1,3 +1,4 @@
+import numpy as np
```
joshreini1, line 1: The `numpy` import is only used for `np.nan` in test fixtures; pandas has `pd.NA` or you could use `None` with explicit dtype. Non-blocking—`np.nan` is idiomatic for reproducing the exact float64-NaN data type that triggers the bug.

joshreini1 (approved):
Clean bug fix for a real crash. The `_searchable()` helper correctly coerces columns to string before `.str.contains`, matching the existing idiom used elsewhere in the file. Two well-scoped unit tests cover both the NaN and numeric cases. The import of `numpy` is needed for the NaN test fixture. Ready to merge.

PR #2771, opened 2026-09-10
fix(database): make re-evaluated record score deterministic (latest wins) (#2768)

### LINKED ISSUE #2768: [BUG] Evaluation history scores do not match individual metric scores
**Bug Description**
The evaluation scores displayed in the table do not consistently match the corresponding scores shown in the detailed evaluation results below the table.

Observed inconsistency on Context Relevance and Groundedness, maybe other metrics as well.

A clear example is Context Relevance:

On the first run, the table shows 0.67, which matches the Context Relevance score shown in the detailed result and its explanation.
After making an update and running the evaluation again, the latest/fresh result appears correct and still shows 0.67.
However, the previously recorded result in the table changes to 1.00, while the detailed result below still shows 0.67.

This causes the historical evaluation score to become inconsistent with the actual evaluation result.

**To Reproduce**
Run an evaluation for the first time.
Check the Context Relevance result.
Verify that:
Table score = 0.67
Detailed score below = 0.67
Make an update to the evaluation/configuration.
Run the evaluation again.
Verify that the latest/fresh result still shows the expected score of 0.67.
Check the previous result in the table.
Observe that the historical Context Relevance score has changed to 1.00, while the detailed result remains 0.67.

Image

Image

Image

The same type of inconsistency can also occur for Groundedness.

**Expected Behavior**
<p data-start="1538" data-end="1640" class="PDq2pG_selectionAnchorContainer">Historical evaluation results should remain <strong data-start="1582" data-end="1595">unchanged</strong> after subsequent evaluation runs or updates.<span aria-hidden="true" class="PDq2pG_selectionAnchor"></span></p>
<p data-start="1642" data-end="1752">The score shown in the table should match the score stored/displayed for that specific evaluation run.</p>
<p data-start="1754" data-end="1766">For example:</p>
<div class="group TyagGW_tableContainer"><div tabindex="-1" class="TyagGW_tableWrapper flex flex-col-reverse w-fit">
Result | Table | Detailed result
-- | -- | --
First run | 0.67 | 0.67
After second run | Should remain 0.67 | 0.67

</div></div>
<p data-start="1911" data-end="1998">Subsequent runs should not modify the scores of previously recorded evaluation results.</p>

**Relevant Logs/Tracebacks**
No error/traceback observed.

**Environment:**

- OS: MacOS
- Python Version: 3.10
- TruLens version: 3.14

**Additional Context**

Abelo9996:
Closes #2768

## Problem

A previously recorded evaluation score in the records table changes after a later evaluation run, and diverges from the detailed result for the same record (e.g. Context Relevance shows 0.67 in the detail but 1.00 in the table after a re-run).

When a metric is re-evaluated on the same record, there are several `EVAL_ROOT` spans for one `(record, metric)`. `_get_records_and_feedback_otel_from_events` sets the record's score with `feedback_result["mean_score"] = eval_root_score` for every `EVAL_ROOT` span it iterates, so the reported score is whichever span is processed last. The events have no defined order, so the table score is nondeterministic and can differ from the detail. This is the case the existing `# TODO(otel): This isn't going to work if there are multiple with the same name.` flags.

## Fix

Keep the latest `EVAL_ROOT` by timestamp for each `(record, metric)`, so the score is deterministic and reflects the most recent evaluation. The per-call breakdown and eval cost still accumulate across all eval spans as before.

## Tests

Adds `tests/unit/test_records_reevaluation_score.py`: two `EVAL_ROOT` spans (scores 0.67 then 1.00 at different timestamps) inserted in both orders, asserting the table score is the latest (1.00) regardless of insertion order. On `main` one order returns 0.67 and the test fails; with this change both orders return 1.00. Lint and format clean under the pinned ruff.

### DIFF
--- a/src/core/trulens/core/database/base.py
+++ b/src/core/trulens/core/database/base.py
@@ -1036,13 +1036,22 @@ class DB(serial_utils.SerialModel, abc.ABC, text_utils.WithIdentString):
                         record_attributes.get(SpanAttributes.SPAN_TYPE)
                         == SpanAttributes.SpanType.EVAL_ROOT.value
                     ):
-                        # NOTE: EVAL_ROOT.SCORE should provide the mean score of all related EVAL spans
-                        feedback_result["mean_score"] = eval_root_score
-                        # TODO(SNOW-2112879): HIGHER_IS_BETTER has not been populated in the OTEL spans yet
-                        feedback_result["direction"] = record_attributes.get(
-                            SpanAttributes.EVAL_ROOT.HIGHER_IS_BETTER,
-                            None,
-                        )
+                        # A record and metric can have several EVAL_ROOT spans
+                        # when the metric is re-evaluated. Keep the latest one by
+                        # timestamp so the reported score is deterministic rather
+                        # than whichever span happens to be iterated last.
+                        prev_ts = feedback_result.get("_score_ts")
+                        if prev_ts is None or event.start_timestamp >= prev_ts:
+                            feedback_result["_score_ts"] = event.start_timestamp
+                            # NOTE: EVAL_ROOT.SCORE should provide the mean score of all related EVAL spans
+                            feedback_result["mean_score"] = eval_root_score
+                            # TODO(SNOW-2112879): HIGHER_IS_BETTER has not been populated in the OTEL spans yet
+                            feedback_result["direction"] = (
+                                record_attributes.get(
+                                    SpanAttributes.EVAL_ROOT.HIGHER_IS_BETTER,
+                                    None,
+                                )
+                            )
                         # Add call data for EVAL_ROOT spans
                         args_span_id = self._extract_namespaced_attributes(
                             record_attributes,
--- a/src/core/trulens/core/database/base.py
+++ b/src/core/trulens/core/database/base.py
@@ -1190,6 +1190,9 @@ class DB(serial_utils.SerialModel, abc.ABC, text_utils.WithIdentString):
             for feedback_name, feedback_result in record_data[
                 "feedback_results"
             ].items():
+                # Drop the internal latest-score timestamp used only to pick the
+                # most recent EVAL_ROOT; it is not part of the public schema.
+                feedback_result.pop("_score_ts", None)
                 # NOTE: we use the mean score as the feedback result
                 record_row[feedback_name] = feedback_result["mean_score"]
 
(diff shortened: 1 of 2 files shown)


src/core/trulens/core/database/base.py
```diff
@@ -1036,13 +1036,22 @@ def _get_records_and_feedback_otel_from_events(
...
-                        feedback_result["direction"] = record_attributes.get(
-                            SpanAttributes.EVAL_ROOT.HIGHER_IS_BETTER,
-                            None,
-                        )
+                        # A record and metric can have several EVAL_ROOT spans
+                        # when the metric is re-evaluated. Keep the latest one by
+                        # timestamp so the reported score is deterministic rather
+                        # than whichever span happens to be iterated last.
+                        prev_ts = feedback_result.get("_score_ts")
+                        if prev_ts is None or event.start_timestamp >= prev_ts:
```
joshreini1, line 1044: Remove `_score_ts` from `feedback_result` after the event loop completes, before returning from `_get_records_and_feedback_otel_from_events`. This internal field leaks into the returned dict and pollutes the schema. Add cleanup after processing all events for each feedback_key: `feedback_result.pop('_score_ts', None)`.

tests/unit/test_records_reevaluation_score.py
```diff
@@ -0,0 +1,108 @@
...
+            "old": _eval_root(
+                "r1", "Context Relevance", 0.67, "2026-01-01T00:00:05", "old"
+            ),
+            "new": _eval_root(
+                "r1", "Context Relevance", 1.0, "2026-01-02T00:00:05", "new"
+            ),
+        }
+        for key in order:
+            db.insert_event(events[key])
+        df, _ = db.get_records_and_feedback()
```
joshreini1, line 95: Consider extracting the assertion `self.assertEqual(score, 1.0)` from both test calls so each order is a separate named test method (`test_score_latest_when_old_first`, `test_score_latest_when_new_first`). Separate test methods produce clearer failure reports when one order breaks but not the other, though the current approach is acceptable for a regression test.

joshreini1 (changes requested):
The fix correctly addresses the nondeterministic score issue by keeping the latest EVAL_ROOT span by timestamp. The test design is solid and covers both insertion orders. Request changes: 1 blocker (internal `_score_ts` field must be cleaned up before returning to avoid polluting the public schema with implementation details).

Abelo9996:
Addressed both: dropped the internal `_score_ts` from `feedback_result` before it is returned, and split the re-evaluation test into `test_score_latest_when_old_first` and `test_score_latest_when_new_first`.

Abelo9996:
recheck

joshreini1 (approved):

sfc-gh-jreini (approved):

PR #2772, opened 2026-09-10
fix(otel): pass byteorder to int.from_bytes in _otel_id for Python 3.10 (#2764)

### LINKED ISSUE #2764: TraceAssembler._otel_id() breaks on Python 3.10: int.from_bytes() missing required 'byteorder' argument
**Bug description**

`_otel_id()` in `src/core/trulens/core/otel/client_hooks/tracing.py` calls `int.from_bytes()` without the `byteorder` argument:

```python
def _otel_id(seed: str, bits: int) -> int:
    size = bits // 8
    value = int.from_bytes(hashlib.sha256(seed.encode()).digest()[:size])
    return value or 1
```

`byteorder` only became optional (defaulting to `"big"`) in Python 3.11 ([bytes.org changelog](https://docs.python.org/3/whatsnew/3.11.html) / [bpo-45391](https://bugs.python.org/issue45391)). On Python 3.10 it's still a required positional argument, so every call to `_otel_id()` — and therefore `TraceAssembler.assemble()`, which calls it for every span id — raises `TypeError` on 3.10.

**Reproduction** (verified on real Python 3.10.20, not just read from the changelog):

```
$ python3.10 -c "
import hashlib
seed = 'trace:test'
size = 16
value = int.from_bytes(hashlib.sha256(seed.encode()).digest()[:size])
print(value)
"
TypeError: from_bytes() missing required argument 'byteorder' (pos 2)
```

Same code on 3.11+ works because of the new default.

**Where I ran into it**: this repo's own CI. On [PR #2757](https://github.com/truera/trulens/pull/2757) (unrelated — an `openeval` connector under `src/connectors/`), the `PR Validation Eval (PRBranchProtect py310-static)` job fails with 26 test failures + 2 collection errors, all tracing back to this exact `TypeError` at `tracing.py:196` (`TraceAssembler.assemble()` → `_otel_id()`). `py311-static`, `py313-static`, and `default` all pass on the same commit, which is consistent with the 3.10-vs-3.11 `byteorder` default change and nothing else. I confirmed this reproduces identically on current `main` (no relation to that PR's diff, which never touches this file).

**Fix**: pass `byteorder` explicitly, e.g. `int.from_bytes(hashlib.sha256(seed.encode()).digest()[:size], byteorder="big")` (matching the value 3.11+ defaults to) — a one-line, backward-compatible change on every supported Python version.

**Environment**: reproduced on Python 3.10.20; also present as of `main` at the time of filing (confirmed by reading the current file content).

I don't have write access to this repo so I can't open a PR directly, but happy to if that's useful — it's a genuinely one-line fix.

Abelo9996:
Closes #2764

## Problem

`_otel_id()` in `src/core/trulens/core/otel/client_hooks/tracing.py` calls `int.from_bytes()` without `byteorder`. `byteorder` only became optional (defaulting to `"big"`) in Python 3.11; on Python 3.10 it is a required positional argument. So every id derivation, and therefore `TraceAssembler.assemble()` (which calls `_otel_id` for every span id), raises `TypeError` on Python 3.10, which `pyproject.toml` still supports.

## Fix

Pass `byteorder="big"` explicitly. `"big"` matches the 3.11+ default, so ids are unchanged on newer versions and the function stops crashing on 3.10.

## Tests

Adds `tests/unit/test_otel_id_byteorder.py`, which pins the big-endian interpretation, checks the function returns an `int` without raising (the 3.10 failure path), and that the result is never zero. Lint and format clean under the pinned ruff.

Thanks to @adhabnr-ux for the report and @chrikrah for the 3.10 reproduction.

### DIFF
--- a/tests/unit/test_otel_id_byteorder.py
+++ b/tests/unit/test_otel_id_byteorder.py
@@ -1,7 +1,7 @@
 """Regression test for _otel_id byte order and Python 3.10 compatibility.
 
 int.from_bytes() requires byteorder on Python 3.10 (it only became optional,
-defaulting to "big", in 3.11). _otel_id() omitted it, so every id derivation,
+defaulting to "big", in 3.11). tracing._otel_id() omitted it, so every id derivation,
 and therefore TraceAssembler.assemble(), raised TypeError on 3.10. The explicit
 byteorder must be "big" so ids are unchanged on 3.11+.
 """
@@ -9,7 +9,7 @@ byteorder must be "big" so ids are unchanged on 3.11+.
 import hashlib
 import unittest
 
-from trulens.core.otel.client_hooks.tracing import _otel_id
+from trulens.core.otel.client_hooks import tracing
 
 
 class TestOtelIdByteOrder(unittest.TestCase):
@@ -19,15 +19,15 @@ class TestOtelIdByteOrder(unittest.TestCase):
         expected = int.from_bytes(
             hashlib.sha256(seed.encode()).digest()[:size], byteorder="big"
         )
-        self.assertEqual(_otel_id(seed, bits), expected)
+        self.assertEqual(tracing._otel_id(seed, bits), expected)
 
     def test_returns_int_without_raising(self):
         # Would raise TypeError on Python 3.10 before the fix.
-        self.assertIsInstance(_otel_id("span:xyz", 64), int)
-        self.assertEqual(_otel_id("span:xyz", 64), 15422170499572505575)
+        self.assertIsInstance(tracing._otel_id("span:xyz", 64), int)
+        self.assertEqual(tracing._otel_id("span:xyz", 64), 15422170499572505575)
 
     def test_never_zero(self):
-        self.assertGreaterEqual(_otel_id("anything", 64), 1)
+        self.assertGreaterEqual(tracing._otel_id("anything", 64), 1)
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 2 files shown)


tests/unit/test_otel_id_byteorder.py
```diff
@@ -0,0 +1,34 @@
...
+
+int.from_bytes() requires byteorder on Python 3.10 (it only became optional,
+defaulting to "big", in 3.11). _otel_id() omitted it, so every id derivation,
+and therefore TraceAssembler.assemble(), raised TypeError on 3.10. The explicit
+byteorder must be "big" so ids are unchanged on 3.11+.
+"""
+
+import hashlib
+import unittest
+
```
joshreini1, line 11: Change `from trulens.core.otel.client_hooks.tracing import _otel_id` to `from trulens.core.otel.client_hooks import tracing` and use `tracing._otel_id(...)` in tests. AGENTS.md requires importing modules, not functions/classes directly, even in tests—this keeps import patterns consistent across the codebase.

joshreini1 (commented):
Clean fix for a real Python 3.10 compatibility bug. The explicit `byteorder="big"` argument is correct and matches the 3.11+ default, keeping IDs stable across versions. Test coverage is solid: big-endian interpretation validated, TypeError regression prevented, and zero-fallback behavior confirmed. One import style issue to align with AGENTS.md conventions. Verdict: Comment: 1 should-fix (import style).

joshreini1:
Please read the [TruLens Contributor License Agreement](https://github.com/truera/trulens/blob/main/CLA.md), then post a new comment containing exactly:

```
```

The checked TruLens standards item is a separate certification and does not sign the CLA.

sfc-gh-jreini (approved):

Abelo9996:

Abelo9996:
Addressed: import the `tracing` module instead of `_otel_id` directly, per AGENTS.md.

PR #2787, opened 2026-09-12
fix(providers/ollama): treat blank env as unconfigured

sbguangha:
# Description

Blank `OLLAMA_BASE_URL` / `OLLAMA_API_KEY` values (typical leftover `KEY=` lines in a `.env` template) were treated as configured. That replaced the documented local default URL with an empty server address, and left the OpenAI client with an empty API key so `Ollama()` failed to construct even though Ollama does not require a key.

`os.environ.get("OLLAMA_BASE_URL", default)` returns `""` when the variable is present but empty, so the default never applied. The constructor now treats a blank env value the same as unset (`or` fallback). An explicit `base_url=""` / `api_key=""` argument is unchanged.

## Type of change

  not work as expected)
- [x] New Tests

## Certification

### DIFF
--- a/src/providers/ollama/trulens/providers/ollama/provider.py
+++ b/src/providers/ollama/trulens/providers/ollama/provider.py
@@ -140,12 +140,19 @@ class Ollama(openai_provider.OpenAI):
             model_engine = self.DEFAULT_MODEL_ENGINE
 
         if base_url is None:
-            base_url = os.environ.get("OLLAMA_BASE_URL", self.DEFAULT_BASE_URL)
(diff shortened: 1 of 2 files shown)


sbguangha:

joshreini1 (approved):
Clean fix for a real bug. The `or` fallback correctly treats blank env values as unset, restoring the documented local default behavior. The new test covers the regression path and mirrors the existing test structure. Comments explain the rationale clearly. **Verdict: Approved — ready to merge.**

sfc-gh-jreini (approved):

PR #2790, opened 2026-09-15
docs: TRULENS_OTLP_ENDPOINT and TRULENS_DATABASE_PATH are read but documented nowhere

xizhuomengcontin:
## What

This module is configured entirely by seven environment variables and names none of them in its docstring. Two of the seven appear nowhere in the repository outside the source.

## The two gaps

| Variable | Read at | Mentioned in docs |
| --- | --- | --- |
| `TRULENS_OTLP_ENDPOINT` | `exporting.py:62` | **nowhere** |
| `TRULENS_DATABASE_PATH` | `exporting.py:19` | **nowhere** |

The other five are documented somewhere in the repo: `TRULENS_DESTINATION` and `TRULENS_DATABASE_URL` in three files each, `TRULENS_SNOWFLAKE_CONNECTION`, `TRULENS_SNOWFLAKE_DATABASE` and `TRULENS_SNOWFLAKE_SCHEMA` in one each. So the module is mostly documented and these two were missed, rather than the whole surface being undocumented.

## Why `TRULENS_OTLP_ENDPOINT` is the one that costs time

```python
if destination == "otlp":
    endpoint = os.environ.get("TRULENS_OTLP_ENDPOINT")
    return core_session.TruSession(otel_exporter="otlp", otlp_endpoint=endpoint)
```

`os.environ.get` with no default, passed straight through. Set `TRULENS_DESTINATION=otlp` and misspell the endpoint variable — or never learn it exists — and nothing raises. `otlp_endpoint=None` lets the exporter use its own default, so spans go somewhere, just not where you meant. Compare `_snowflake_session`, which raises `"Set TRULENS_SNOWFLAKE_CONNECTION for Snowflake export."` when its variable is missing: the Snowflake path tells you, the OTLP path does not.

I have not changed that behaviour — defaulting is reasonable, and raising instead would break anyone relying on the exporter's own configuration. The docstring now says it plainly instead.

`TRULENS_DATABASE_PATH` is the quieter of the two: it decides where the SQLite file lands and the default is under `~/.trulens/`, which is easy to miss when wondering where the local data went.

## Scope

Module docstring only. No code, no defaults, no behaviour. Every statement is read off this file — the `0700` mode, the `~/.trulens/client-hooks.sqlite` default, the required-vs-optional split on the Snowflake path, and the `ValueError` on an unrecognised destination.


Abelo9996 (commented):
Accurate docs change that closes a real gap. I checked each variable against `exporting.py` and the defaults, the Snowflake vars, and the silent `TRULENS_OTLP_ENDPOINT` fallback are all described correctly.

- **should-fix: coordinate with #2785 (already approved).** It adds an `ai_gateway` destination and new vars (`TRULENS_AI_GATEWAY_URL`, `..._PAT_FILE` or `..._TOKEN`, `TRULENS_OTLP_PROTOCOL`) to this same docstring. If this lands after #2785 without a rebase, the new docstring ships stale: no `ai_gateway`, the "any other value raises" line becomes false, and those vars stay undocumented, which is the exact gap this PR closes. Rebase and fold them in.
- **NIT: `0o700` is applied to the parent directory (`~/.trulens`) via `mkdir(mode=...)`, not the SQLite file.** Reword to "the parent directory is created mode 0700."

Verdict: Comment. 1 should-fix (rebase on #2785), 1 nit. Not blocking.

src/core/trulens/core/otel/client_hooks/exporting.py
```diff
@@ -1,4 +1,19 @@
-"""Destination configuration and span export for coding-agent hooks."""
+"""Destination configuration and span export for coding-agent hooks.
+
+Everything here is configured through the environment; there are no arguments to pass.
+
+`TRULENS_DESTINATION` selects the destination and defaults to `local`:
+
+- `local` or `database` -- `TRULENS_DATABASE_URL` if set, otherwise a SQLite file at
+  `TRULENS_DATABASE_PATH` (default `~/.trulens/client-hooks.sqlite`, created mode 0700).
```
sfc-gh-jreini, line 8: Change `mode 0700` to `mode 0o700` to match Python's octal literal syntax (as written at `exporting.py:24`). Both notations are understood in Unix contexts, but `0o700` is more precise when documenting Python code.

sfc-gh-jreini (approved):
Clean documentation addition that fills a real gap. The seven environment variables are now clearly documented with accurate defaults and behavior. All claims verified against the code: the local default at line 56, the ~/.trulens/client-hooks.sqlite path at line 21, the 0o700 mode at line 24, the required vs optional Snowflake variables at lines 36-47, the OTLP silent-fallback behavior at line 62, and the ValueError at line 66. The note about OTLP typos causing silent misbehavior is particularly valuable. Verdict: Ready to merge.

xizhuomengcontin:

PR #2791, opened 2026-09-15
feat(core): instrument official MCP client calls

### LINKED ISSUE #2780: [FEAT] Framework-agnostic MCP client auto-instrumentation
# feat(core): Framework-agnostic MCP client instrumentation

## Problem

TruLens already has MCP span attributes, framework-specific instrumentation, and an MCP tracing cookbook. However, the cookbook still requires a custom attribute extractor and an instrumented wrapper around the official MCP Python SDK's `ClientSession.call_tool()`.

Provide an explicit, framework-independent way to instrument this SDK boundary directly from `trulens-core`, without a separate app integration package.

## Expected Developer Experience

With `trulens-core` and the MCP Python SDK installed:

```python
from trulens.core.otel.mcp import instrument_mcp

instrument_mcp()

# Inside the application's existing TruLens recording context:
result = await mcp_session.call_tool(
    "get_weather", {"city": "Chicago"}
)
```

The SDK call automatically produces an MCP span under the active application trace. Users retain their existing application recorder and exporter configuration; this function does not create either.

Calling `instrument_mcp()` repeatedly is safe. Importing the module alone must not patch anything. Support applies to frameworks that call the instrumented official Python SDK method, not every MCP implementation.

## Implementation Scope

- Add `instrument_mcp()` in `src/core/trulens/core/otel/mcp.py`.
- Keep the MCP SDK optional: importing core or the new module must work without it. Calling `instrument_mcp()` without the SDK should raise an actionable dependency error.
- Instrument the official SDK's async `ClientSession.call_tool()` using existing TruLens instrumentation primitives where suitable.
- Verify and document supported SDK versions; do not assume methods such as `acall_tool` exist.
- Reuse `SpanAttributes.SpanType.MCP` and existing MCP attribute names without depending on LangChain or LangGraph.
- Preserve SDK arguments, return values, exceptions, and cancellation behavior.

## Span Contents

| Field | Expected behavior |
|---|---|
| Tool name | Extract `name` from positional or keyword arguments |
| Server name | Use reliable session metadata when available; otherwise record unknown |
| Input arguments | Structured JSON serialization of tool arguments |
| Output content | Structured serialization of the SDK result |
| Error status | Detect both raised exceptions and `CallToolResult.isError` |
| Execution time | Measure the awaited SDK call duration |

Review the existing `mcp_span()` helper rather than reusing it unchanged: its current extraction uses string conversion and exception-only error detection. Do not guess server identity from tool names.

Payload capture must respect applicable capture/redaction settings and size limits. Serialization failures must not break tool execution. Do not capture transport credentials.

## Required Cookbook Update

Update `examples/cookbooks/mcp_instrumentation.ipynb` in the same PR:

- Import `instrument_mcp` from core and call it once during setup.
- Remove the custom `mcp_attributes` extractor, tracing decorator, and tracing-only `MCPToolCallResult` wrapper.
- Call `mcp_session.call_tool()` directly from the LangChain tool adapter; retain the adapter and response-to-text conversion needed by the example.
- Keep existing LangGraph/TruGraph dependencies while the notebook still uses them; no new app package is required.
- Explain that MCP instrumentation is framework-independent while LangGraph is this example's agent framework.
- Preserve the local weather server, tool-quality evaluations, dashboard walkthrough, and resource cleanup.

## Acceptance Criteria

- [ ] Core and the new module import without the MCP SDK installed; activation gives a clear dependency error.
- [ ] Framework-free async recording tests against the local weather server capture two SDK calls as correctly parented MCP spans.
- [ ] Positional/keyword arguments, successful results, `isError` results, exceptions, and cancellation are covered.
- [ ] Repeated activation does not double-wrap; concurrent calls retain correct trace parents.
- [ ] LangGraph coexistence produces no duplicate SDK-call spans while preserving legitimate outer tool spans.
- [ ] Structured capture, unknown server identity, and payload capture controls are tested.
- [ ] Existing MCP tests pass and the updated cookbook is validated.

## Out of Scope

MCP server-side instrumentation, cross-process trace propagation, other SDKs/languages, resources/prompts/discovery tracing, new evaluators, and new dashboard features.

adity982:
## Summary

- add `trulens.core.otel.mcp.instrument_mcp()` for framework-independent instrumentation of the official MCP Python SDK
- lazily import the optional `mcp` dependency and make repeated activation idempotent without import-time patching
- record MCP tool/server metadata, bounded opt-in JSON payloads, SDK `isError` results, exceptions, cancellation, and execution time while preserving SDK behavior
- replace the cookbook's bespoke MCP extractor with direct `ClientSession.call_tool()` instrumentation

## Validation

- `python -m pytest tests/unit/test_otel_mcp.py -q` (3 passed)
- `python -m ruff check src/core/trulens/core/otel/mcp.py tests/unit/test_otel_mcp.py` (passed)
- `python -m ruff format --check src/core/trulens/core/otel/mcp.py tests/unit/test_otel_mcp.py` (passed)
- `python -m compileall -q src/core/trulens/core/otel/mcp.py tests/unit/test_otel_mcp.py` (passed)
- `python -m json.tool examples/cookbooks/mcp_instrumentation.ipynb` (passed)

The full Poetry/Make gate was not runnable on this Windows host because Poetry is not installed; the focused test environment used the repository's core OTEL dependencies without changing project manifests.

## Certification

Closes #2780


src/core/trulens/core/otel/mcp.py
```diff
@@ -0,0 +1,203 @@
...
+    if arguments is None and len(args) > 2:
+        arguments = args[2]
+    tool_name = tool_name or "unknown"
+    return {
+        SpanAttributes.MCP.TOOL_NAME: str(tool_name),
+        SpanAttributes.MCP.SERVER_NAME: _server_name(session),
+        SpanAttributes.MCP.INPUT_ARGUMENTS: _serialize(arguments),
+        SpanAttributes.MCP.OUTPUT_CONTENT: _serialize(result),
+        SpanAttributes.MCP.OUTPUT_IS_ERROR: bool(
+            exception is not None or getattr(result, "isError", False)
```
feiiiiii5, line 118: `getattr(result, "isError", False)` reads a field that no longer exists on the current SDK, so this is always `False` unless the call raised. Measured against `mcp==2.2.0` from PyPI:

```text
CallToolResult.model_fields -> ['meta','content','structured_content','is_error','result_type']
r = CallToolResult(content=[{'type':'text','text':'boom'}], is_error=True)
getattr(r, 'isError', False)  -> False     # this line
r.is_error                     -> True
hasattr(r, 'isError')          -> False
```

Feeding a real failing result through `_mcp_attributes` gives a span that contradicts itself:

```text
ai.observability.mcp.output_is_error = False
ai.observability.mcp.output_content  = '{"content":[...],"is_error":true,...}'
```

The mistake is easy to make and the reason is in the model config: `alias_generator=to_camel` means the *wire* name really is still `isError` (`model_dump(by_alias=True)` yields it), but attribute access uses the field name, and `populate_by_name=True` only makes construction accept both spellings. `mcp==1.9.0` does expose `.isError` (its fields are `['meta','content','isError']`), so this was correct until the 2.x rename.

Why the suite does not see it: `_Result` at `tests/unit/test_otel_mcp.py:25` sets `self.isError = is_error` by hand, so the assertion `assertTrue(span.attributes[MCP.OUTPUT_IS_ERROR])` checks the fake's 1.x spelling, not the SDK's 2.x one.

Minimal change is to prefer the new name and fall back:

```python
getattr(result, "is_error", getattr(result, "isError", False))
```

and for the fake, either set `is_error` (keeping one case at the old name to prove the fallback) or build the assertion on a real `CallToolResult`.

src/core/trulens/core/otel/mcp.py
```diff
@@ -0,0 +1,203 @@
...
+    create_function_call_context_manager,
+)
+from trulens.core.otel.instrument import _finalize_span
+from trulens.core.otel.instrument import instrument
+from trulens.otel.semconv.trace import SpanAttributes
+import wrapt
+
+_PATCH_FLAG = "__trulens_mcp_call_tool_instrumented__"
+_SECRET_KEY = re.compile(
+    r"(?:api[_-]?key|authorization|cookie|credential|password|private[_-]?key|secret|token)",
```
feiiiiii5, line 26: The bare `token` alternative matches any key *containing* it, which includes the usage counters. With `TRULENS_CAPTURE_TOOL_PAYLOADS=true`, calling the module's own `_serialize`:

```text
{"total_tokens":1234,"prompt_tokens":900,"city":"Chicago"}
  -> {"city":"Chicago","prompt_tokens":"[REDACTED]","total_tokens":"[REDACTED]"}
{"usage":{"input_tokens":10,"output_tokens":5}}
  -> {"usage":{"input_tokens":"[REDACTED]","output_tokens":"[REDACTED]"}}
{"tokenizer":"cl100k_base"}  -> {"tokenizer":"[REDACTED]"}
```

`api_key`/`authorization`/`secret` clearly belong in this list, and a tool payload carrying a raw `token` may well be a credential — so this is a judgement call, not an obvious bug. Flagging it because token counts are the numbers TruLens cost tracking consumes, and #2796/#2797 are about recovering exactly those. If the tradeoff is intended, a word in the docstring would be enough: that a tool result reporting usage under `*_tokens` will have it redacted. Otherwise anchoring the alternatives so `token` only matches whole keys (e.g. `(?:access|auth|id|refresh|bearer)[_-]?token` plus a standalone `token`) keeps the secrets out and lets the counters through.

feiiiiii5 (commented):
Ran the parts of the repo's gate that the PR notes could not be run on the author's host, then read the wrapper against the real SDK. Two findings, both measured; the verification results are here so a reviewer does not have to redo them.

**Clean, verified on Linux (CPython 3.11.15, pytest 9.1.1, `mcp` from PyPI):**

- `ruff check` and `ruff format --check` at the repo's own pin (0.16.3, `preview = true`): both new files pass.
- The Makefile runs `tests/unit/test_otel*.py` as its *own* batch with `-n auto --dist=loadscope` and warns those tests fail without process isolation, so that shape matters and a plain `pytest tests/unit/test_otel_mcp.py` does not cover it. Measured: the file passes alone (3 passed), and inside a batch with the three siblings that also touch global instrument state — `test_otel_instrument.py test_otel_async_concurrency.py test_otel_guardrail_spans.py` — the run is **36 passed where the same batch without the new file is 33 passed**. So it adds exactly its three tests and leaks nothing into its neighbours (`TRULENS_CAPTURE_TOOL_PAYLOADS` is restored in `tearDown`, and `enable_all_instrumentation()` in `setUp` matches what three existing files already do).
- The positional contract the attribute helper relies on is right: `_set_span_attributes` prepends the instance (`args_with_self_possibly = (instance,) + args`) before `_resolve_attributes` calls `attributes(ret, exception, *args, **all_kwargs)`, so `args[0]` really is the session.

**Two problems, both silent, both in the part the tests cannot reach:**

1. `OUTPUT_IS_ERROR` is permanently `False` against the current SDK. `_Result` in the test hand-sets `self.isError`, which is the pre-2.0 field name; `mcp` 2.x renamed the Python attribute to `is_error`. Tool failures that arrive the way the MCP spec intends — in band, no exception — are therefore recorded as successful. Inline detail and measurements on the `getattr` line.
2. The `token` alternative in the redaction key regex also matches the token *counts*, so the usage numbers this kind of integration exists to observe are erased from `input_arguments` / `output_content`. Inline with the observed output.

One structural note behind both: `mcp` is not a declared dependency anywhere — no match in `pyproject.toml` and no `name = "mcp"` entry in `poetry.lock` — so no CI job will ever install the SDK, and the hand-written fake is the only "SDK" these tests can meet. That is why a rename in the upstream model is invisible here. Cheapening the fake is not the fix; a single test that imports the real `CallToolResult` and asserts the attribute the wrapper reads would fail loudly on the next rename instead of quietly recording the wrong thing.

Nothing in the first two bullets should be read as doubt about the design — the idempotent class-level patch, the lazy optional import, the `get_baggage` short-circuit and the `_finalize_span` reuse all look right, and the cancellation test asserting exactly one span is a good catch to have written. Not affiliated with the project; posting as an interested reviewer.

Issue #2789, opened 2026-09-14
Propagate W3C traceparent on outbound LLM HTTP requests


sfc-gh-jreini:
## Problem

TruLens instruments LLM application code and exports client-side spans via OTLP, but it does not inject the W3C `traceparent` header into outbound HTTP requests to the LLM provider. This means the provider's server-side span (e.g. from a Snowflake Cortex AI Gateway) starts a new trace instead of becoming a child of the client's span. Client and server traces are disjoint and can only be correlated by content + timing.

## Expected behavior

When TruLens instruments a function that makes an HTTP call to an LLM provider (openai, anthropic, etc.), the active OTel span context should be propagated via the standard `traceparent` header on the outbound request. This connects the provider's server span as a child of the client span in a single trace.

## Proof of concept

A 5-line custom httpx transport demonstrates this works end-to-end against a Snowflake AI Gateway:

```python
from opentelemetry.trace.propagation.tracecontext import TraceContextTextMapPropagator

propagator = TraceContextTextMapPropagator()

class TraceparentTransport(httpx.HTTPTransport):
    def handle_request(self, request):
        headers = {}
        propagator.inject(headers)
        for k, v in headers.items():
            request.headers[k] = v
        return super().handle_request(request)

client = OpenAI(
    base_url=f"{GATEWAY_URL}/v1",
    api_key=token,
    http_client=httpx.Client(transport=TraceparentTransport()),
)
```

Result: all spans (client app spans + gateway server span) share one trace_id with correct parent_span_id chain:

```
rag-pipeline (root)
├── retrieve (parent: rag-pipeline)
├── generate (parent: rag-pipeline)
│   └── chat openai-gpt-5.4 [SERVER] (parent: generate)  ← gateway span
```

Without this, the gateway span is in a separate trace.

## Suggested approach

Instrument the httpx/requests transport used by the openai/anthropic Python SDKs to call `TraceContextTextMapPropagator.inject()` into outbound request headers when an OTel span is active. This could be done in the existing provider instrumentation or as a general httpx instrumentation hook.

sfc-gh-jreini:
[...]

PR #2792, opened 2026-09-17
feat(core): propagate W3C trace context on outbound HTTP requests

selvaganesh19:
### Problem

_TruLens_ instruments the caller, but the provider runs on the far side of an HTTP call. Nothing carries the active span's context across that boundary, so the provider's server span starts a trace of its own and the two ends end up correlated only by content and timing. Against a gateway, the gateway's server span lands in a different trace from the client span that called it.

### Solution

Adds `trulens.core.otel.propagation` with two pieces:

- `inject_trace_context(headers)` writes the active span's context using the globally configured propagators, and returns whether anything was written. It is inert when the span context is invalid, so globally disabling propagation keeps it disabled.
- `tracing_transport(inner)` returns an httpx transport that injects before delegating, so it can back a client handed to a provider SDK:

```python
client = httpx.Client(transport=propagation.tracing_transport())
provider = OpenAI(http_client=client)
```

This is opt-in. Nothing changes for existing users until they pass the transport.

### Note on the dependency

httpx is not a dependency of trulens-core, so it is imported inside `tracing_transport` rather than at module scope, and the module stays importable without it. The openai and anthropic SDKs already require httpx, so it is present wherever this is useful. No dependency was added.

### Testing

`tests/unit/test_otel_propagation.py` — 5 tests:

- injection with an active span, asserting the header carries that span's trace id
- no injection and no header written when no span is active
- existing headers preserved
- a request through the transport end to end, asserting the header reaches the request
- a request through the transport outside any trace still succeeds

```shell
$ python -m pytest tests/unit/test_otel_propagation.py -q
5 passed

$ python -m ruff check src/core/trulens/core/otel/propagation.py tests/unit/test_otel_propagation.py
All checks passed!

$ python -m ruff format --check src/core/trulens/core/otel/propagation.py tests/unit/test_otel_propagation.py
2 files already formatted
```

I did not run the wider suite, so I am not claiming it passes.

### Open question

The issue left the shape open between an opt-in transport and _TruLens_ instrumenting the client the provider SDKs construct. I went with opt-in because it changes nothing for existing users. If you would rather have the automatic path, I am happy to rework it — the injection helper is the same either way.

Refs #2789.

## Certification


Abelo9996 (commented):
Clean, well-scoped addition. Opt-in, no new core dependency (httpx imported lazily inside `tracing_transport`), uses the global propagators so it respects whatever is configured, and the 5 tests cover the real paths including running outside a trace. A couple of things worth addressing before merge:

- **should-fix: sync-only transport.** `tracing_transport` returns an `httpx.BaseTransport` and overrides `handle_request`, so it only backs a sync `httpx.Client`. The OpenAI and Anthropic SDKs are very commonly used with their async clients (`AsyncOpenAI` + `httpx.AsyncClient`), which need an `httpx.AsyncBaseTransport` implementing `handle_async_request`. As written an async client cannot use this. Either add an async variant or call out the sync-only limitation in the docstring so async users are not surprised.
- **should-fix: `close()` is not delegated to the inner transport.** `_TracingTransport` overrides only `__init__` and `handle_request`. With the default inner it constructs an `httpx.HTTPTransport()`, which owns a connection pool, but `httpx.Client.close()` calls the wrapper's `close()`, which falls through to the no-op base and never closes the inner transport. Delegating `close()` (and `__enter__`/`__exit__`) to `self._inner` avoids leaking the pool on client teardown.
- **nit: return value vs docstring.** `inject_trace_context` returns True whenever the span context is valid, before knowing whether the configured propagator wrote a header. With a no-op propagator installed it returns True but writes nothing, which is slightly at odds with "whether anything was written." Minor wording or a check on the mapping would align them.

Verdict: Comment. 2 should-fix (async transport, close delegation), 1 nit. Not blocking.

sfc-gh-jreini (commented):
Clean implementation of W3C trace propagation with solid test coverage. The injection logic is correct and the opt-in design is reasonable. Two issues: (1) async httpx clients aren't supported—common with OpenAI/Anthropic SDKs—needs either an async variant or clear sync-only documentation to prevent runtime errors, and (2) the custom transport should implement close() to properly release connection pool resources. VERDICT: Comment — 2 should-fix items, no blockers.

src/core/trulens/core/otel/propagation.py
```diff
@@ -0,0 +1,79 @@
...
+        True when a valid span context was available and injected.
+    """
+
+    if not get_current_span().get_span_context().is_valid:
+        return False
+
+    inject(headers)
+    return True
+
+
```
sfc-gh-jreini, line 50: Add async support or document the sync-only limitation clearly in the docstring. OpenAI and Anthropic SDKs both support `httpx.AsyncClient`, which requires `AsyncBaseTransport.handle_async_request()` instead of sync `handle_request()`. Without an `async_tracing_transport()` function or explicit "sync clients only" documentation, async users will hit type or runtime errors when they try `AsyncClient(transport=tracing_transport())`.

src/core/trulens/core/otel/propagation.py
```diff
@@ -0,0 +1,79 @@
...
+        ) -> None:
+            self._inner = (
+                inner_transport
+                if inner_transport is not None
+                else httpx.HTTPTransport()
+            )
+
+        def handle_request(self, request: httpx.Request) -> httpx.Response:
+            inject_trace_context(request.headers)
+            return self._inner.handle_request(request)
```
sfc-gh-jreini, line 89: Implement `close(self)` to delegate to `self._inner.close()` so connection pools are released when the client shuts down. Without forwarding lifecycle methods, resources leak—httpx transports manage connection state and must clean up.
```python
def close(self) -> None:
    self._inner.close()
```

tests/unit/test_otel_propagation.py
```diff
@@ -0,0 +1,88 @@
...
+    def setUp(self) -> None:
+        self.provider = TracerProvider()
+        self.provider.add_span_processor(
+            SimpleSpanProcessor(InMemorySpanExporter())
+        )
+        self.tracer = self.provider.get_tracer(__name__)
+
+
+class TestInjectTraceContext(_TracingTestCase):
+    def test_injects_active_span_context(self) -> None:
```
sfc-gh-jreini, line 58: Consider extracting trace_id formatting into a helper or test setUp to reduce duplication between test_injects_active_span_context (line 35) and test_transport_injects_on_outbound_request (line 70). Not blocking—tests are clear as written—but DRYing the pattern would make future trace ID assertions easier to update.

selvaganesh19:
Thanks both — all three points addressed in adcf828.

**Async transport.** Added `async_tracing_transport()`, returning an `httpx.AsyncBaseTransport` that implements `handle_async_request` and `aclose`. `tracing_transport()` stays sync and its docstring now names the async counterpart, since httpx dispatches the two through different base classes and they are not interchangeable:

```python
client = httpx.AsyncClient(transport=async_tracing_transport())
provider = AsyncOpenAI(http_client=client)
```

**Lifecycle.** Both transports now delegate `close()`/`aclose()` and the sync and async context manager protocols to `self._inner`. Two tests cover it: one calls `close()` on the transport directly, one goes through `httpx.Client.close()`, since that is the path where the pool leak showed up.

**Return value.** `inject_trace_context` now injects into a scratch mapping, returns `False` when that mapping stays empty, and only then updates the real headers. With a no-op propagator the span context is valid but nothing is written, and the return value now says so. Test added that installs a no-op propagator, asserts `False`, and asserts the caller's mapping is untouched.

**Test duplication.** Added a `current_trace_id()` helper on the test case base class; both trace-id assertions go through it now.

Test count is 5 → 11, all passing:

```
$ python -m pytest tests/unit/test_otel_propagation.py -q
11 passed

$ python -m ruff check src/core/trulens/core/otel/propagation.py tests/unit/test_otel_propagation.py
All checks passed!

$ python -m ruff format --check src/core/trulens/core/otel/propagation.py tests/unit/test_otel_propagation.py
2 files already formatted
```

I still have not run the wider suite, so I make no claim about it.

One thing I left alone but can change if you would rather: the two transport classes are defined inside their factory functions, so each call to `tracing_transport()` builds a new class. That keeps httpx out of the module's import time, which is why I did it, but if you would prefer them hoisted to module scope with a lazy httpx import, that is a small change.

selvaganesh19 (commented):
-Done in adcf828 — added async_tracing_transport(), and the sync docstring now names the async counterpart.
- Done in adcf828 — both transports delegate close()/aclose() and the context manager protocols to self._inner.
- Done in adcf828 — added a current_trace_id() helper on the test base class.

sfc-gh-jreini (approved):
Thanks for addressing feedback! This looks great. Will merge once CLA check passes.

To make it pass by signing CLA, follow the info from this comment: https://github.com/truera/trulens/pull/2792#issuecomment-5709924181

selvaganesh19:

Abelo9996:
Thanks for addressing all three. The `async_tracing_transport` variant and the `close`/`aclose` delegation both look right, and injecting into a scratch mapping to decide the return value is a clean way to report whether anything was actually written. Nice work.

PR #2793, opened 2026-09-17
fix: get_leaderboard_aggregates crashes unconditionally with OTel tracing disabled

### LINKED ISSUE #2729: [BUG] get_leaderboard_aggregates always crashes when TRULENS_OTEL_TRACING is disabled

avanish-garg:
# Description

`_get_leaderboard_aggregates_pre_otel` unconditionally raised `NotImplementedError` while
building its SQL statement — `cost_json` is a `Text` column (`TYPE_JSON = Text` in
`database/orm.py`), and SQLAlchemy doesn't support `[]` subscript indexing on `Text`
columns. Anyone running with `TRULENS_OTEL_TRACING=0` (the supported disabled-tracing
mode) lost leaderboard aggregates entirely, including the dashboard leaderboard view.
Reported in #2729 and independently cross-verified there by dosu.

Fixing that expression exposes a second, previously-unreachable bug in the very next
line: `self.orm.Record.latency` doesn't exist as a column at all — `Record` only has
`perf_json` (also `Text`, same shape as `cost_json`). Python evaluates `sa.select()`'s
positional args left to right, so the `cost_json` failure always masked this one.

Fix: reuse `_extract_tokens_and_cost`/`_extract_latency`, the helpers this file already
uses elsewhere for this exact "Text column holding JSON" shape, rather than writing new
per-dialect JSON+datetime SQL. `_get_leaderboard_aggregates_pre_otel` now fetches raw
per-record rows (app grouping fields + `cost_json` + `perf_json`), applies those helpers,
and aggregates in pandas. The feedback-score half of the method (`fb_stmt`) is untouched.

Closes #2729

- `_json_path_expr` (the dialect-aware SQLite/MySQL/Postgres JSON-extraction helper
  already used by the OTel path) isn't reusable here — it's gated to real `sa.JSON`-typed
  columns via an `isinstance` check, and both `cost_json`/`perf_json` are `Text`. Hence
  the pandas-level fix rather than a SQL-level one.
- The empty-database case (`base_df.empty` → early return) is preserved exactly.

## Type of change

- [x] New Tests

## Certification


sfc-gh-jreini (approved):
LGTM - thanks for contributing!

avanish-garg:

feiiiiii5 (commented):
I ran this against both trees and the fix holds up. Confirmations first, then five things worth a look — none of them are reasons to hold the PR.

## What I verified

**The test bites.** Two worktrees, one venv each so the imported source is unambiguous (`trulens.core.database.sqlalchemy.__file__` printed from inside each run):

| tree | `cost_json["n_tokens"]` present | PR's test file |
|---|---|---|
| base `33e4f661` | yes (1 occurrence) | **2 failed** — `NotImplementedError: Operator 'getitem' is not supported on this expression` |
| head `53a029e0` | no | **2 passed** |

Both tests fail at base, including `test_get_leaderboard_aggregates_empty_database` — which is the detail that confirms your framing: the raise happens at *statement construction*, so it fires even with zero rows and on every backend. `is_otel_tracing_enabled()` returns `False` under `TRULENS_OTEL_TRACING=0` and the dispatch at `sqlalchemy.py:1874/1878` sends it to `_get_leaderboard_aggregates_pre_otel`, so the test is exercising the path it claims to.

**The four factual claims in the description check out** against base: `TYPE_JSON = Text` (`orm.py:36`), `Record.cost_json`/`perf_json` are `Column(TYPE_JSON, nullable=False)` (`orm.py:243-244`), `Record` has no `latency` column at all, and `_extract_tokens_and_cost` really returns `["total_tokens", "total_cost", "cost_currency"]` (`sqlalchemy.py:2733`) — so the `agg()` keys are not invented.

**The aggregation is semantically faithful**, which is the thing I most expected to find wrong: `avg(cost_json["n_tokens"])` → `mean(total_tokens)`, `sum(cost_json["cost"])` → `sum(total_cost)`, `count(distinct record_id)` → `nunique(record_id)`. And the tuple order out of `groupby(as_index=False).agg(...)` matches the positional `columns=[...]` at `sqlalchemy.py:1789-1800` exactly. Row order doesn't matter because the feedback join is key-based (`base_df.merge(pivot, on=[...])`, `:1830`), so pandas sorting the groups is harmless.

## Five things worth a look

**1. Nothing asserts a latency *value*.** `assertIn("Average Latency (s)", row.index)` only checks the column exists. The fixture records carry a default `Perf`, so latency is 0 either way, and a mis-wired `_extract_latency` — or aggregating the wrong column — would still pass. Given that the missing `Record.latency` is half the story in your description, it'd be nice to pin the number: build one record with a known `perf` and assert the mean. This is the one I'd actually change.

**2. One transaction became two.** At base a single `with self.session.begin()` (`:1722`) covered both `base_rows` (`:1753`) and `fb_rows` (`:1787`). The new code closes it after the record read and opens a second one for the feedback read, so the two halves of a leaderboard row are no longer read from one snapshot — a write landing in between produces a row that never existed. Low impact for a dashboard, but keeping both inside the original `with` costs nothing.

**3. Aggregation moved from the database into pandas.** `_get_leaderboard_aggregates_pre_otel` now transfers every `Record` row, with two JSON text blobs each, and materialises them as a DataFrame; previously the DB grouped and returned one row per app. On a large database that's O(records) memory where it used to be O(apps). I don't think you had much choice — the OTel path can aggregate in SQL because its columns are genuine `JSON` (`orm.py:586`, `:601`, `:630`) while the pre-OTel ones are `Text`, and per-dialect JSON-over-Text SQL is exactly what you were avoiding. Worth a sentence in the comment saying the trade is deliberate, or a follow-up issue, so the next reader doesn't "optimise" it back into broken SQL.

**4. Correctness now depends on dict insertion order ~40 lines away.** `agg(Records=..., **{"Total Tokens": ..., "Average Latency (s)": ..., "Total Cost (USD)": ...})` produces tuples consumed positionally by the `columns=[...]` list at `:1789`. It's right today, but a reordered dict entry silently renames a column rather than failing. Building the DataFrame with names and then reindexing to the expected column list would make a mismatch loud.

**5. The module-scope `os.environ` write, and its comment.** `is_otel_tracing_enabled()` is not cached — it re-reads `os.getenv` at each of its 11 call sites in `session.py` — so there's no frozen flag to get ahead of, and the comment "must be done before any TruSession is created to avoid freezing the experimental flag" isn't quite what's happening. What *is* true is that pytest imports every test module during collection, so this assignment sets the mode process-wide for the rest of the run. I tried to demonstrate an impact and could not: `tests/unit/test_otel_tru_basic.py` gives 3 passed/1 skipped alone and 5 passed/1 skipped when run in the same process as this module. So this is hygiene rather than a breakage — a `monkeypatch`/fixture-scoped env, or a subprocess, would keep the intent without the leak.

## Two notes that aren't about the code

**CI may never run this test.** All 11 check runs on `53a029e0` are green, but none of them looks like a pytest job (static py310/py311/py313, conda build, lint & format, detect changed paths, CLA, GitGuardian), and the repo's own `.github/workflows` contains only `certifications.yml`, `cla.yml` and `request-area-reviews.yml`. If the unit tests run in a pipeline I can't see, ignore this; if they don't, the regression test only protects anyone who runs it locally.

**#2730 collides with this.** It fixes the same issue (#2729), edits the same file, and adds a file at the same path `tests/unit/test_leaderboard_pre_otel.py`, so whichever lands second will conflict. #2730 was approved on 2026-08-26, had changes requested the next day, and has been quiet since 2026-09-13. Flagging it only as merge mechanics — which of the two to take is the maintainers' call and not something I have context on.

---

Reproduce: two worktrees (`git worktree add --detach <sha>`), one venv per tree with `uv pip install -e <tree>/src/core -e <tree>/src/otel/semconv pytest tqdm pandas pydantic pyyaml sqlalchemy opentelemetry-sdk opentelemetry-api`, then `PYTHONPATH=<tree> python -m pytest tests/unit/test_leaderboard_pre_otel.py -q` from inside each tree. Python 3.11.15, macOS arm64. Printing `trulens.core.database.sqlalchemy.__file__` per tree is what makes the comparison meaningful — with a single editable install both runs silently execute the same source.

AI-assisted review; the measurements above are mine.

feiiiiii5:
Correction to the last section of my review above, because I got a fact wrong and it changes the picture.

I wrote that #2730 "was approved on 2026-08-26, had changes requested the next day, and has been quiet since 2026-09-13". That is wrong: I read the review timestamps and not the commit list. #2730's author has pushed since —

| commit | date | subject |
|---|---|---|
| `e92c4120` | 2026-09-10 | Merge branch `main` into the fix branch |
| `5f8672f3` | 2026-09-15 | *fix(database): split pre-OTEL cost sum by currency — report Snowflake Credits separately from USD* |
| `ea51aa5d` | 2026-09-17 | *style(database): apply ruff formatting to changed files* |

so it is actively maintained, currently `mergeable=true` with `mergeable_state=blocked`, eight commits, +309/-11.

That makes the collision sharper than I made it sound rather than softer: two open PRs, both rebased onto recent `main`, both rewriting `_get_leaderboard_aggregates_pre_otel`, and both adding a file at the same path `tests/unit/test_leaderboard_pre_otel.py` (#2730's is 235 lines, this one's is 79). Whichever merges second has to resolve against the other by hand. Also worth knowing for whoever decides: #2766, which addressed the mixed-currency half of #2759, merged on 2026-09-08, and `5f8672f3` above extends the same per-currency idea to the pre-OTel path — so #2730 now overlaps #2766's territory too, while this PR does not touch currency at all. (#2759 itself is still open; its other half is the "Total Tokens" avg-vs-sum question, which neither PR resolves.)

Which of the two to take is the maintainers' call and I have no context on it. I'm only correcting my own error so the decision isn't made on a false premise: #2730 is not abandoned.

Everything else in my review concerns this PR's own diff and is unaffected — the base/head runs (`33e4f661` → 2 failed, `53a029e0` → 2 passed), the four verified claims, and points 1-5.

avanish-garg:
Flagging something I noticed after this was approved! @feiiiiii5 @sfc-gh-jreini 

While looking into the collision with #2730 (same issue, different approach), I saw your
review there retract an approval and request SQL-based aggregation instead of moving it
into Python/pandas, for scalability reasons. This PR takes the same Python/pandas
approach on that review — I want to make sure that's still fine with you rather than
assume, since the two PRs are functionally in the same position architecturally.

Two things worth noting on the technical merits:
- This PR doesn't have the currency-mixing bug `tonydzi` found on #2730 (summing USD and
  Snowflake credits together) — it preserves the existing hardcoded-zero-credits behavior
  rather than attempting a split, so no new regression there.
- That said, the `_json_path_expr` + `sa.case` pattern `tonydzi` posted on #2730 for the
  cost/currency split is genuinely usable here too, and would get this closer to a
  SQL-based aggregation for the cost fields specifically. The harder part is `perf_json`
  (latency) — that needs per-dialect datetime-diff-over-JSON-text SQL, which is why I went
  with pandas originally rather than hand-rolling that.

Happy to rework this as SQL-based (reusing the currency-split pattern from #2730's review)
if that's your preference given the #2730 feedback — just didn't want to redo the
architecture without checking first, since you'd already approved this version.

feiiiiii5:
Small correction, so this does not get settled on a wrong premise: I did not review #2730. Its review list is joshreini1 APPROVED (2026-08-26), joshreini1 CHANGES_REQUESTED (2026-08-27) and tonydzi COMMENTED (2026-09-13), with nothing from me in the reviews, the review comments or the issue comments. The retract-the-approval and "SQL rather than frontend Python" direction is joshreini1's.

Nor was my comment on this PR an approval - it is a `COMMENTED` review, so there is no endorsement of mine for you to be consistent with.

The bit that does matter is that both signals come from the same maintainer. joshreini1 on #2730, about the same legacy non-OTEL path this PR changes:

> This PR should keep its scope on the legacy non-OTEL path, but implement that aggregation in SQL rather than moving it into frontend Python.

and sfc-gh-jreini - also Josh Reini - left "LGTM - thanks for contributing!" here on 2026-09-17. Which of those two governs #2793 is not mine to decide, so I would get that from them rather than rework the architecture on my account.

What I can add is why I did not treat the pandas move as a defect in my review, since it bears on whether this is a straight port of what #2730 was asked to do: the pre-OTel columns are `Text`, not `JSON` (`orm.py:36`, with `cost_json` / `perf_json` at `:243-244`), so aggregating them in SQL means per-dialect JSON-over-text expressions. The `sa.case` + `_json_path_expr` pattern from #2730 handles the cost/currency split, where the values are scalars; it does not cover a datetime difference over the `perf_json` text blob, which is the obstacle you describe. If the maintainer does ask for SQL here, that asymmetry is worth putting in front of them before any rewrite.

avanish-garg:
Thanks for the correction — my mistake, I conflated your review with joshreini1's. You're right that the approval here is `sfc-gh-jreini` and the "SQL not pandas" ask on #2730 is `joshreini1` — and since their profiles confirm those are the same person's two accounts (pre- and post-Snowflake), the real question is just which of his own two calls governs this PR, which only he can resolve.

@sfc-gh-jreini — could you clarify whether the pandas approach here is fine as approved, or whether you'd like it reworked to SQL to match what you asked for on #2730? Happy to do the rework (reusing the currency-split pattern from #2730's review for the cost fields) if that's the direction — just don't want to guess.

Thanks also for the technical note on why `perf_json`/`cost_json` being `Text` rather than `JSON` makes a full SQL port asymmetric — that's useful context either way.

joshreini1:
Thanks for this, @avanish-garg — closing in favor of #2730, which fixes the same issue and landed first (2026-08-26 vs 2026-09-17). Not a reflection on the quality of the work here, and I want to be specific about the reasoning rather than just pointing at the date.

### Why #2730

**Aggregation stays in SQL.** #2730 originally took the same pandas approach as this PR, and I approved it, then retracted and asked for SQL-level aggregation instead — the pre-OTel path should return one row per app rather than transferring every `Record` with two JSON text blobs and grouping in Python. That's O(apps) instead of O(records). The author reworked it in 1407e183. Merging this PR would land the approach I'd asked to change, so the inconsistency is mine, not yours — my `sfc-gh-jreini` approval here on 09-17 contradicted my own August review request on #2730, and I should have caught the overlap then.

**Two behaviors that #2730 also corrects.** Both are in the aggregation this PR keeps:

- `Total Tokens` uses `mean`. That faithfully preserves the original `avg()`, but the label has always said "Total" and the OTel path sums. #2730 changes it to `sum`.
- `Total Cost (USD)` sums every record regardless of `cost_currency`, and `base_df["Total Cost (Snowflake Credits)"] = 0.0` stays hardcoded. So a legacy-mode Snowflake user sees credits reported as dollars, with nothing raising. The OTel path splits on currency with `sa.case` and carries an explicit comment about it: "Mixing currencies in a single total is wrong and is what `Cost.__add__` refuses to do." #2730 now does the same split, after @tonydzi measured the problem there.

### One correction on the premise

The description here says `_json_path_expr` "isn't reusable here — it's gated to real `sa.JSON`-typed columns via an `isinstance` check." That gate is actually in `_json_extract_otel`, a separate method scoped to the OTel `Event` table that happens to call `_json_path_expr`. `_json_path_expr` itself takes any column object and dispatches on dialect only, with a Snowflake override in `SnowflakeSQLAlchemyDB`. #2730 calls it directly on the Text `cost_json` and it works — I also confirmed Snowflake's `JSON_EXTRACT_PATH_TEXT` handles a VARCHAR argument. Easy conflation to make, and worth recording since it was the stated reason for choosing pandas over SQL.

### What was right here

The diagnosis was correct and well written: `TYPE_JSON = Text` as root cause, the `getitem` failure happening at statement-construction time so it fires on every backend and even on an empty database, and `Record.latency` not existing as a column at all — a second bug the first one masked, since `sa.select()` evaluates its arguments left to right. Your empty-database test is what demonstrates the construction-time behavior most clearly. Thanks also to @feiiiiii5 for the two-worktree verification, which is the right way to test this given a single editable install silently makes both trees execute the same source.

Two notes from that review worth carrying forward regardless of which PR landed: the module-scope `os.environ["TRULENS_OTEL_TRACING"] = "0"` write leaks process-wide once pytest imports the module during collection, and is better done with a `monkeypatch` fixture. And the single transaction that became two — the base implementation read `base_rows` and `fb_rows` under one `with self.session.begin()`, so splitting it means the two halves of a leaderboard row no longer come from one snapshot.

Sorry for the wasted cycles — two PRs shouldn't have been open this long against the same issue with an approval on each. Genuinely hope you'll send more; please ping me on an issue before starting if you want me to confirm nothing is already in flight.

avanish-garg:
Sounds reasonable, thanks @joshreini1

PR #2795, opened 2026-09-18
`fix(bedrock): disclose response_format is not honored, not silently drop it`

### LINKED ISSUE #2794: [BUG] Bedrock provider silently ignores response_format instead of disclosing it like LangChain does
**Bug Description**

`LLMProvider.generate_score()` always passes `response_format` when calling
`_create_chat_completion()`, with the comment "Try structured outputs first; provider
will probe and fall back if unsupported." Every provider either implements structured
output support (anthropic, cortex, litellm, ollama, google, openai) or, when it genuinely
can't, discloses that fact — the LangChain provider explicitly logs `"Ignoring
response_format in LangChain provider; not supported via chain.invoke."` when this
happens.

The Bedrock provider's `_create_chat_completion` accepts `response_format` in its
signature but never references it anywhere else in the file — confirmed with `grep`, not
just a read-through. It's silently dropped with zero signal, unlike LangChain's disclosed
equivalent.

**To Reproduce**

```python
from trulens.providers.bedrock import Bedrock

provider = Bedrock(model_id="amazon.nova-lite-v1:0")
provider.generate_score(system_prompt="Rate this from 0-10: ...")
# response_format is passed in under the hood and silently ignored —
# no log, no warning, nothing.
```
**Expected Behavior**

Since Bedrock's InvokeModel-based request/response bodies (hand-rolled per model
family — Nova, Titan, Anthropic, Cohere, AI21, Mistral, Meta) don't have a JSON-schema
mechanism wired up for any of them, fully honoring response_format would require moving
all seven families onto Bedrock's unified Converse API — a larger change than a
disclosure fix warrants. At minimum, this should follow the same pattern the LangChain
provider already established for the identical "can't support this" situation: a debug
log disclosing that response_format is being ignored, not silence.

**Relevant Logs/Tracebacks**

None — that's the point. No error, no warning, nothing at any log level.

**Environment:**

OS: any
Python Version: any supported
TruLens version: current main
Versions of other relevant installed libraries: n/a (no live AWS call needed to
reproduce — the gap is in the provider's own code, not the Bedrock API response)

**Additional Context**

Scores still compute correctly today — the base class's plain-string-response fallback
path (feedback_generated.re_configured_rating) already handles a provider that never
returns a structured response. This is a diagnosability gap, not a correctness bug: users
get worse score-parsing reliability than other providers with zero indication why.

Have a fix ready (PR incoming): adds the same logger.debug disclosure LangChain already
uses for this exact situation, plus a test extending the existing _DummyBedrockClient
mocking harness in test_bedrock_capabilities.py (no AWS credentials needed).

avanish-garg:
# Description

`LLMProvider.generate_score` always passes `response_format` when calling
`_create_chat_completion` ("Try structured outputs first; provider will probe and fall
back if unsupported"). Every other provider either implements it (anthropic, cortex,
litellm, ollama, google, openai) or, if it genuinely can't, discloses that fact —
`langchain`'s provider explicitly `logger.debug`s "Ignoring response_format in LangChain
provider; not supported via chain.invoke."

Bedrock's `_create_chat_completion` accepts `response_format` in its signature but
references it nowhere else in the 266-line file — confirmed via `grep`, not just a
read-through. Its `InvokeModel` request/response bodies are hand-rolled per model family
(Nova, Titan, Anthropic, Cohere, AI21, Mistral, Meta) and none of them wire up a
JSON-schema mechanism, so properly honoring it would mean moving all seven families onto
Bedrock's unified `Converse` API — a much bigger change than this gap warrants for a first
PR. This mirrors `langchain`'s existing pattern for the same situation instead: a debug
log disclosing that the parameter is being ignored, rather than the current silent no-op.

Scores still come out correct today (the base class's plain-string fallback path already
handles a provider that never returns a structured response) — this is a diagnosability
gap, not a correctness bug.

Closes #2794 

Test extends the repo's own existing `_DummyBedrockClient` mocking harness in
`test_bedrock_capabilities.py` — no AWS credentials needed.

## Type of change

- [x] New Tests

## Certification


src/providers/bedrock/trulens/providers/bedrock/provider.py
```diff
@@ -64,6 +64,18 @@ def _create_chat_completion(
     ) -> str:
         assert self.endpoint is not None
 
+        # Bedrock's raw InvokeModel request/response shapes below are
+        # hand-rolled per model family (Nova, Titan, Anthropic, Cohere, AI21,
+        # Mistral, Meta) and none of them wire up a JSON-schema mechanism, so
+        # response_format can't be honored here. Disclose this the same way
+        # the LangChain provider does for the same "can't support structured
+        # outputs" situation, rather than silently dropping it.
+        if response_format is not None:
```
sfc-gh-jreini, line 73: Move the `response_format` check before the `messages`/`prompt` validation block (starting line 67) so the disclosure appears in logs even when the method immediately fails — this makes the "I tried passing `response_format` but the call failed" debugging scenario clearer. Non-blocking.

sfc-gh-jreini (approved):
Solid improvement: the PR fixes a real diagnosability gap where `response_format` was silently dropped. The disclosure message correctly mirrors the LangChain provider pattern (verified at `langchain/provider.py:90`), the test properly extends the existing `_DummyBedrockClient` harness, and the log placement is correct (checks the parameter before any model-family dispatch). The PR description accurately explains why implementing structured outputs across seven families is out of scope. Ready to merge.

avanish-garg:
The failing "PR Validation Eval" check isn't related to this change — verified before posting rather than assuming.

The 5 test failures listed in the check's annotations are all in unrelated files (`test_otel_utils.py`, `test_otel_async_concurrency.py`, `test_otel_metrics.py`) — none touch `bedrock` or `response_format`. My change is a single `logger.debug` call added to `Bedrock._create_chat_completion`.

Confirmed this is pre-existing, not something this PR introduces: on a clean checkout of `main` (no changes at all), running just `test_otel_utils.py::TestOtelUtils::test_warns_when_disabled` and both `test_otel_metrics.py::TestOtelMetricsSpanProcessor` tests together already reproduces one of the same failures (`test_otlp_option_validation`: `ValueError not raised`) — this looks like test-isolation/global-singleton state leaking across tests in the same process, unrelated to any specific PR's diff.

Flagging in case a re-run helps, but wanted to confirm first that this isn't something I need to fix in the diff itself.

src/providers/bedrock/trulens/providers/bedrock/provider.py
avanish-garg: Line 67 is actually inside the explanatory comment above the response_format check, not the start of the messages/prompt block — that block starts at line 78, well after the check at line 72. So the disclosure already logs before the method can raise on missing messages/prompt. This has been the ordering since the original commit (verified there's only ever been the one commit on this branch, and main hasn't touched this file since I branched). Let me know if you were looking at a different revision — happy to adjust if I'm missing something.

avanish-garg:
any update @sfc-gh-jreini ?

PR #2797, opened 2026-09-18
feat(cursor): recover token usage from the transcript on stop

### LINKED ISSUE #2796: [FEAT] Cursor integration has no cost/token tracking, unlike Claude Code and OpenCode
**Feature Description**

`src/apps/cursor/trulens/apps/cursor/client.py`'s `client_spec` has no `extract_overrides`,
unlike both sibling "Agent Skills" clients — `claude` (`src/apps/claude/.../client.py`) and
`opencode` (`src/apps/opencode/.../client.py`) both have one. `extract_overrides` is the
mechanism these clients use to backfill data the live hook payload doesn't carry, by
reading a local transcript/session file referenced in the hook payload.

I checked Cursor's own hooks documentation (https://cursor.com/docs/agent/hooks) directly
rather than assuming parity is needed:

- Cursor's hook payloads *do* include `model` directly, and `afterAgentResponse` includes
  the response text (`"text"` field) directly — correctly explaining why Cursor doesn't
  need an override for those two fields, unlike Claude Code, whose `Stop` hook doesn't
  reliably carry them.
- **Token usage (input/output tokens) is confirmed absent from every Cursor hook
  payload** — the docs explicitly note it's "deliberately excluded."
- Cursor's hook payload *does* include a `transcript_path` field (`CURSOR_TRANSCRIPT_PATH`
  env var), structurally identical in purpose to the field Claude Code's
  `_transcript_overrides` already reads to recover token counts. So the same fix pattern
  looks directly applicable here.

**Reason**

Cursor's own hooks documentation isn't specific about the transcript file's internal
schema. Cross-referencing several independent third-party tools that already parse Cursor
transcripts for their own purposes, the consistent claim is that Cursor uses the same
JSONL format as Claude Code, with the same `role`/`message.content[]` shape — consistent
with Cursor and Claude Code sharing the same underlying `/v1/sessions` hook contract. I
don't have a live Cursor installation to verify this against a real transcript file
myself, so I'm not confident enough in the exact schema to submit a PR yet — flagging the
gap and the fix's shape here first, in case a maintainer or another contributor with
Cursor access can confirm the transcript format (or has already reverse-engineered it for
something else).

**Importance of Feature**

Without this, Cursor users of TruLens's Agent Skills integration get zero cost/token
tracking today, while Claude Code and OpenCode users of the same integration surface get
real cost capture — an inconsistency in a tool whose core value proposition is cost/usage
observability. Happy to build the fix (reusing Claude Code's already-proven
`_transcript_overrides` parsing logic) once the transcript format is confirmed, or to try
it myself against a synthetic fixture with the caveat disclosed if that's preferred.

avanish-garg:
# Description

Cursor's hook payloads never carry input/output token counts on any event — confirmed
directly against Cursor's own hooks documentation, not assumed. Unlike Claude Code and
OpenCode's `client_spec`s, Cursor's had no `extract_overrides` at all to backfill this.

Scoped narrower than Claude Code's equivalent override: Cursor's docs confirm `model` is
present directly on every hook payload, and response text is present on
`afterAgentResponse` via the `"text"` field, already covered by this client's own
`field_aliases.response`. Neither needs an override here — only token usage does.

**Important caveat, disclosed rather than glossed over**: Cursor's transcript file format
is not officially documented. This mirrors `trulens.apps.claude.client._transcript_overrides`'s
approach (same `type`/`message.content[]`/`message.usage.{input,output}_tokens` shape,
same "only the current turn" isolation via a human-prompt boundary reset) based on:
Cursor's docs confirming the `transcript_path` field exists in the hook payload, and
several independent third-party tools that already parse Cursor transcripts (agentsview,
engram, entireio/cli, GrayCodeAI/trace) consistently describing the same JSONL/role/content
shape — consistent with Cursor and Claude Code sharing the same underlying `/v1/sessions`
hook contract. I don't have a live Cursor installation to verify this against a real
transcript file. If the actual format differs, this fails soft (returns `{}` on any parse
mismatch, same as the existing `OSError`/`JSONDecodeError` handling) rather than raising —
worst case is no regression, not a crash. Flagging for reviewer confirmation before merge.

Closes #2796

If a reviewer can confirm (or correct) the real transcript schema against an actual
Cursor session, that would let this be tightened up or fixed before merge rather than
after.

## Type of change

- [x] New Tests

## Certification


tests/unit/test_otel_client_hook_plugins.py
```diff
@@ -339,6 +339,89 @@ def test_claude_stop_uses_only_current_turn_and_skips_bad_transcript_lines(
     assert event.output_tokens == 4
 
 
+def test_cursor_stop_recovers_usage_from_transcript(tmp_path):
```
sfc-gh-jreini, line 342: Add a test verifying 'current turn only' behavior — the PR description explicitly claims this mirrors Claude Code's human-prompt boundary reset, but there's no test proving it. Copy the structure of `test_claude_stop_uses_only_current_turn_and_skips_bad_transcript_lines` (lines 258-339 above) for Cursor: create a transcript with two user prompts separated by assistant messages carrying token usage, then assert only the tokens after the second user prompt are counted. Without this, the reset logic in `_is_human_prompt` and lines 95-97 of `client.py` is untested.

src/apps/cursor/trulens/apps/cursor/client.py
```diff
@@ -29,6 +31,93 @@ def _config(command: str):
...
+    if payload.get("hook_event_name") != "stop":
+        return {}
+    transcript_path = payload.get("transcript_path")
+    if not transcript_path:
+        return {}
+
+    input_tokens = 0
+    output_tokens = 0
+    try:
+        with Path(str(transcript_path)).open(encoding="utf-8") as transcript:
```
sfc-gh-jreini, line 88: Break this line to stay under 80 characters per project standards (currently ~86 chars). Split after the `open()` call:
```python
with Path(str(transcript_path)).open(
    encoding="utf-8"
) as transcript:
```

tests/unit/test_otel_client_hook_plugins.py
```diff
@@ -339,6 +339,89 @@ def test_claude_stop_uses_only_current_turn_and_skips_bad_transcript_lines(
...
+            "transcript_path": str(transcript),
+        },
+    )
+
+    # model comes through unchanged from the payload - no override needed.
+    assert event.model == "claude-sonnet-4-5"
+    assert event.input_tokens == 15
+    assert event.output_tokens == 7
+
+
```
sfc-gh-jreini, line 393: Add test coverage for error paths: missing `transcript_path` in the payload (should return `{}`), `OSError` when the file doesn't exist (same), and malformed usage values (`"bad"` instead of int, or `None`) to verify `_usage_value`'s safe handling. The Claude Code tests include malformed token handling (line 287 above); Cursor's should too.

sfc-gh-jreini (changes requested):
The implementation correctly mirrors Claude Code's transcript-based token recovery pattern and handles parse errors gracefully. However, the test suite is incomplete: it's missing the critical 'current turn only' test that verifies usage resets on human prompts (the core isolation guarantee mentioned in the PR description), plus coverage for error paths and malformed data. One line exceeds the 80-character project limit. Request changes: 1 blocker, 2 should-fix items.

avanish-garg:
Addressed both should-fix items, both in one new commit:

**Current-turn isolation test** — added, mirroring the Claude Code test's structure exactly (human-prompt entry resets the running count, bad JSON lines and malformed usage values are skipped, only tokens after the most recent prompt count). This was the real gap — the `_is_human_prompt` reset logic had no test proving it worked.

**Error-path coverage** — added two tests: missing `transcript_path` in the payload, and `transcript_path` pointing at a nonexistent file. Both already fell back to `{}` in the implementation, just had no test confirming it.

**Line-length nit** — checked before changing anything: line 88 is 77 characters, and running `ruff format`/`ruff check` (the project's own enforced formatter per the Standards doc) shows no violation — in fact `ruff format` actively collapses a manual line-break there back into one line. Left `client.py` unchanged rather than fight the formatter; happy to revisit if there's a different lint config I'm missing.

avanish-garg:
any update on this @sfc-gh-jreini ?

PR #2798, opened 2026-09-18
test(feedback): run the score-range suite in CI and record its asymmetry

feiiiiii5:
# Description

Closes the two open notes from @joshreini1's review on #2653. No behaviour changes: the
`llm_provider.py` diff is comment-only, and the test file is moved with a docstring correction.

**1. The #2653 regression tests have never run in CI.** `make test-unit` enumerates its inputs
literally:

```make
$(PYTEST) $(shell ls tests/unit/test_*.py tests/unit/providers/test_*.py | grep -v test_otel)
```

Every PR job reaches that one line through `test-%-basic` / `-optional` / `-snowflake`, and
`ci-eval-pr.yaml` only ever passes `testSuite: unit`. `tests/integration/` is named by no Makefile
target, no `.azure_pipelines/` template and no workflow file, so anything placed there is not
collected by any job. Moving `test_score_range_validation.py` to `tests/unit/` is what the review
asked for ("Consider moving it to `tests/unit/` so it runs in the normal `make test-unit` sweep"),
and the file fits that suite on its own terms: it contacts no external service, the provider being
mocked at the transport boundary.

**2. The list/scalar asymmetry is now recorded in the code**, at the branch that has it, per
"worth calling out in a comment". Dropping an out-of-range item instead of raising is deliberate —
raising would discard the valid ratings shipped in the same response — but nothing said so, and the
`-1.0` sentinel when *every* item is out of range is a consequence a caller could easily assume
away.

Verified on the moved file (CPython 3.11.15, pytest 9.1.1, editable install of `src/core` +
`src/feedback` + `src/otel/semconv` from the branch tree, provenance printed from inside the run):

- `pytest --rootdir=. tests/unit/test_score_range_validation.py` -> **8 passed** in the default
  (basic) mode that the CI jobs use. The file carries no `required_only` / `optional` /
  `snowflake` / `huggingface` marker, so `conftest.py` keeps it in the basic sweep rather than
  skipping it.
- **The suite bites.** Making `_validate_score_range` return before its comparison turns exactly
  those tests red with `Failed: DID NOT RAISE ParseError` -> `4 failed, 4 passed`; restoring the
  file byte-identically (sha256 checked) returns `8 passed`. Before this move, no CI job would have
  observed either state.
- Enumeration, read off the commits rather than a working tree: applying the `make test-unit` glob
  to `33e4f661` matches the file **0** times, to `620bf1f3` matches it **1** time.
- Neighbouring parser suites unaffected: `tests/unit/test_*score*.py tests/unit/test_*feedback*.py`
  -> **101 passed, 13 skipped, 189 subtests passed**.
- `ruff check` and `ruff format --check` at the repo's own pin (0.16.3) are clean on both files.
- Full diff: 2 files, **+9 / −1**.

One thing this does *not* do, so it is not mistaken for a claim: `tests/integration/` holds nine
files in the same situation, including `test_score_parsing_normalization.py` and
`test_score_parsing_pipeline.py`, none of which CI collects. Moving all of them is a separate call
about what that directory is for, so it is left alone here.

## Type of change

  not work as expected)
- [x] New Tests

"New Tests" in the CI sense: eight tests that previously executed nowhere now execute in every PR
job. No assertion was written or changed in the move.

## Certification


sfc-gh-jreini (approved):
Clean fix that addresses both review notes from #2653. The comment at lines 277-282 clearly explains why the list path drops out-of-range items instead of raising (preserving valid ratings in the same response), and the test move to `tests/unit/` is correct: it mocks at the transport boundary with no external service. The `-1.0` sentinel behavior is now documented. Ready to merge.

feiiiiii5:

PR #2799, opened 2026-09-19
fix(snowflake): external agent app ids were a 1-tuple, so every listed app warned about a custom app_id

feiiiiii5:
## What happened

`SnowflakeEventTableDB.get_apps()` builds one `AppDefinition` per external agent version, and it wrapped the computed id in a one-element tuple:

```python
app_id = (
    app_schema.AppDefinition._compute_app_id(
        app_name, app_version
    ),
)
```

`AppDefinition.__init__` always recomputes the id (`src/core/trulens/core/schema/app.py:151`) and then compares it with whatever the caller passed:

```python
if app_id is not None and kwargs["app_id"] != app_id:
    warnings.warn("Custom `app_id` values are deprecated. …", DeprecationWarning, …)
    # After dep period:
    # raise ValueError("`app_id` does not match with `app_name` and `app_version`! …")
```

A `str` never equals a 1-tuple, so **every** app the connector lists is reported as if it had passed a deprecated custom id. Today that is a false `DeprecationWarning` per app; when the commented-out `ValueError` is enabled, the same line turns `get_apps()` into a hard failure for external agents. Every other `_compute_app_id` call site (`core/database/base.py:1360`, `core/schema/app.py:151`, `migrations/versions/6_populate_app_name_and_version_data.py:44`, `scripts/seed_dashboard_test.py:46`, `tests/unit/test_otel_get_records_and_feedback.py`) passes the plain string, so the parens are not a convention — they are a slip.

## What this PR does

* Passes the string `_compute_app_id(...)` returns, instead of a tuple containing it. The stored/reported `app_id` value is unchanged (it was already recomputed by `__init__`); what changes is that the caller no longer contradicts it.
* Adds `TestEventTableDBExternalApps` to `tests/unit/test_snowflake_external_agent_dao.py`: it builds the DB without a live session, promotes `DeprecationWarning` to an error, calls `get_apps(app_name="my_agent")` and asserts each yielded `app_id` equals `AppDefinition._compute_app_id(name, version)`.

## How it was verified

Environment: this commit (`33e4f66`), Python 3.11.15, `trulens-core` and `trulens-connectors-snowflake` installed editable from the checkout (verified by printing `__file__` for both modules), no Snowflake credentials and no network calls — the mocks are the same style the neighbouring DAO tests already use.

Before the source change, the new test fails at the call site:

```
$ TEST_SNOWFLAKE=1 python -m pytest tests/unit/test_snowflake_external_agent_dao.py -k ExternalApps
E   DeprecationWarning: Custom `app_id` values are deprecated. Use `app_name` and/or `app_version` instead.
src/core/trulens/core/schema/app.py:156: DeprecationWarning
FAILED ...::TestEventTableDBExternalApps::test_listed_apps_do_not_report_a_custom_app_id
1 failed, 9 deselected in 1.90s
```

After it:

```
$ TEST_SNOWFLAKE=1 python -m pytest tests/unit/test_snowflake_external_agent_dao.py
tests/unit/test_snowflake_external_agent_dao.py ..........        [100%]
10 passed in 2.87s

$ TEST_SNOWFLAKE=1 python -m pytest tests/unit/test_snowflake_sql_utils.py tests/unit/test_snowflake_metrics_usage.py \
      tests/unit/test_snowflake_run_status.py tests/unit/test_snowflake_run_dao.py tests/unit/test_otel_snowflake_event_table_exporter.py
45 passed in 2.03s
```

`ruff check` and `ruff format --check` (repo config, the version `make lint`/`make format` use) are clean on both files. I also ran the two files through `AGENTS.md`'s 80-column and import-style rules.

## Notes for review

* The test carries `@pytest.mark.snowflake` because that is what every test in this file uses and it needs the optional `trulens.connectors.snowflake` package to import. It needs no credentials. If you would rather have it in the `optional` group so it runs in a wider job, that is a one-line marker change — say which you prefer. For the same reason I am not claiming a CI job runs it: there is no pytest workflow in `.github/workflows`, so the marker groups are driven from your internal CI.
* The cortex-agent branch (`agent_versions = ["base"]`) is not covered; the test drives the external-agent branch, which is the one that reaches the tuple. Covering the other branch means a second `SHOW AGENTS IN ACCOUNT` mock and does not touch the fixed line.
* Not changed here on purpose: `AppDefinition.__init__` compares `kwargs["app_id"] != app_id` without knowing the caller's intent, which is why a wrong type reads as "custom id deprecated" rather than "wrong type". That is a separate design question and would touch every provider of an id.
* CLA: this account still owes the one-time TruLens CLA signature, which only my human maintainer can give — it is the only thing blocking #2798 (already `APPROVED`) as well as this PR. Nothing else here needs a human step.

## Certification

## Authorship and review

Written with an AI assistant (Qoder CLI, Claude); I have read every line and can defend it, and the commands above are real output rather than a description of intent. A separate fresh-context review pass over this diff was still running when I opened the PR; its findings will be fixed or answered here in a follow-up comment rather than left implicit.


feiiiiii5:
### Fresh-context review pass

A reviewer with no sight of the authoring conversation went through this diff. It confirmed the diagnosis and the marker choice, corrected one thing I said about CI, and produced two follow-ups.

**Correction to my description.** I wrote that I could not claim a CI job runs the new test. That was too modest: the `snowflake` marker group is part of PR validation — `.azure_pipelines/ci-eval-pr.yaml` runs `pytestMarker: snowflake, testSuite: unit`, which goes through `make test-unit-snowflake` (`TEST_SNOWFLAKE=true`), and `templates/run-tests.yaml:32` skips that leg only on Python 3.12. The marker means "snowflake *dependencies*" (see the marker list in `pyproject.toml`), not credentials, and `tests/unit/test_snowflake_sql_utils.py:15` is the precedent for a mocked, credential-free test carrying it. The test should therefore run here. I fixed the skip message, which named the wrong gate, in a23e7ba.

**Also confirmed.**
- Passing no `app_id` at all is behaviourally identical to passing the computed string, so the explicit value is an assertion about what the connector means, not a workaround; `AppDefinition`'s guard is doing its job and is untouched here.
- The test is not a tautology: today it fails on the warning; it would fail on the equality assert if `AppDefinition` ever honoured a caller-supplied id; and when the deprecation period ends it fails on the `ValueError`. `list()` inside the `with` block is load-bearing because `get_apps` is a generator.

**Two things I left out on purpose.**
1. The cortex-agent branch (`agent_versions = ["base"]`) and the `app_name=None` path are not exercised — the test pins `SHOW AGENTS IN ACCOUNT` to an empty frame. Happy to add either here if a reviewer prefers that over a follow-up.
2. A separate bug in the same method, different root cause, so not bundled: when there are no external agents, listing them raises —

   ```
   $ db._external_agent_dao._list_agents() -> pd.DataFrame()   # as the DAO returns for "no agents"
   $ list(db.get_apps())
   KeyError: 'name'
   ```

   at `snowflake_event_table_db.py:290`, even though `dao/external_agent.py:132` guards precisely that shape (`agents.empty or "name" not in agents.columns`). I will file it as an issue with this repro and send a two-line guard as its own PR — say the word if you would rather see it folded in here.

sfc-gh-jreini (approved):
Clean fix for a type mismatch that caused false deprecation warnings on every external agent. The tuple wrapping was clearly wrong—`AppDefinition.__init__` recomputes the id and compares it with what the caller passed, and a string never equals a 1-tuple. The test is well-structured, follows project patterns (mocking, unittest.TestCase, pytest marker), and uses the right approach (promoting DeprecationWarning to error). Ready to merge once CLA is signed.

feiiiiii5:
Follow-up on the two points in my earlier comment here, now that both are measured.

**1. The sibling guard went out as its own PR:** #2801, from the issue I filed as
#2800 for the `KeyError: 'name'` case (`get_apps()` on an account with no external
agents). Different root cause from this one, and the branches do not collide —
`git merge-tree --write-tree fix/snowflake-get-apps-no-external-agents
fix/snowflake-external-app-id-tuple` exits 0 with a single tree, so they can land
in either order. #2801's positive-control test deliberately asserts on `app_name`
and `app_version` rather than `app_id` so it does not pin the value this PR changes.

**2. Why the `DeprecationWarning` assertion in this PR's test is load-bearing.**
`AppDefinition.__init__` (`src/core/trulens/core/schema/app.py:154-157`) computes
the id itself and warns only when the caller's `app_id` differs from it:

```python
if app_id is not None and kwargs["app_id"] != app_id:
    warnings.warn("Custom `app_id` values are deprecated. ...")
```

A 1-tuple is never equal to the computed string, so on `main` the warning does
fire — I saw it in the run output while working on #2800:
`snowflake_event_table_db.py:307: DeprecationWarning: Custom `app_id` values are
deprecated`. With the plain string it matches and nothing is raised, which is what
`warnings.simplefilter("error", DeprecationWarning)` in
`test_listed_apps_do_not_report_a_custom_app_id` turns into a failure at base.
So that test is not just asserting the shape I chose, it is asserting the
user-visible symptom (every listed app reported as a custom app id) disappears.

feiiiiii5:

PR #2801, opened 2026-09-19
fix(snowflake): list no apps when the account has no external agents

### LINKED ISSUE #2800: [BUG] SnowflakeEventTableDB.get_apps() raises KeyError: 'name' when the account has no external agents
### Bug description

Listing apps through the Snowflake connector raises `KeyError: 'name'` instead
of returning an empty listing, on any account that has no external agents.

`src/connectors/snowflake/trulens/connectors/snowflake/snowflake_event_table_db.py:291`:

```python
app_names = self._external_agent_dao._list_agents()["name"].values
```

`_list_agents` runs `SHOW EXTERNAL AGENTS;` through
[execute_query][trulens.connectors.snowflake.dao.sql_utils.execute_query], and
that helper returns a bare `pd.DataFrame()` when a non-`SELECT` statement comes
back with no rows (`sql_utils.py:60-61`: `if not result: return
pd.DataFrame()`). A
frame with no rows also has no columns, so the `["name"]` index fails:

```
  File ".../trulens/connectors/snowflake/snowflake_event_table_db.py", line 291, in get_apps
    app_names = self._external_agent_dao._list_agents()["name"].values
  ...
KeyError: 'name'
```

The DAO already knows this shape: `check_agent_exists` guards it with
`if agents.empty or "name" not in agents.columns` (`dao/external_agent.py:132`).
`get_apps` indexes the same result unguarded.

Reachable from `TruSession.get_apps()` (`src/core/trulens/core/session.py:981`)
and the dashboard's app listing (`src/dashboard/trulens/dashboard/streamlit.py:110`,
`dashboard/utils/dashboard_utils.py:590`).

### To Reproduce

No Snowflake account needed — the fake session below is the same shape the
connector gets from a real one when the account has no external agents:

```python
from trulens.connectors.snowflake.snowflake_event_table_db import (
    SnowflakeEventTableDB,
)


class Result:
    def collect(self):
        return []  # what `SHOW EXTERNAL AGENTS;` returns on an empty account


class Session:
    def sql(self, query, params=None):
        return Result()


db = SnowflakeEventTableDB(snowpark_session=Session())
list(db.get_apps())  # KeyError: 'name'
```

Expected: an empty listing, as it is for the SQLAlchemy connector.

### Environment

Measured on `main` at `33e4f66`, `trulens-core` / `trulens-connectors-snowflake`
2.14.0, snowpark 1.55.0, pandas 3.0.6, Python 3.11.

### Additional context

I have the one-line guard and a regression test for this ready to send as its
own PR, kept separate from #2799 because that one is about the `app_id` value
rather than the empty listing.

feiiiiii5:
Closes #2800.

## What

`SnowflakeEventTableDB.get_apps()` raised `KeyError: 'name'` on a Snowflake
account that has no external agents, instead of yielding an empty listing.

```python
app_names = self._external_agent_dao._list_agents()["name"].values
```

`_list_agents` sends `SHOW EXTERNAL AGENTS;` through
`dao/sql_utils.py::execute_query`, and for a non-`SELECT` statement with no rows
that helper returns a frame with no columns at all (`sql_utils.py:60-61`):

```python
if not result:
    return pd.DataFrame()
```

So the very first call on a fresh account — `TruSession.get_apps()`
(`src/core/trulens/core/session.py:981`), or the dashboard's app listing
(`src/dashboard/trulens/dashboard/streamlit.py:110`) — dies on the index rather
than reporting "no apps". The SQLAlchemy connector returns an empty listing for
the same case.

The DAO already tests this shape in two places, both against a
`list_agent_versions` result:

```python
if agents.empty or "name" not in agents.columns:      # dao/external_agent.py:132
if (existing_versions.empty                            # dao/external_agent.py:49-53
        or "name" not in existing_versions.columns ...):
```

This adopts the same condition at the one call site that lacked it.

## Change

Four lines in `get_apps`: keep the frame in a local, stop the generator when it
has no `name` column, and otherwise index it as before.

```python
if app_name is None:
    agents = self._external_agent_dao._list_agents()
    if agents.empty or "name" not in agents.columns:
        return

    app_names = agents["name"].values
```

Fixing `execute_query` instead was considered and doesn't work as stated: on zero
rows there is no `result[0]._fields` to read column names from
(`sql_utils.py:63`), so it would need an extra round trip per empty `SHOW`. The
consumer-side guard is what this connector already does.

`get_apps` also indexes `list_agent_versions(...)["name"]` a few lines below, same
crash class. Applied per review: an agent that reports no versions is now skipped
rather than raising, matching `create_agent_if_not_exist`, which already treats an
empty version listing as a state it must handle (`dao/external_agent.py:49-53`).
I could not construct a Snowflake account state that reaches it, so it ships for
defensive consistency rather than as a reproduced crash.

## Tests

New `tests/unit/test_snowflake_event_table_db.py`, four cases:

1. `test_zero_row_show_statement_yields_no_columns` pins the precondition — the
   empty, columnless frame comes out of the production `execute_query`, not out
   of a mock, so the regression test cannot pass vacuously.
2. `test_get_apps_lists_agents_that_exist` is the positive control: the guard
   must not turn into a blanket "no apps". It asserts on `app_name` and
   `app_version` only, deliberately not on `app_id`, because #2799 changes that
   value in this same function.
3. `test_get_apps_without_external_agents_yields_nothing` is the bug.
4. `test_get_apps_skips_an_agent_that_reports_no_versions` covers the second
   guard and asserts a *sibling* agent is still listed, so the fix cannot be a
   blanket early return either way.

The fakes implement only the three `SHOW` statements `get_apps` makes, and
`FakeSession.sql` raises for anything else, so a future query added to this
method fails the test loudly instead of silently reading an empty frame.

### Measured

cwd `/private/tmp/trulens`, `/private/tmp/tlvenv/bin/python` = 3.11 with
`trulens-core` 2.14.0 and `trulens-connectors-snowflake` 2.14.0 installed
editable from this tree, snowpark 1.55.0, pandas 3.0.6. Each run printed
`trulens.connectors.snowflake.snowflake_event_table_db.__file__` first to confirm
which tree executed.

| run | source state (proven by echoing `git hash-object` of the file before the run) | result |
| --- | --- | --- |
| no guard at all | `main` `33e4f66`, blob `1f220532…` | `2 failed, 2 passed` — both listing cases raise `KeyError: 'name'` |
| agents guard only | pushed head `7b8cd30`, blob `97663cfe…` | `1 failed, 3 passed` — the version listing still raises |
| both guards | this head, blob `0e13e6fe…` | `4 passed` |
| same file, marker env removed | this head | `4 skipped` — the new file cannot break the basic group |
| neighbours | each of `tests/unit/test_snowflake_*.py`, `tests/unit/test_otel_snowflake_event_table_exporter.py` | 9, 10, 12, 13, 8 and 2 passed |
| lint | `ruff check` / `ruff format --check` (0.16.3, the rev pinned in `.pre-commit-config.yaml`) on both files | clean, no added line past 80 columns |

### Which CI runs it

The file carries the `snowflake` marker, so PR validation reaches it:
`.azure_pipelines/ci-eval-pr.yaml:152-155` includes
`pytestMarker: snowflake, testSuite: unit`, `templates/run-tests.yaml:31` calls
`make test-unit-snowflake`, and the `Makefile:296-297` pattern rule runs
`SKIP_BASIC_TESTS=1 TEST_SNOWFLAKE=true make test-unit`. `run-tests.yaml:32`
skips that marker group on py3.12 only, so the 3.10/3.11/3.13 legs execute it.
Locally the same variable is required, otherwise pytest prints "Skipping snowflake
tests" and reports no failures for the wrong reason.

## Independent review

A reviewer with no sight of the authoring conversation re-ran the mutation
(delete the guard → `1 failed, 2 passed`; restore → `3 passed`, source hash
identical before and after), confirmed the empty frame is produced by real
`execute_query` rather than handed by a mock, checked every caller wraps in
`list()` or iterates so a `return` inside the generator is safe, and cherry-picked
this commit onto #2799's branch in a scratch worktree: auto-merged with no
conflict markers, and `test_get_apps_lists_agents_that_exist` passes on both
branches — which is the check that the positive control does not pin `app_id`.
It also produced the `execute_query` objection and the second DAO precedent above.
(Those counts are for the three-test file as it stood during that pass; the
four-test counts are in the table above.)

## Review round

`@sfc-gh-jreini` reviewed the first push as "correct and well-tested" and asked for
one consistency change — guard `list_agent_versions(...)` the same way — plus a
nit to assert the frame type in the precondition test. Both are in `b433387`:
the version listing now skips a versionless agent instead of raising, and the new
fourth test keeps a *sibling* agent with versions in the expected output so the
guard cannot become a blanket stop. I still have no Snowflake account state that
reaches the versionless case, so it ships as the requested consistency guard
rather than as a reproduced crash.

## Relation to #2799

Same function, different root cause, so a separate PR as offered in that thread.
The two branches do not collide: `git merge-tree --write-tree
fix/snowflake-get-apps-no-external-agents fix/snowflake-external-app-id-tuple`
exits 0 and prints a single tree (`ccba9c6` at `b433387`, re-run after the second
commit; `db5411b` at the first), with no conflict markers, i.e. they combine
cleanly in either order.

## Notes

- Correction on my side: the first push of this branch (`3a0c7cb`) carried the
  upstream committer's email as its author and committer, because I derived the
  commit identity from `git log -1` of the base commit instead of from this
  account. `7b8cd30` is the same tree with the identity set to this account's own
  noreply address; the diff is unchanged and the 3 tests were re-run after it.
  Flagging it because the CLA check passed on `3a0c7cb` for the wrong reason —
  someone else's signed address — and is red again on `7b8cd30`, as it is on
  #2798 and #2799.
- So the `cla` check stays red until the account owner posts the one-time CLA
  signature the CLA bot asks for. That is a human action and I have not taken it.
  It is new on this side: `gh pr checks` on #2617 and #2653 (both merged) lists 9
  checks and no `cla` entry. One signature comment covers #2798, #2799 and this
  PR, so it is one ask rather than three.

## Certification


sfc-gh-jreini (commented):
The fix is correct and well-tested. It adds the same defensive guard already used in the DAO (external_agent.py:132), backed by three solid unit tests covering the crash case, positive case, and precondition verification. The PR description is exceptionally thorough, including independent review and merge conflict analysis. One consistency gap: lines 298-300 also index list_agent_versions(...)["name"] without a guard; while you couldn't reproduce the crash, apply the same defensive pattern there for consistency. VERDICT: Comment (1 should-fix for defensive consistency).

tests/unit/test_snowflake_event_table_db.py
```diff
@@ -0,0 +1,108 @@
...
+                "tests are disabled."
+            )
+
+    def test_zero_row_show_statement_yields_no_columns(self):
+        """Record what an empty account looks like to its callers.
+
+        ``execute_query`` builds its frame from the rows it got back, so a
+        statement with no rows has no ``name`` column at all rather than an
+        empty one.
+        """
```
sfc-gh-jreini, line 75: Consider adding `self.assertIsInstance(df, pd.DataFrame)` before the length check to make the precondition test even more explicit about what `execute_query` returns — though the current assertions are sufficient.

feiiiiii5:
Both points addressed in b433387.

**The consistency guard.** `list_agent_versions(...)` is now handled the same way,
and the new fourth case asserts that a *sibling* agent which does have versions is
still listed — so the guard cannot quietly degrade into a blanket stop in either
direction. Measured across three source states, each printing
`git hash-object` of `snowflake_event_table_db.py` first so the control could not
silently measure the patched tree:

| source state | blob | result |
| --- | --- | --- |
| `main` `33e4f66`, no guards | `1f220532…` | `2 failed, 2 passed` |
| previous head `7b8cd30`, agents guard only | `97663cfe…` | `1 failed, 3 passed` |
| this head `b433387`, both guards | `0e13e6fe…` | `4 passed` |

The other six `tests/unit/test_snowflake_*` files are unchanged at 9/10/12/13/8/2
passed, and `ruff` 0.16.3 check plus `format --check` are clean with nothing past
80 columns.

One thing I want to keep honest: I still have no Snowflake account state that
produces an empty `SHOW VERSIONS` for an agent that exists. So this ships as the
consistency guard you asked for, backed by the DAO's own precedent in
`create_agent_if_not_exist` (`dao/external_agent.py:49-53` treats an empty version
listing as a state to handle), not as a reproduced crash. If you know the account
state where it does fire, I will add it to #2800.

The nit is applied too: the precondition test now asserts `isinstance(df,
pd.DataFrame)` before the length and column checks.

FYI on the checks: `cla` flipped from green to red between the two pushes because
the first push carried the wrong commit identity (an upstream committer's email,
which the CLA bot treats as already signed). `b433387` is authored by this account,
so the one-time signature really is still outstanding — spelled out in the
description's Notes.

tests/unit/test_snowflake_event_table_db.py
feiiiiii5: Done in b433387 — the precondition test now opens with `self.assertIsInstance(df, pd.DataFrame)` before the length and column checks. `ruff` 0.16.3 check and `format --check` are still clean on both files.

feiiiiii5:
Following up on the one should-fix from the review — "lines 298-300 also index `list_agent_versions(...)["name"]` without a guard".

That is covered by the commit pushed right after the review, b433387 ("fix(snowflake): skip an agent whose version listing comes back empty"). Both `["name"]` reads in `get_apps` now go through the same pattern as the DAO:

- `_list_agents()` → `if agents.empty or "name" not in agents.columns: return` (head lines 291-295)
- `list_agent_versions(app_name)` → `if versions.empty or "name" not in versions.columns: continue` (head lines 302-308)

The line numbers in the review are from `main`; in the diff those two sites are the first and second hunks, which is easy to miss since the second hunk starts inside the `for app_name in app_names` loop.

Tests covering each guard, in `tests/unit/test_snowflake_event_table_db.py`:

- `test_zero_row_show_statement_yields_no_columns` - the precondition: a zero-row result really does come back with no `name` column, so the guard is reachable rather than defensive ceremony.
- `test_get_apps_without_external_agents_yields_nothing` - the `_list_agents()` guard.
- `test_get_apps_skips_an_agent_that_reports_no_versions` - the `list_agent_versions()` guard: one agent with no version rows is skipped, the next one is still returned.
- `test_get_apps_lists_agents_that_exist` - unchanged behaviour when both frames are populated.

I re-read both guards and all four tests in the head tree today before posting this; if you would prefer the version guard expressed as a single helper shared with the `_list_agents()` path, say the word and I will fold them together.

feiiiiii5:

sfc-gh-jreini (approved):
Thanks for addressing feedback. Looks good!

Issue #2446, opened 2026-05-04
AutoGen Integration: Instrument Multi-Agent Conversations


sfc-gh-jreini:
## Summary

Add a `TruAutoGen` auto-instrumentation wrapper (following the `TruLlama`/`TruChain` pattern) for Microsoft AutoGen (AG2) multi-agent conversations, capturing agent messages, tool calls, and conversation flow as structured OTEL spans.

## Motivation

AutoGen is widely used for multi-agent chat patterns (two-agent debate, group chat, nested chats). Users need to evaluate individual agent contributions, conversation quality, and tool usage within AutoGen workflows — with zero code changes to their AutoGen app.

### Package Structure

Create `src/apps/autogen/` as a new installable package (`trulens-apps-autogen`).

### API

```python
from trulens.apps.autogen import TruAutoGen

agent = ConversableAgent(...)
tru_agent = TruAutoGen(
    agent,
    app_name="debate_agent",
    app_version="v1",
    metrics=[coherence, relevance],
)

with tru_agent as recording:
    result = agent.initiate_chat(other_agent, message="...")
```

### Auto-Instrumentation

`TruAutoGen` should define `Default.CLASSES` and `Default.METHODS` to auto-instrument:
- `ConversableAgent.generate_reply()` → `AGENT` span
- `ConversableAgent.execute_function()` → `TOOL` span
- `GroupChat.select_speaker()` → `WORKFLOW_STEP` span
- LLM calls within agents → `GENERATION` span

### Captured Attributes
- `agent.name`, `agent.system_message`
- `message.role`, `message.content`
- `groupchat.speaker_selection_method`
- Full message history per turn

## Acceptance Criteria

- [ ] `TruAutoGen` class in `src/apps/autogen/` extending `core_app.App`
- [ ] Installable as `pip install trulens-apps-autogen`
- [ ] Group chat produces span tree: GroupChat → Speaker turns → Agent replies → Tool calls
- [ ] Message content captured as span attributes
- [ ] Metrics can evaluate individual agent replies via selectors
- [ ] `main_input()` and `main_output()` methods implemented
- [ ] Example notebook: "Evaluate AutoGen Group Chat Quality"
- [ ] Works with AutoGen v0.4+ / AG2

## Related

- Issue #2423 (conversation_id for thread grouping)
- Issue #2440 (nested app support)

snakefood3232:
[...]

PR #2802, opened 2026-09-19
feat(apps): add TruAutoGen for AutoGen (AG2) multi-agent conversations

desusaiteja:
# Description

Adds `trulens-apps-autogen`, a `TruAutoGen` recorder for [AutoGen (AG2)](https://github.com/ag2ai/ag2) multi-agent conversations, closing #2446.

Wrapping any one agent is enough — `TruAutoGen` instruments the AutoGen classes, so the agents a recorder never sees (a recipient passed to `initiate_chat`, the members of a group chat, agents created after the recorder) are recorded too.

```python
from trulens.apps.autogen import TruAutoGen

tru_recorder = TruAutoGen(
    user, app_name="research_team", app_version="v1", main_method=user.initiate_chat
)

with tru_recorder as recording:
    user.initiate_chat(manager, message="Why is the sky blue?")
```

A group chat comes out as:

```
[record_root]     ConversableAgent.initiate_chat
  [agent]           chat_manager.generate_reply
    [workflow_step]   select_speaker: researcher
    [agent]           researcher.generate_reply
      [tool]            execute_function: word_count
      [generation]      OpenAIWrapper.create
    [workflow_step]   select_speaker: writer
    [agent]           writer.generate_reply
```

Because each agent turn is its own `AGENT` span, a metric can score agents individually rather than scoring the conversation as a whole:

```python
Feedback(provider.coherence_with_cot_reasons, name="Agent Coherence").on({
    "text": Selector(
        span_type=SpanAttributes.SpanType.AGENT,
        span_attribute=SpanAttributes.AGENT.OUTPUT_MESSAGE,
    ),
})
```

**Instrumented methods.** `ConversableAgent.generate_reply` / `execute_function` / `initiate_chat`, `GroupChat.select_speaker`, and `OpenAIWrapper.create`, each with its async twin. The spec lives in `AutoGenInstrument.Default.CLASSES` / `.METHODS` and is applied both through the usual `Instrument` object-graph walk and, once, at class level — the class-level pass is what catches agents that are not reachable from the wrapped one.

**Two changes outside the new package:**

- **`ai.observability.agent.*` in the semantic conventions.** The issue asks for `agent.name`, `agent.system_message`, per-turn message history and speaker selection method; only the agent's name had an existing home (`workflow.agent_name`, which is still emitted alongside). Added `AGENT` with name, description, system message, input messages/message/role, output message and error, plus `workflow.speaker_selection_method` and `workflow.participant_agent_names`. All additive, and documented in the conventions table.
- **Tool names on TOOL spans.** `_set_span_attributes` unconditionally overwrote `gen_ai.tool.name` with the instrumented function's name, so every AutoGen tool span read `ConversableAgent.execute_function` rather than the tool that ran. It now prefers a name the instrumentation supplies and falls back to the function name as before, so nothing that does not set the attribute changes.

**Costs.** The generation span deliberately does not set `ai.observability.cost.*`. TruLens sums cost attributes that land on a span more than once, and the provider cost computers already write onto whichever span is current — which is this one.

**Which AutoGen.** The `autogen` distribution, `>=0.4,<1.0`, which is AG2's classic `ConversableAgent` / `GroupChat` API. AG2 1.0 moved to a different top-level module with an incompatible API, hence the ceiling.

**Tests.** 13 unit tests in `tests/unit/test_otel_tru_autogen.py`, all hermetic: agents reply from registered reply functions and the generation test uses AutoGen's custom model client hook, so nothing leaves the process. Covers two-agent and group chat span trees, async entry points, tool calls nesting under the agent that ran them, failed tool calls, recording with and without `main_method`, `main_input`/`main_output`, and `TruSession.App` auto-detection. Also adds `test_instrumentation_autogen` to the static instrumentation checks.

## Type of change

  not work as expected)
- [x] New Tests

## Certification


src/apps/autogen/trulens/apps/autogen/tru_autogen.py
```diff
@@ -0,0 +1,706 @@
...
+        return None
+
+    if isinstance(value, str):
+        return value
+
+    try:
+        return json.dumps(value, default=str)
+    except Exception:
+        return str(value)
+
```
sfc-gh-jreini, line 72: Change the bare `except Exception:` to log at warning level instead of debug when span renaming fails. Span naming failures are unusual and may indicate instrumentation issues that should be more visible. Use `logger.warning("Could not rename span to %s.", span_name, exc_info=True)` to make unexpected errors easier to diagnose in production.

src/core/trulens/core/otel/instrument.py
```diff
@@ -343,10 +344,15 @@ def _set_span_attributes(
...
         set_genai_tool_attributes(
             span,
-            # func_name is the instrumented function / tool name.
-            tool_name=func_name,
-            call_arguments=resolved_attributes.get("call_arguments"),
-            call_result=resolved_attributes.get("call_result"),
+            # Instrumentation that knows the tool being called reports it
+            # directly; otherwise func_name is the instrumented function.
+            tool_name=resolved_attributes.get(
+                GenAIAttributes.TOOL.NAME, func_name
```
sfc-gh-jreini: Verify the new fallback logic for `tool_name` doesn't break existing instrumentation where the tool name was intentionally derived from `func_name`. The change looks correct — preferring an explicit `GenAIAttributes.TOOL.NAME` when present, falling back to `func_name` otherwise — but confirm this was tested against existing TOOL span instrumentation (MCP, general tool calls) to ensure they still report the expected names.

src/apps/autogen/trulens/apps/autogen/tru_autogen.py
```diff
@@ -0,0 +1,706 @@
...
+    from autogen.oai.client import OpenAIWrapper
+except ImportError:  # pragma: no cover - depends on the autogen build
+    OpenAIWrapper = None
+
+logger = logging.getLogger(__name__)
+
+MAX_INSTRUMENTED_MESSAGES = 50
+"""Most recent messages kept when serializing a conversation history.
+
+A long-running group chat replays its whole history on every turn, so keeping
```
sfc-gh-jreini, line 54: Add `from __future__ import annotations` at the top of the module (before other imports) per AGENTS.md conventions. This enables forward references and deferred evaluation of type hints, which is the project standard. Then call `TruAutoGen.model_rebuild()` after the class definition (which you already do at line 706).

sfc-gh-jreini (commented):
Strong implementation of AutoGen integration with comprehensive tests, excellent documentation, and well-designed semantic conventions. The instrumentation correctly captures agent turns, tool calls, and group chat flows. One SHOULD-FIX around exception handling transparency, and one QUESTION to confirm the tool name fallback logic is intentional. No blockers. Verdict: Approved pending minor clarification — the tool name change warrants explicit confirmation it won't affect existing instrumentation.

desusaiteja:

desusaiteja:
> Strong implementation of AutoGen integration with comprehensive tests, excellent documentation, and well-designed semantic conventions. The instrumentation correctly captures agent turns, tool calls, and group chat flows. One SHOULD-FIX around exception handling transparency, and one QUESTION to confirm the tool name fallback logic is intentional. No blockers. Verdict: Approved pending minor clarification — the tool name change warrants explicit confirmation it won't affect existing instrumentation.

Thanks for the review! All three are addressed now

docs/component_guides/instrumentation/autogen.md
```diff
@@ -0,0 +1,117 @@
...
+
+Because every agent turn is its own span, a metric can be pointed at agent
+replies rather than at the conversation as a whole. Each `AGENT` span is
+scored separately, so a weak agent shows up on its own rather than being
+averaged into the final answer.
+
+!!! example "Score every agent reply"
+
+    ```python
+    from trulens.core import Feedback
```
sfc-gh-jreini, line 91: please update to use newer Metric() api in place of Feedback

examples/expositional/frameworks/autogen/autogen_group_chat_quality.ipynb
```diff
@@ -0,0 +1,320 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "provider = fOpenAI(model_engine=\"gpt-4o-mini\")\n",
+    "\n",
+    "f_answer_relevance = (\n",
+    "    Feedback(provider.relevance_with_cot_reasons, name=\"Answer Relevance\")\n",
```
sfc-gh-jreini, line 175: update here also to use Metric() api in place of Feedback()

src/apps/autogen/trulens/apps/autogen/tru_autogen.py
```diff
@@ -0,0 +1,708 @@
...
+from autogen.agentchat.chat import ChatResult
+
+try:
+    from autogen.oai.client import OpenAIWrapper
+except ImportError:  # pragma: no cover - depends on the autogen build
+    OpenAIWrapper = None
+
+logger = logging.getLogger(__name__)
+
+MAX_INSTRUMENTED_MESSAGES = 50
```
sfc-gh-jreini, line 51: is 50 a reasonable default here? seems kind of small

sfc-gh-jreini (changes requested):

sfc-gh-jreini:
Thanks @desusaiteja ! Added a couple more smaller fixes and then this is good to go.

PR #2804, opened 2026-09-22
Stop the migrations from reconfiguring the host application's logging

YHC66:
# Description

Creating a `TruSession` runs the database migrations, and `migrations/env.py` applied the bundled `alembic.ini` with `logging.config.fileConfig()` on every run. `fileConfig()` defaults to `disable_existing_loggers=True` and replaces the root logger's level and handlers, so after `TruSession()` the host application loses its own logging:

```python
import logging
logging.basicConfig(level=logging.INFO)
log = logging.getLogger("myapp")

from trulens.core import TruSession
TruSession()

log.info("after TruSession")  # never printed
log.disabled                  # True
```

TruLens' own loggers are affected the same way: the warning `is_otel_tracing_enabled()` emits when `TRULENS_OTEL_TRACING=0`, for example, is dropped once any migration has run.

This applies the `alembic.ini` logging configuration only when alembic runs from the command line. `alembic_config()` already sets `calling_context` to `"PYTHON"` on the library path (commented as "skips CLI-specific setup"), which is what the commented-out guard in `env.py` used to check. The library path always loads the `alembic.ini` bundled in the package, so skipping it there removes no setting a user could have changed.

- CLI behaviour is unchanged: without `calling_context`, `fileConfig()` still runs.
- `disable_existing_loggers=False` alone would not be enough: the root logger would still be reset to `WARN` with alembic's handler, so the application's `INFO` logs would still disappear.
- This also fixes `tests/unit/test_otel_utils.py::TestOtelUtils::test_warns_when_disabled`, which fails on `main` when it shares a worker with a test that ran the migrations (`pytest tests/unit/test_otel*.py -n 4 --dist=loadscope`, as `make test-unit` does): the logger it asserts on had been disabled. It failed in 3 of 3 runs on `main` and passes with this change.
- The new `tests/unit/test_migrations_logging.py` runs `upgrade_db()` against a temporary SQLite file and checks that a pre-existing logger stays enabled and the root logger's level and handlers are unchanged. It fails on `main` and passes here.

Verified on macOS with Python 3.12: `tests/unit/test_otel*.py` 194 passed (`main`: 1 failed, 193 passed); the other unit tests 515 passed (`main`: 514, the difference being the new test); `tests/integration/test_database.py` on SQLite gives the same results as `main`.

## Type of change

  not work as expected)
- [x] New Tests

## Certification


sfc-gh-jreini (approved):
Clean fix for a real bug. The migrations were disabling all pre-existing loggers and resetting the root logger on every `TruSession()` call because `fileConfig()` ran unconditionally. Restoring the `calling_context` guard correctly limits logging config to CLI usage, where it belongs. The test verifies loggers stay enabled and root config is untouched. Verified that `migrations/__init__.py:96` does set `calling_context="PYTHON"` as claimed. Ready to merge.

YHC66:

